PR #33, opened 2018-05-11
Implemented daily notification with custom time

javiercbk:
I've implemented it for android, I'll try to make it work for IOS as well

Caveat: The notification time is hardcoded in the example app

screen shot 2018-05-11 at 00 13 21

screen shot 2018-05-11 at 00 13 08

### DIFF
--- a/lib/flutter_local_notifications.dart
+++ b/lib/flutter_local_notifications.dart
@@ -125,7 +125,7 @@ class FlutterLocalNotificationsPlugin {
     {String payload}) async {
     Map<String, dynamic> serializedPlatformSpecifics =
         _retrievePlatformSpecificNotificationDetails(notificationDetails);
-    await _channel.invokeMethod('periodicallyShow', <String, dynamic>{
+    await _channel.invokeMethod('showDailyAtTime', <String, dynamic>{
       'id': id,
       'title': title,
       'body': body,
--- a/lib/flutter_local_notifications.dart
+++ b/lib/flutter_local_notifications.dart
@@ -15,13 +15,19 @@ class Time {
   final int minute;
   final int second;
 
-  Time([this.hour = 0, this.minute = 0, this.second = 0]);
+  Time([this.hour = 0, this.minute = 0, this.second = 0]) {
+    assert(this.hour >= 0 && this.hour < 24);
+    assert(this.minute >= 0 && this.minute < 60);
+    assert(this.second >= 0 && this.second < 60);
+  }
 
-  Map<String, int> get serialized => {
-    "hour": hour,
-    "minute": minute,
-    "second": second,
-  };
+  Map<String, int> toMap() {
+    return <String, int> {
+      "hour": hour,
+      "minute": minute,
+      "second": second,
+    };
+  }
 }
 
 class FlutterLocalNotificationsPlugin {
@@ -125,7 +131,7 @@ class FlutterLocalNotificationsPlugin {
       'body': body,
       'calledAt': new DateTime.now().millisecondsSinceEpoch,
       'repeatInterval': RepeatInterval.Daily.index,
-      'repeatTime': notificationTime.serialized,
+      'repeatTime': notificationTime.toMap(),
       'platformSpecifics': serializedPlatformSpecifics,
       'payload': payload ?? ''
     });
--- a/lib/flutter_local_notifications.dart
+++ b/lib/flutter_local_notifications.dart
@@ -10,6 +10,20 @@ typedef Future<dynamic> MessageHandler(String message);
 /// The available intervals for periodically showing notifications
 enum RepeatInterval { EveryMinute, Hourly, Daily, Weekly }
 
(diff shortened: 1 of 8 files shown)


lib/flutter_local_notifications.dart
```diff
@@ -100,6 +114,23 @@ class FlutterLocalNotificationsPlugin {
     });
   }
 
+  Future showDailyAtTime(int id, String title, String body,
+    Time notificationTime, NotificationDetails notificationDetails,
+    {String payload}) async {
+    Map<String, dynamic> serializedPlatformSpecifics =
+        _retrievePlatformSpecificNotificationDetails(notificationDetails);
+    await _channel.invokeMethod('periodicallyShow', <String, dynamic>{
```
MaikuB: invokeMethod should pass in 'showDailyAtTime' for better separation. This means that the onMethodCall override on the native side will need a different case to handle what needs to be done. If you could also split up repeatNotification to two separate methods (one for interval-based, the other for time-based) then that would be good as well

lib/flutter_local_notifications.dart
```diff
@@ -10,6 +10,20 @@ typedef Future<dynamic> MessageHandler(String message);
...
 enum RepeatInterval { EveryMinute, Hourly, Daily, Weekly }
 
+class Time {
+  final int hour;
+  final int minute;
+  final int second;
+
+  Time([this.hour = 0, this.minute = 0, this.second = 0]);
+
+  Map<String, int> get serialized => {
```
MaikuB: if you could make this a toMap() method to be consistent with the rest of the code then that would be great as reading something like `time.serialized` feels awkward

lib/flutter_local_notifications.dart
```diff
@@ -10,6 +10,20 @@ typedef Future<dynamic> MessageHandler(String message);
 /// The available intervals for periodically showing notifications
 enum RepeatInterval { EveryMinute, Hourly, Daily, Weekly }
 
+class Time {
+  final int hour;
+  final int minute;
+  final int second;
+
+  Time([this.hour = 0, this.minute = 0, this.second = 0]);
```
MaikuB: there should be some validation here to restrict the accepted values, possibly through assert statements. Should add documentation as well through comments but you can leave that for me to do

javiercbk:
Added support to IOS as well

screen shot 2018-05-11 at 04 11 49

screen shot 2018-05-11 at 04 11 41

MaikuB:
Yep I took a look and it looks about right but will need some of the changes I commented about earlier regarding the string that's passed from Dart side via the invokeMethod call. I think once all the changes are made, we're good to merge it. I'll then do some clean up and testing before publishing a new version.

Thanks for your contribution btw. I've just realised that I've been using your JSON to Dart tool. Super handy and I've asked on the Flutter repo about putting a link to your site but go no response :S

lib/flutter_local_notifications.dart
javiercbk: I have chosen this approach because I considered this feature as a Daily `periodicallyShow` use case, that way changes in both Java and Objective-C remain minimal. I can add that change though.

javiercbk:
I've made the requested changes to the Dart's Time class. I'll make the other changes tomorrow.

Glad that my Dart to JSON tool is helping others

javiercbk:
I've added a weekday bitwise flag option. I haven't test it though but I will tonight

MaikuB:
By that are you referring to the ability to just have notifications displayed on weekdays only? If so then I don't think that works for the user notifications framework on iOS with the date components as you'd have to create a separate notification for each day, which means each one has their own ID. On that note, i think it'll be possible to schedule a notification to fire on a specified a day of the week and time but that's a separate discussion :)

javiercbk:
In android it was easy, you just schedule a notification every day and when the intent is fired you just choose not to fire the notification. In IOS things are a bit mor complicated, you have to schedule one notification for each day.

My first solution is to provide the first notification ID and then increment the number. But that API is really ugly, it is just a proof of concept. If it works I would generate the notification id with some uuid or the unix timestamp.

MaikuB:
Yeah I had a feeling Android would be easier and that it would be more harder on iOS. I would say leavethat for the moment and if you could make the other changes then we can merge this PR in. Looking at how to do it for weekdays can be done in a separate PR or something I can look into even

javiercbk:
Ok, let me wrap up what I got and finish this PR tonight

MaikuB:
Cheers mate. I should have some bandwidth to look into the weekday one and touched on the rough idea i've got for it :)

javiercbk:
Done!

MaikuB:
Np, I should be able to merge your changes later. I do prefer Swift as well except I originally ran into issues with debugging in Swift. I heard there use to be issues with plugins written in Swift that were consumed by apps where the iOS head project is in Objective C. Not sure if that issue still exists as it's not something I've been monitoring

Issue #17, opened 2018-04-21
Support for custom notification actions


MaikuB:
The plugin should provide the ability to specify custom notification actions. However, this will depends on the Flutter engine being able to support headless-Dart code as per https://github.com/flutter/flutter/issues/6192 and https://github.com/flutter/flutter/issues/3671

It appears Android support is there but will wait to see on if the engine can support it for iOS applications, and for Flutter to provide abstractions to access the functionality. Without the support being added in, then this won't work for scenarios like when the application has been terminated as the logic associated with the action would've been defined in Dart. 

Update with remaining work (IMO)

- [x] rework callbacks to deal with "foreground" and "background" notification actions. See https://github.com/MaikuB/flutter_local_notifications/pull/1489 for more detailed discussion
- [x] potential issue with setting notification categories affecting ability to see the notification actions https://github.com/MaikuB/flutter_local_notifications/issues/1481

Note that last two perhaps could be omitted given market share and those require using deprecated APIs

Edit: 
- there is headless execution on iOS but there are limitations at the moment. See https://github.com/flutter/flutter/issues/23904. Upvote the issue on the Flutter repo if this concerns you to help the Flutter team prioritise
- removed tasks on supporting older iOS and macOS versions given market share and those require using deprecated APIs

mit-mit:
Hi @MaikuB do you have an example use case where of using a custom notification action on iOS?

MaikuB:
@mit-mit 

- a social media app may notify users when someone has replied their post, in which case actions may be presented to allow the user to either reply or like the reply
- similar to the clock app on Android, an app may present a reminder to a user and they can choose to have rescheduled (think snooze)

I believe the support for this would be covered in https://github.com/flutter/flutter/issues/3671 and help enable more fully featured apps. That thread covers other use cases that would require headless-Dart execution

mitchhymel:
@MaikuB 
I worked a bit on my own [notification plugin](https://github.com/mitchhymel/local_notifications) and while it's not fully featured as yours, I did get custom actions working with headless Dart in Android (at least from my testing of using my app daily for the past month, it seems to be working). 

[...]

PR #38, opened 2018-05-19
Update to include action buttons on android

pinkfish:
Redid the structure a bit to make it easier to include.  Plus added action button support on Android.


MaikuB:
First off all, thanks a lot for submitting a PR. It's good to see other members of the community trying to contribute. Unfortunately, I think I'll have to turn down your changes as both of your changes are things I had already started looking into. Originally I didn't move the Dart classes into lib/src, thinking in future there could be ambiguity if there were platforms with the same class names and the namespaces could be used to solve that issue.

With regards to see notification actions, if you looked at the issues board then you can see I'm already tracking this in #17 and that I'm waiting to see if members of the Flutter team can provide more information. Your changes only covers the scenario where actions launch an app but the actions should also be able to execute Dart code without having the app running. My testing shows that this is possible but it would seem like the APIs exposed by Flutter on Android are missing the capability to pass the payload back. Thus, I'm waiting to see if there the APIs will be changed to support this

pinkfish:
Yes, I ran into that same problem.  Not sure why it is in an int vs a long,
but I get the error at times when it does a callback.  The fix works :)

MaikuB:
Thanks for getting back to me :)
Actually I think your case might be bit different, I'd imagine your using numbers that requires more than 32 bits, a Dart int will become a Long on the Android side. The values used in the aforementioned issue should fit within an int

PR #51, opened 2018-06-03
added feature to put end time in repeated reminders

babanomania:
i wanted a feature to give me a notification every 1 minute till 30 minutes only. i,e  to start at a given time and run till an end time with a repeat interval. It can now done this like below. 
Unfortunately i didn't have xcode setup hence couldn't complete the code for the iOS part.

   var startTime = new DateTime.now().add(new Duration(seconds: 5));
    var endTime = new DateTime.now().add(new Duration(minutes: 30));

    var androidPlatformChannelSpecifics = 
              new NotificationDetailsAndroid(
                  'show minute channel id',
                  'show minute channel name',
                  'show minute description');

    var platformChannelSpecifics = new NotificationDetails(androidPlatformChannelSpecifics);

    await flutterLocalNotificationsPlugin.periodicallyShowTill(
        0,
        'show every minute title',
        'notification every minute',
        startTime,
        endTime,
        RepeatInterval.EveryMinute,
        platformChannelSpecifics
      );

### DIFF
--- a/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -64,6 +64,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
     public static String NOTIFICATION_ID = "notification_id";
     public static String NOTIFICATION = "notification";
     public static String REPEAT = "repeat";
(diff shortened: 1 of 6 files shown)


MaikuB:
Thanks for trying to submit a PR. However, if you looked at iOS APIs then you would see that it's not actually possible to do (e.g. https://developer.apple.com/documentation/usernotifications/untimeintervalnotificationtrigger). As such, I won't be merging this in. I'd advise looking at the APIs on both platforms first and/or opening an issue before trying to create a PR. Will add a note along these lines to the readme

Issue #61, opened 2018-06-19
Multiple Daily Notifications


JeffScaturro:
Thanks so much for writing this plugin, it's incredibly useful. Hopefully those final issues get cleared out with flutter soon so it can be officially adopted. 

Onto the not so great part, I have two different and independent notifications scheduled to be shown daily calling [`showDailyAtTime `](https://github.com/MaikuB/flutter_local_notifications#show-a-daily-notification-at-a-specific-time). One is shown in the morning, and the other in the evening. They both have unique notification ids and titles, while the body is the same.

If I schedule them to display relatively close to the current time, they will both execute and display. However the next day only the last one scheduled will display, which is the evening one in our case. The code execution is as follows:
- (*await*) cancel *all* existing notifications through plugin
- (*await*) schedule morning reminder 
- (*await*) schedule evening reminder

Have you seen anything similar or know of any current issues that may cause this? Thanks.

MaikuB:
Thanks for the compliments though I don't expect it to be an official plugin :)

Anyway, I've not heard of any such issues. Sounds like you're saying both notifications were showing and then one would disappear? If so, perhaps this is the OS doing it? I don't normally leave my notifications to the next day to see but there shouldn't be anything to do with the API that would result in the behaviour you're describing.

JeffScaturro:
Right, I'd schedule the morning & evening notifications, say around mid-day. I'd get the evening notification that night. The next morning there would not be a notification, however in the evening there would be one. I am testing this on an iOS device (iPhone) currently, so not sure if this is seen on Android yet. I wanted to check prior to diving too deep to see if it was familiar at all or not. I have some more testing planned for Tuesday & Wednesday and can update this issue with any findings and ultimately close it out. If any changes come up I'll be sure to drop a PR! 

Thanks again.

MaikuB:
Might be worth scheduling just one in the morning to see what happens but yeah if you find something wrong then a PR would be much appreciated :)

MaikuB:
Yeah I think you're onto something there. Tagging @javiercbk in case he may know too or have run into the same issue. I suspect what needs to be done is to change the components used as per the line code you linked to.

[...]

PR #64, opened 2018-06-21
[#61] iOS: Refactor date components used in repeat notification scheduling.

JeffScaturro:
### Issue #61 

Description
---
- Notifications scheduled for a time earlier in the day than the time the code was called resulted in invalid triggers (never getting fired).
- For instance, it's noon and you are scheduling a morning notification to be shown daily. These wouldn't fire because our trigger is initialized with today's date resulting in the impossibility to be executed.

What Was Changed
---
- Drop `NSCalendarUnitYear|NSCalendarUnitMonth|NSCalendarUnitDay` from the `NSDateComponents` we use for our `UNCalendarNotificationTrigger`. The trigger should ignore the year, month, and day and instead only be fired when it's the appropriate time and/or weekday.
- Specify the calendar on the `NSDateComponents` since we no longer init with components from a date and calendar.

Testing
---
- Regression tested scheduling daily notifications.
- Regression tested scheduling weekly at day and time notifications.
- Schedule a daily notification to be shown for a time earlier in the day; that is, schedule a notification to be shown daily at 8:00am if it's later than that already in the day when you schedule it with the plugin.

Little trick to test:
- [ ] Schedule a daily and weekly notification for an hour ago (and today's weekday). 
- [ ] Go into your iPhone's Settings -> General -> Date & Time. 
- [ ] Turn off `Set Automatically`.
- [ ] Move the time before each notification. As you drag the minute past it (and release) you should get your notification!

### DIFF
--- a/ios/Classes/FlutterLocalNotificationsPlugin.m
+++ b/ios/Classes/FlutterLocalNotificationsPlugin.m
@@ -276,13 +276,13 @@ typedef NS_ENUM(NSInteger, RepeatInterval) {
         if (notificationDetails.repeatTime != nil) {
             NSDate *now = [NSDate date];
             NSCalendar *calendar = [[NSCalendar alloc] initWithCalendarIdentifier: NSCalendarIdentifierGregorian];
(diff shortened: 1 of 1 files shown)


MaikuB:
Awesome, thanks! I'll publish this tomorrow

PR #79, opened 2018-07-15
Example fix: Added listview to the column to prevent overflowing

hussien4321:
This is my first time creating a pull request on a public repo so i hope this is the correct way to do so :)

While running the example code on my device, i noticed that the bottom of the page was having an issue rendering as the list of buttons was longer than the height of the screen, causing an overflow error. I have just added a fix for this by wrapping the column in a listview which allows the user to scroll if page overflows.

an screenshot of the bug:
screenshot_20180715-011256

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -45,129 +45,134 @@ class _MyAppState extends State<MyApp> {
         body: new Padding(
           padding: new EdgeInsets.all(8.0),
           child: new Center(
-            child: new Column(
+            child: new ListView(
+              scrollDirection: Axis.vertical,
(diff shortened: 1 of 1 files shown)


MaikuB:
Yep you've created the PR correctly. I had noticed the problem and was going to fix it soon and do have a solution on the way so I won't be taking the PR in but appreciate you taking the time to do so. Some general comments though:

- The ListView should've wrapped the Padding widget (i.e. it should've been set as the body of the scaffold). By having the Padding widget wrap the ListView, the scrollable area
- A SingleChildScrollView would've been more appropriate

PR #119, opened 2018-11-02
Add callback when a notification is shown, limit notification id and doc updates

MaikuB:


cdharris:
Awesome work, thank you! Will be great when the flutter team sorts out headless execution of dart code with access to plugins, from what I read, hopefully it's close.

MaikuB:
Thanks though I noticed an issue at the moment where it seems like having a custom Application custom in Android (required so that headless execution can fire other plugins), messes with handling when the user taps on a notification :(

Not sure if I've run into a bug with Flutter itself there

cdharris:
Ahh ok, appreciate the info, still some way to go here then. thanks for keeping this uptodate with Flutter's bleeding edge. Eagerly awaiting some of these features.

MaikuB:
No worries. I have been able to reproduce the issue with the geofencing example written by one of the Google engineers and raised an issue on this so unfortunately will have to hold off on releasing the update that includes the callback when a notification is shown. I may comment out the code related to the callback in the mean time and release the other changes that were in the PR

PR #138, opened 2018-12-17
Added dart, gradle and objective-c color code to README.md

riccardoratta:


MaikuB:
For my own benefit, could you explain more about this PR? From what I can gather, it would be around helping to format the sample code, however I don't see much difference in the output

riccardoratta:
It's just to color the code in the README so it's easier to read it. This change should not be visible in the pub.dart site but only here in github. This is some screenshot:

##### Before

screenshot 2018-12-19 at 13 05 39


##### After

screenshot 2018-12-19 at 13 48 05


PS: you have already done it in the README in the xml snippets. Also this [list](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml) contains the github  supported languages.

Issue #140, opened 2018-12-19
Scheduled notifications are immediately notified on the iOS 9.3.5 device


ganessaa:
We appreciate this wonderful plug-in.

But we have one problem.

When we register scheduled notifications on iOS 9.3.5 devices, they are notified immediately without waiting for the registration date and time.

Is this a specification?

We checked the following documents.
https://pub.dartlang.org/packages/flutter_local_notifications

We copied and pasted didReceiveLocalNotification and didFinishLaunchingWithOptions code into AppDelegate.m, but these problems were not solved.

Are we doing something wrong?
We intend to check all documents.
But since we are not English native speakers.
We may be missing something.

```
$ flutter doctor -v
[✓] Flutter (Channel dev, v1.1.1, on Mac OS X 10.13.6 17G4015, locale ja)
    • Flutter version 1.1.1 at /Applications/flutter
    • Framework revision 65df90d8b5 (7 days ago), 2018-12-11 19:29:13 -0800
    • Engine revision 25b58c0c67
    • Dart version 2.1.0 (build 2.1.0-dev.9.4 f9ebf21297)

[✓] Android toolchain - develop for Android devices (Android SDK 27.0.3)
    • Android SDK at /Users/xxxxx/src/android-sdks
    • Android NDK location not configured (optional; useful for native profiling support)
    • Platform android-27, build-tools 27.0.3
    • ANDROID_HOME = /Users/xxxxx/src/android-sdks
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1136-b06)
    • All Android licenses accepted.

[✓] iOS toolchain - develop for iOS devices (Xcode 10.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 10.1, Build version 10B61
    • ios-deploy 1.9.4
    • CocoaPods version 1.5.3

[✓] Android Studio (version 3.2)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 31.1.1
    • Dart plugin version 181.5656
    • Java version OpenJDK Runtime Environment (build 1.8.0_152-release-1136-b06)

[✓] VS Code (version 1.30.0)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 2.21.1

[✓] Connected device (1 available)
[...]

PR #141, opened 2018-12-19
#140 Fixed an issue that is immediately notified on devices before iOS 10

ganessaa:
We confirmed that the values change on both flutter side and Objective-C side. For that reason, we passed it as a character string from flutter and changed it to analyze as longLongValue.

We feel that we need to make similar modifications in `scheduledDate.millisecondsSinceEpoch` used for other uses.(We have not done it.)


ganessaa:
We can not remove flutter's `toString()`. The value is also truncated between flutter and Objective-C. We avoided it by passing that value as a string.

We can not remove codes on both flutter side and Objective-C side. Both of them are truncated.

(Perhaps the problem between flutter and Objective-C should be fixed without using strings in the original flutter project.)

MaikuB:
To be clear what I'm referring to is to remove the toString on Flutter but keep the objective C code to parse the value as a long. Has that been tried? I think the truncation is to do with how it use to parse it as an integer

ganessaa:
We can not erase toString() on Flutter side.
We already checked the following values.

flutter_local_notifications.dart
```
print(scheduledDate.millisecondsSinceEpoch);
```

FlutterLocalNotificationsPlugin.m
```
NSLog(@"%@" , call.arguments[MILLISECONDS_SINCE_EPOCH]);
NSLog(@"%ld" , [call.arguments[MILLISECONDS_SINCE_EPOCH] longLongValue] / 1000);
NSLog(@"%@" , @([call.arguments[MILLISECONDS_SINCE_EPOCH] longLongValue] / 1000));
```

Logs without toString()
```
1545269441429
```
```
-918785131
-918785
-918785
```

Logs with toString()
```
1545269975724
```
```
1545269975724
1545269975
1545269975
```

As we are not English native speakers, we may be misunderstanding your question.
Does this information satisfy you?

MaikuB:
That answers the question. Thanks

ganessaa:
=)

MaikuB:
Have you checked android? The changes made in Flutter to pass a string would mean changes are required on the Android side when parsing the values

ganessaa:
Ah! You are right.
We have to check on Android.
We will tell about result later.

ganessaa:
We were not used to creating plugins and our request did not work correctly on Android.
We withdraw the request.

PR #142, opened 2018-12-20
#140 Fixed a problem in which scheduled notifications are immediately notified on devices with pre-iOS 10 devices

ganessaa:
As we already told you ...

Fixed the problem that the value of `millisecondsSinceEpoch` is truncated between Flutter and Objective-C and in Objective-C code.

We changed that value to a string and fixed the problem that it is truncated between Flutter and Objective-C.

We also fixed a problem that would be truncated within Objective-C code using `longLongValue` instead of `integerValue`.

Along with that, on the Java side, we need to change this value from a character string to a long value, so we added that processing.

We confirmed that this change works correctly on iOS 10 and later devices, devices before iOS 10 and Android devices.

Perhaps this may be related to these.
https://github.com/flutter/flutter/issues/21313

We tried a method that does not use strings, but that could not be realized.

Also, it might be preferable to use Iso8601String for this string, but we did not adopt it because we were concerned about new bugs such as time zones.

### DIFF
--- a/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
+++ b/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
@@ -116,10 +116,12 @@ public class NotificationDetails {
         notificationDetails.title = (String) arguments.get(TITLE);
         notificationDetails.body = (String) arguments.get(BODY);
         if (arguments.containsKey(MILLISECONDS_SINCE_EPOCH)) {
-            notificationDetails.millisecondsSinceEpoch = (Long) arguments.get(MILLISECONDS_SINCE_EPOCH);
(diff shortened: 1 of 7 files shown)


MaikuB:
You might be right about how it is related to that issue. Did you think you could try using the plugin without the changes in this PR against the master branch/channel of Flutter?  See this link for how to switch channels https://github.com/flutter/flutter/wiki/Flutter-build-release-channels

ganessaa:
The problem between Flutter and Objective-C seems to have been improved by Flutter's master branch. We will create a simpler PR.

MaikuB:
I just wanted to ask you before you have a set of devices that can reproduce the issue. Do you run into the problem at all on the master channel/branch? If not, then once what is in master has been pushed out to stable. There would be no need for the PR.

Most developers will be on the stable channel/branch so if a PR made based on the master channel then it won't help. I hope that made sense

Issue #162, opened 2019-01-30
Move away from deprecated Support Library to AndroidX


vladaman:
Android Support Library is being deprecated and even flutter and firebase plugins requires you to migrate to AndroidX.

Currently with Flutter 1.1.8 and AndroidX migration I get following error:
```
/home/vlad/.pub-cache/hosted/pub.dartlang.org/flutter_local_notifications-0.4.5/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java:17: error: package android.support.annotation does not exist
import android.support.annotation.NonNull;
                                 ^
/home/vlad/.pub-cache/hosted/pub.dartlang.org/flutter_local_notifications-0.4.5/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java:139: error: cannot find symbol
    @NonNull
     ^
  symbol:   class NonNull
  location: class FlutterLocalNotificationsPlugin
2 errors
```

MaikuB:
Thanks, I'm aware of this but there were open issues around migrating to AndroidX that from what I've seen, have only been recently resolved

JeffScaturro:
@MaikuB, given those are now resolved, do you have any more hesitation in migrating? If not, are you currently working on a branch with support for AndroidX? We have a flutter app that we're releasing shortly and were planning on migrating to X. Tonight I was going to look into what it would take to get this plugin migrated, but I haven't been following AndroidX much.

MaikuB:
I don't know if migration will break devs on stable, which is what I use and would imagine most are devs are on. I believe additional effort is required on devs to migrate their apps to Android and is supposedly covered in 

https://developer.android.com/jetpack/androidx/migrate

Based on what I've seen/read (https://github.com/flutter/flutter/issues/27106 and https://github.com/flutter/flutter/wiki/Postmortem:-AndroidX-plugin-migration), the tooling on stable hasn't been updated so that new apps support AndroidX so migration of apps will be required.

Currently got no branch to migrate the plugin but I believe the migration process for the plugin should be straightforward

MaikuB:
Thanks I'll also do some testing as well when I get a chance to, particularly on a new app by calling `flutter create ...` on the stable channel though I imagine you'll also test that

MaikuB:
Have migrated to AndroidX in version 0.5.0 so will close this. Thanks again @JeffScaturro for the PR :)

Issue #159, opened 2019-01-22
Notification.MessagingStyle support on Android


Hixie:
https://developer.android.com/reference/android/app/Notification.MessagingStyle

MaikuB:
I assume that you've linked [Notification.MessagingStyle](https://developer.android.com/reference/android/app/Notification.MessagingStyle) to instead of [NotificationCompat.MessagingStyle](https://developer.android.com/reference/android/support/v4/app/NotificationCompat.MessagingStyle) to provide richer notifications?

If so, it would sound like, and correct me if I'm wrong, for API 28+ [Notification.MessagingStyle](https://developer.android.com/reference/android/app/Notification.MessagingStyle) should be used. This would in turn mean that plugin will need to switch to using [Notification.Builder](https://developer.android.com/reference/android/app/Notification.Builder) instead of [NotificationCompat.Builder](https://developer.android.com/reference/android/support/v4/app/NotificationCompat.Builder)

For API < 27, can use [NotificationCompat.MessagingStyle](https://developer.android.com/reference/android/support/v4/app/NotificationCompat.MessagingStyle). The API surface of the plugin should still provide the means to specify the details of each [person](https://developer.android.com/reference/android/app/Person.Builder)

Hixie:
I have no strong opinion, I just noticed that this style wasn't offered by the plugin but is listed in the Android documentation.

MaikuB:
Noted, this is partly due to not being able to version 28 of the support library before as apparently other devs had pointed out there was an [issue](https://github.com/flutter/flutter/issues/19644) with Flutter when it came to using that version of the library. Though I can see you've said you've believed it's fixed.

Was also hoping to migrate to AndroidX but from what I gather, there are open issues (e.g. https://github.com/flutter/flutter/issues/23995) on that so guess will have to defer that to later

MaikuB:
Given the plugin has been migrated to AndroidX, it looks like there's no need to be concerned with branching off the code anymore. AndroidX has the Person class that was appeared to be missing in the support library

MaikuB:
Added in v0.5.1. Let me know if you've got any feedback :)

MaikuB:
Closing this. Please feel free to reopen if you feel there's something missing

PR #163, opened 2019-01-31
Migrating to AndroidX.

jeffscaturro-aka:
**Issue:** "AndroidX is a major improvement to the original Android Support Library. Like the Support Library, AndroidX ships separately from the Android OS and provides backwards-compatibility across Android releases. AndroidX fully replaces the Support Library by providing feature parity and new libraries." _(https://developer.android.com/jetpack/androidx/)_

**What was changed:** drop the deprecated support library and migrate to Android X. Changes made referencing the [AndroidX migration guide](https://developer.android.com/jetpack/androidx/).

**Testing:** full regression testing on Android devices.

**Testing results:** *coming soon*

**Related issues:** [ #159, #162 ]

### DIFF
--- a/example/android/gradle/wrapper/gradle-wrapper.properties
+++ b/example/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Sat Nov 10 09:17:46 AEDT 2018
+#Thu Jan 31 21:56:59 AEDT 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.6-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.1-all.zip


MaikuB:
Thanks for the PR. There are some additional changes that I believe should go in. More specifically, I'm expecting changes in the gradle.properties file (probably not needed but better to be safe) and build.gradle for the example app as well (compile and target SDK should be 28). Take a look at the guide at https://developer.android.com/jetpack/androidx/migrate and suggest taking a look at the changes the Flutter team did to update the first party plugins. I think if you used Android Studio to handle the migration then it would've done all of the work. Not if you had made the changes manually as what I was trying to do a while back was to open the example app in Android Studio and use the migration option under the refactor menu once it had been opened

note: I was starting on this earlier but can't continue until much later so happy to take a PR :)

JeffScaturro:
Scheduled notifications are working on Galaxy S8 device (`SM G950U1`) running 8.0.0.

For sanity and full regression testing, iOS notifications still work (iPhone X - 12.1.3).

JeffScaturro:
I am done testing as well. Got more notifications this morning.

PR #171, opened 2019-02-15
Support multi plugins for background operation

xtelinco:
Hi,

When you are running this plugin with multiple isolates, the use of a static method channel means that the first instance of the plugin for one of the isolates get over written by the instance for the new isolate.

Hence you can't send notification intents to both dart isolates. Moving the method channel within the plugin gets around this issue.

Not sure if it causes other issues however. I'm not fully sure I understand why you created it as static. There may be a very good reason!

Here is the PR.

Cheers


MaikuB:
Thanks for the PR. There wasn't a particular reason. Will have to test to see the effects of this change but i don't think it'll cause an issue

PR #185, opened 2019-03-17
feat(): Override toString() on Time

Simpler1:
This pull request adds the override for toString() on the time class so that it automatically pads the hours, minutes, and seconds.  For a time of 8 minutes after 3 am, `print($time)` will produce `03:08:00`.

This will reduce the formatting in the existing example file.

For example:
`        'Weekly notification shown on Monday at approximately ${_toTwoDigitString(time.hour)}:${_toTwoDigitString(time.minute)}:${_toTwoDigitString(time.second)}',
`

can be rewritten as:
`        'Weekly notification shown on Monday at approximately $time',
`

Thanks

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -578,7 +578,7 @@ class _HomePageState extends State<HomePage> {
     await flutterLocalNotificationsPlugin.showDailyAtTime(
         0,
         'show daily title',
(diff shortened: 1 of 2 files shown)


MaikuB:
Sorry but won't be taking in this PR. It essentially changes the functionality of a core class in the library just to refactor code that is specific to the example app. I'd rather leave the presentation up to the consumers as some might want to display time in a different format

PR #228, opened 2019-05-11
add ticker support

byrdkm17:


lib/src/flutter_local_notifications.dart
```diff
@@ -122,7 +122,8 @@ class FlutterLocalNotificationsPlugin {
       'title': title,
       'body': body,
       'platformSpecifics': serializedPlatformSpecifics,
-      'payload': payload ?? ''
+      'payload': payload ?? '',
+      'ticker' : ticker
```
MaikuB: given this is an Android-specific setting, instead of exposing it as a parameter, could you move this setting to the Android-specific _AndroidNotificationDetails_ class? this will making more changes as a result of this

flutter_local_notifications.iml
```diff
@@ -4,14 +4,15 @@
...
+      <excludeFolder url="file://$MODULE_DIR$/.dart_tool" />
       <excludeFolder url="file://$MODULE_DIR$/.idea" />
       <excludeFolder url="file://$MODULE_DIR$/.pub" />
       <excludeFolder url="file://$MODULE_DIR$/build" />
+      <excludeFolder url="file://$MODULE_DIR$/example/.dart_tool" />
       <excludeFolder url="file://$MODULE_DIR$/example/.pub" />
       <excludeFolder url="file://$MODULE_DIR$/example/build" />
     </content>
     <orderEntry type="sourceFolder" forTests="false" />
-    <orderEntry type="library" name="Dart Packages" level="project" />
```
MaikuB: any reason on why this was removed?

lib/src/flutter_local_notifications.dart
MaikuB: also upon moving it to the _AndroidNotificationDetails_ class, could you add some documentation on the new field? could almost be a copy and paste of what's written in the Android docs

flutter_local_notifications.iml
byrdkm17: this is auto modify by IDE, i will fixed it

MaikuB:
i need to find some time to test it out even though it looks fine

MaikuB:
merged in, thanks. will try to get this out later today

PR #229, opened 2019-05-11
Fix typo with 'recieve'

wanbok:
There is no more 'recieve' finally. 😂


MaikuB:
doh looks like i missed one, lol. thanks

Issue #216, opened 2019-04-19
We need an ability to optionally use setExactAndAllowWhileIdle for Android


alexandr2levin:
The name speaks for itself :D

MaikuB:
This has been noted but has some extra implications when it comes to recurring notifications as that would require scheduling the next notification after one has been fired. Happy to look at a PR for this if anyone wants to take a stab at it

MaikuB:
Closing this should be supported in 0.8.0

PR #239, opened 2019-05-27
Implement exact and allow while idle options for android

alexandr2levin:
It isn't final implementation. @MaikuB, do you like the idea?

### DIFF
--- a/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -16,8 +16,10 @@ import android.net.Uri;
 import android.os.Build;
 import android.text.Html;
 import android.text.Spanned;
+import android.util.Log;
 
 import androidx.annotation.NonNull;
+import androidx.core.app.AlarmManagerCompat;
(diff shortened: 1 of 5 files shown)


alexandr2levin:
https://github.com/MaikuB/flutter_local_notifications/issues/216

MaikuB:
Thanks for attempting this. Just a note first that you should be able to create a draft PR to discuss with others.

My thoughts on this were to create something like a `ScheduledNotificationDetails` class that extends from the Android-specific `NotificationDetails` one and it would have a property that describes the behaviour. This could be an enum called something like `ScheduledNotificationTiming` with the following possible values


- Exact
- ExactAndAllowWhileIdle (this should probably be the default)
- Inexact
- InexactAndAllowWhileIdle


You currently have two booleans, which means one can be toggled without the other on but they are related to each other. The enum would help depict that relationship a lot more clearer, clarify what the behaviour and make it easier for devs to use.

The schedule method would be updated to take an instance of the `ScheduledNotificationDetails` class. That way the `repeat` method can be left as it was before and can be revisited later

MaikuB:
Do you plan to make more changes based on my reply?

MaikuB:
Closing this due to lack of further response and will use #256 to continue work on this. Thanks again for attempting this though

PR #249, opened 2019-06-14
PT-2060 Add Period and Ovulation reminders setting

levchukAY:
https://youtrack.digitalchemy.us/youtrack/issue/PT-2060


MaikuB:
Thanks for the PR. A few of things

- The PR's title and description on what the changes are about
- None of the plugins in Flutter's ecosystem that I've seen need to deal with wrapping the calls with an AsyncTask on the Android side so I don't see why you feel there's a need to do so for this particular plugin. Calling a plugin's method is done asynchronously already (see https://flutter.dev/docs/development/platform-integration/platform-channels)

In short, I'm not sure there's much to gained as it looks its wrapping asynchronous work behind another asynchronous operation. If the problem you're trying to solve is to execute plugin methods in the background in your app then that IMO should be something looked at on the Flutter side e.g. via isolates.

PR #256, opened 2019-06-25
Implement precision on ScheduledAndroidNotificationDetails

lucaslcode:
#216 and other pull request #239 

I tried to implement with a new class as you mentioned in that thread.

I wanted to avoid having 'exact' as the default, because the whole point of the change to inexact was so Android can manage wakelocks better etc.
I also wanted to avoid a breaking change, so I made it fall back to 'inexact' if not specified, but that basically does make it a breaking-ish change, because the old behavior was exact.

In extending the AndroidNotificationDetails class, I had to re-specify all the defaults, which seems not so good. I don't know if there's a better way to do it.

Finally, it would be good to have an enum in the java side like the dart, but it ended up converting between ints and the enum and it was a bit complicated.

### DIFF
--- a/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
+++ b/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
@@ -97,6 +97,8 @@ public class NotificationDetails {
     public static final String BODY = "body";
 
     public static final String TICKER = "ticker";
+    
+    public static final String PRECISION = "precision";
 
     public Integer id;
     public String title;
@@ -139,6 +141,7 @@ public class NotificationDetails {
     public Integer ledOnMs;
     public Integer ledOffMs;
     public String ticker;
+    public Integer precision;
 
 
     // Note: this is set on the Android to save details about the icon that should be used when re-hydrating scheduled notifications when a device has been restarted.
@@ -208,6 +211,9 @@ public class NotificationDetails {
                 }
             }
             notificationDetails.ticker = (String) platformChannelSpecifics.get(TICKER);
+            if (platformChannelSpecifics.containsKey(PRECISION)) {
+                notificationDetails.precision = (Integer) platformChannelSpecifics.get(PRECISION);
+            }
         }
         return notificationDetails;
     }
--- a/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
+++ b/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
@@ -6,7 +6,6 @@ import android.os.Build;
 import com.dexterous.flutterlocalnotifications.BitmapSource;
 import com.dexterous.flutterlocalnotifications.NotificationStyle;
 import com.dexterous.flutterlocalnotifications.RepeatInterval;
-import com.dexterous.flutterlocalnotifications.ScheduledNotificationPrecision;
 import com.dexterous.flutterlocalnotifications.models.styles.BigPictureStyleInformation;
 import com.dexterous.flutterlocalnotifications.models.styles.BigTextStyleInformation;
 import com.dexterous.flutterlocalnotifications.models.styles.DefaultStyleInformation;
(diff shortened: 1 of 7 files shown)


MaikuB:
Appreciate if you could let me know if plan to continue working on this as well

lucaslcode:
Yes I do plan on getting around to it soon! Agree with everything you
recommended.

lucaslcode:
CodeFactor is failing because there's no default case, any recommendation for that?

android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
```diff
@@ -97,6 +98,8 @@
     public static final String BODY = "body";
 
     public static final String TICKER = "ticker";
+    
+    public static final String PRECISION = "scheduledAndroidNotificationPrecision";
```
lucaslcode: I opted to use 'scheduledNotificationPrecision' only (except here for the key), since we're in android platform anyway

lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -132,7 +137,8 @@ class AndroidNotificationDetails {
       this.ledColor,
       this.ledOnMs,
       this.ledOffMs,
-      this.ticker});
+      this.ticker,
+      this.scheduledAndroidNotificationPrecision = ScheduledAndroidNotificationPrecision.Inexact});
```
lucaslcode: Do you think it might be better to set the default as Exact for a minor version increment, then change it to Inexact for a major increment?

MaikuB:
For the default case, I would go with making it do the same as inexact. That's the current behaviour and should be the path that would execute for older notifications that have been saved to shared preferences as the precision would be null

android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
MaikuB: That's fine though it I'm still seeing Android here

lib/src/platform_specifics/android/notification_details.dart
MaikuB: Should be the other way around if exact was the default as that's a breaking change in behaviour and thus should be a major version change. I'd be inclined to leave it as inexact unless all the methods for setting up recurring notifications can use the new enum. Reason being is that without that change at least the Android behaviour is consistent. If the default was to be so that it's consistent with iOS then if I'm not mistaken, it should be exact and allowed while idle

MaikuB: Hmm just realised I may have confused myself and as a result caused some confusion whilst looking at this PR, so apologies first about that. If you look at the original code I had it actually was for exact timing. If the call to `set` was done in API version > 19 (i.e. beyond KitKat) then it would've been inexact. However, there were API version checks that make sure it calls exact. The implementation of `AlarmManagerCompat.setExact` does the same thing as the code I originally had. So you're right in that the default behaviour as it stands right now is exact. Though this means the default value should also be set to exact if backwards compatibility is to be maintained

MaikuB: Following on from that point, I think we could remove the inexact option as it calls `AlarmManager.set`, which will be exact on older Android versions

MaikuB:
In case you miss my other comments, just a correction that the default case should be exact

MaikuB:
Just checking in to see if you'll be looking to do more changes on this. If you are then I can wait, otherwise I can look to commit directly to this PR

lucaslcode:
I may not have time for a few weeks - feel free!

MaikuB:
Made changes and merged now. Thanks!

PeterZainzinger:
@MaikuB  do you know how to pass `ScheduledAndroidNotificationPrecision` - does not seem to be used anywhere :thinking: ?

MaikuB:
Ah thanks for picking that up, it needs to be removed as it's no longer used.

PeterZainzinger:
Ok, that's what I thought. Is there now a way to specify to precision at all?

MaikuB:
If you look at the PR and changelog, you should see that there's a boolean flag (`allowWhileIdle`) for this

PeterZainzinger:
Ok, thanks for the hint :+1: !

PR #277, opened 2019-07-30
Payload as map

athornz:
I've implemented these changes so I can receive both Firebase messages and local notifications with the onSelected callback in my app.

This PR is to open a discussion around these changes and whether it might be something that should be merged.

Basically, because of the known issue around not being able to have multiple notification delegates on iOS, `local notifications` overrides Firebase Messaging callbacks.

When clicking on a Firebase notification, local notifications still gets invoked, but it only extracts and passes the payload back to the onSelected callback. A Firebase notification doesn't have a payload so the payload is null.

By passing back the userInfo dictionary, we can extract the payload on the dart side and add an extra parameter to onSelected which is all the (firebase) data from the notification - it allows the user to handle both Firebase and local notifications.

Android doesn't have the same problem (I think) so the payload is just passed back but within a map.

### DIFF
--- a/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -14,6 +14,7 @@ import android.media.AudioAttributes;
 import android.media.RingtoneManager;
 import android.net.Uri;
 import android.os.Build;
+import android.os.Bundle;
 import android.text.Html;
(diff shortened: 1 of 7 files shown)


MaikuB:
FYI that one of the members of the Flutter team is looking into the issue and in fact has a PR that pending approval. As such, I don't think this PR is necessary

athornz:
Great! Do you have a link?

MaikuB:
https://github.com/flutter/engine/pull/9864

It's referenced in the issue I created on the Flutter repository around this

sante85:
please reopen

Issue #278, opened 2019-07-30
Conflicts when using plugin from both background isolate and foreground


athornz:
I'm using a headless flutter view to run dart code in the background. I've found when registering this plugin on the background runner, trying to use it in the foreground no longer works.

One of the reasons for this seems to be that a lot of the plugins state variables are static rather than instance variables.

MaikuB:
Which state are you referring to that is static that you think is causing the issue? Unless I've missed something, the majority of the static members are just constants and don't actually carry state. If you have a solution for the problem then a PR is welcome though I'm wondering if you're running into the issue I've mentioned here. Bit of a different case as I was attempting headless execution but that also uses a different isolate

https://github.com/flutter/flutter/issues/23904

Perhaps you'll need to look at messaging passing as well though it's not something I've really looked into

athornz:
These variables are the ones I'm referring to:

```
FlutterMethodChannel* channel;
bool appResumingFromBackground;
bool displayAlert;
bool playSound;
bool updateBadge;
bool initialized;
NSUserDefaults *persistentState;
NSObject<FlutterPluginRegistrar> *_registrar;
```

In particular, the channel will get overwritten as soon as this plugin is registered with a different registrar.

MaikuB:
Thanks will try to look at your PR later. Just checking though, it solves the problem you're having?

athornz:
Not completely, I think the remaining issues will be solved once the Flutter PR you mentioned is merged.

MaikuB:
The PR I mentioned before (in a PR you tried to submit) is around delegates though so seems separate from the issue you're having. Did you mean to refer to the issue I've linked around dealing with platform channels and isolates?

athornz:
Yes the delegate PR: https://github.com/flutter/engine/pull/9864

So basically, this PR fixes my issue with registering FlutterLocalNotifications in a background isolate preventing the foreground isolate from working.

It's not a complete fix though as FlutterLocalNotifications can't be used from the background isolate. 

It can be registered (via GeneratedPluginRegistrant), but not used. 

If `initialize` is called on the background isolate, the UNNotificationCenter delegate gets set to the background instance of FlutterLocalNotifications, and the foreground instance will no longer work.

Once the Flutter PR is merged though, multiple plugins should be able to be delegates for UNNotificationCenter and this should solve the issue.

Does that sound right?

MaikuB:
[...]

PR #280, opened 2019-07-30
Instance variables

athornz:
Related: #278 

This PR moves class variables to instance variables to address issues when this plugin is registered multiple times.

Previously, when the plugin is registered a second time, the channel, registrar, and UNUserNotificationCenter delegate would be recreated/set to the new plugin. As these variables were class variables this would affect all instances of the plugin.

### DIFF
--- a/ios/Classes/FlutterLocalNotificationsPlugin.m
+++ b/ios/Classes/FlutterLocalNotificationsPlugin.m
@@ -6,7 +6,6 @@
 
 @implementation FlutterLocalNotificationsPlugin{
     FlutterMethodChannel* _channel;
-    bool appResumingFromBackground;
     bool displayAlert;
     bool playSound;
     bool updateBadge;
@@ -87,10 +86,9 @@ typedef NS_ENUM(NSInteger, RepeatInterval) {
     if (self) {
         _channel = channel;
         _registrar = registrar;
-        appResumingFromBackground = YES;
-        
         persistentState = [NSUserDefaults standardUserDefaults];
     }
+
     return self;
 }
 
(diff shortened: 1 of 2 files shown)


MaikuB:
Thanks for the PR. Took this opportunity to clean up some code I had left behind as well. iOS build seems to be stalling since my commit but should be fine so will merge it

ios/Classes/FlutterLocalNotificationsPlugin.m
```diff
@@ -141,14 +146,11 @@ - (void)initialize:(FlutterMethodCall * _Nonnull)call result:(FlutterResult _Non
...
-    /*if (call.arguments[ON_NOTIFICATION_CALLBACK_DISPATCHER] != [NSNull null]) {
-        [self startHeadlessService:[call.arguments[CALLBACK_DISPATCHER] longValue]];
-        [self setCallbackDispatcherHandle:[call.arguments[ON_NOTIFICATION_CALLBACK_DISPATCHER] longValue] key:ON_NOTIFICATION_CALLBACK_DISPATCHER];
-    } else {
-        [persistentState removeObjectForKey:ON_NOTIFICATION_CALLBACK_DISPATCHER];
-    }*/
+    
     if(@available(iOS 10.0, *)) {
         UNUserNotificationCenter *center = [UNUserNotificationCenter currentNotificationCenter];
+        center.delegate = self;
```
MaikuB: FYI i'm reverting this change unless you have any other suggestions as it looks like it results in late registration that has led to #295 where the plugin can't handle where the notification has launched the app

Issue #50, opened 2018-05-31
flutter_local_notifications conflict with the firebase_messaging 1.0.1


MandyKuo8:
The firebase_messaging doesn't work properly in iOS platform when we apply  firebase_messaging plugin and flutter_local_notifications plugin together, but it works fine in Android platform.

Without  the firebase_messaging plugin, our app works fine and can receive all of the callbacks (onMessage, onLaunch, onResume) from firebase_messaging.
Unfortunately, the callbacks don't be called once we apply the  flutter_local_notifications plugin. 
_firebaseMessaging.configure(
    onMessage: onMessage, onLaunch: onLaunch, onResume: onResume);
Anyone can help to look into this issue?
Here is the firebase_messaging link.
https://pub.dartlang.org/packages/firebase_messaging

MaikuB:
Sorry hadn't realised you also posted the issue but if you could try out what I suggested in my email against the dev branch of this plugin then that would be good. As mentioned in my reply to you, I suspect it's to do with both plugins registering for handling incoming notifications and my last commit to dev will hopefully provide a way to prevent that on iOS as that should be done by Firebase.

MandyKuo8:
Hi Michael,
The solution you provide works. Would you please apply this patch and publish the latest package to Flutter? Thank you.

MaikuB:
I need to do more testing first. You should also be aware that I believe that by having the new setting turned off means you won't be able to use my plugin to display local notifications on iOS. Not sure if that will be a deal breaker for your app but I don't think there's a way around. Partly why I said your team may need to write their own code to handle notifications

MaikuB:
[...]

PR #283, opened 2019-07-31
Added the ability to enable/disable the delegate registration in NotificationCenter via iOS Settings object

olivierbrand:
I have added a possible way to enable/disable the delegate registration to notification center. As I am using firebase messaging, my delegate was getting overridden as you explain in the README, so instead of hacking the code, I thought it would be more elegant to do it this way. By default the value is true, so it does not impact existing code base


MaikuB:
Thanks for the PR. This is actually something I did last year in https://github.com/MaikuB/flutter_local_notifications/issues/50 but then didn't merge in (reasons in the thread). At this stage I think I think I'll probably not take this PR as it's still a workaroundand there's been some movement on the main Flutter repo to try to see if it can be addressed so may end up being reverted as well. If this issue doesn't get addressed in the Flutter engine then I'll likely reinstate the changes I did last year

olivierbrand:
Maiku, thanks. As we do not have a date or even know if this issue will be fixed, why not enabled this now? 
Also, we are enabling custom sound for our remote push notifications BUT the firebase plugin does not allow (to my knowledge) creating android notification channels. I have looked at your code and the channel creation is embedded on your side in the local notification show method. I have changed your plugin to expose a createNotificationChannel that does not force sending a notification just for the sake of creating a channel. Do you think you would enable such feature or do you think, that once again firebase plugin should fix this?

PR #299, opened 2019-08-25
add custom day interval option

DimiDeKerf:
This PR adds the ability to set a custom day interval when a notification should be shown at a specified time. This allows the user to set the day interval on a specific time of the notifications, f.e. every other day at 10:00, every 30 days, every 365 days...

The different periodically notifications current available are great, but lack support for this situation. It would be really helpful to have a more dynamic option available.

### DIFF
--- a/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -258,6 +258,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
 
         AlarmManager alarmManager = getAlarmManager(context);
         long repeatInterval = 0;
+        int dayInterval = notificationDetails.dayInterval == null ? 1 : notificationDetails.dayInterval;
         switch (notificationDetails.repeatInterval) {
             case EveryMinute:
                 repeatInterval = 60000;
@@ -266,7 +267,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
                 repeatInterval = 60000 * 60;
                 break;
             case Daily:
-                repeatInterval = 60000 * 60 * 24 * notificationDetails.dayInterval;
+                repeatInterval = 60000 * 60 * 24 * dayInterval;
                 break;
             case Weekly:
                 repeatInterval = 60000 * 60 * 24 * 7;
--- a/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -258,7 +258,6 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
 
         AlarmManager alarmManager = getAlarmManager(context);
         long repeatInterval = 0;
-        int dayInterval = notificationDetails.dayInterval == null ? 1 : notificationDetails.dayInterval;
         switch (notificationDetails.repeatInterval) {
             case EveryMinute:
                 repeatInterval = 60000;
@@ -267,7 +266,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
                 repeatInterval = 60000 * 60;
                 break;
             case Daily:
-                repeatInterval = 60000 * 60 * 24 * dayInterval;
+                repeatInterval = 60000 * 60 * 24 * (notificationDetails.dayInterval == null ? 1 : notificationDetails.dayInterval);
                 break;
             case Weekly:
                 repeatInterval = 60000 * 60 * 24 * 7;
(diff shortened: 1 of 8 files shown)


MaikuB:
Thanks for the PR. I'll try to take a closer within the next few days

DimiDeKerf:
Thanks! Great plugin btw, it is really useful!

luke1313:
i would love to have this merged into master aswell! wanted to do the fork myself today

lib/src/flutter_local_notifications.dart
```diff
@@ -217,6 +223,32 @@ class FlutterLocalNotificationsPlugin {
     });
   }
 
+  /// Shows a notification on the given day interval at the specified time
+  Future<void> showCustomDayIntervalAtTime(
```
MaikuB: What's the reason behind having another method? If I'm not mistaken, it looks like an optional parameter for the interval in could've been added to `showDailyAtTime` method that would have a default value of zero for backwards compatibility. I'm guessing it was to avoid making changes to the existing method but IMO it's be cleaner to modify it and could potential do similar changes for the other methods.

luke1313: i actually like to have another method because it won't be daily - it will be in two or three days. the daily method could call this method with 1 though and weekly with 7 to make the code cleaner for future development

MaikuB: Yep I understand it's not every day. What I'm suggesting here is identical to how recurrence rules are configured for calendar events (see https://tools.ietf.org/html/rfc5545#page-40 in case you're not familiar). So daily here is just the type of frequency

DimiDeKerf: Yeah I was thinking the same way. But I didn't want to adjust the existing methods like you've mentioned and I thought that the `showDailyAtTime` would become more of a shorthand to have daily notifications.
I'll update the PR with an optional parameter that will default to 1 day, thanks guys for your feedback!

android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -266,7 +266,7 @@ private static void repeatNotification(Context context, NotificationDetails noti
                 repeatInterval = 60000 * 60;
                 break;
             case Daily:
-                repeatInterval = 60000 * 60 * 24;
+                repeatInterval = 60000 * 60 * 24 * notificationDetails.dayInterval;
```
MaikuB: `notificationDetails.dayInterval` will be null for notifications that had been scheduled before these changes were made so will be need be treated as though it's been set to 1

DimiDeKerf: Oh indeed! I'll set the dayInterval to 1 when notificationDetails.dayInterval is null, thanks for spotting this one!

android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -266,7 +267,7 @@ private static void repeatNotification(Context context, NotificationDetails noti
                 repeatInterval = 60000 * 60;
                 break;
             case Daily:
-                repeatInterval = 60000 * 60 * 24;
+                repeatInterval = 60000 * 60 * 24 * dayInterval;
```
MaikuB: A couple of refactoring/coding tips

1. `dayInterval` is only needed here so it's definition should be moved here to be more clear about its scope
2. having said that, it doesn't need to be declared and could be part of the expression for calculating the value of `repeatInterval` (see [here](https://refactoring.com/catalog/inlineVariable.html) where you may find a lot of useful refactoring techniques)

ios/Classes/FlutterLocalNotificationsPlugin.m
```diff
@@ -399,7 +404,7 @@ - (void) showLocalNotification:(NotificationDetails *) notificationDetails {
     if(notificationDetails.secondsSinceEpoch == nil) {
         if(notificationDetails.repeatInterval != nil) {
             NSTimeInterval timeInterval = 0;
-            
+            NSInteger dayInterval = notificationDetails.dayInterval == nil ? 1 : [notificationDetails.dayInterval integerValue];
```
MaikuB: similar comments to what I made about the Android change. Just a note that the issue I mentioned was more of a concern on Android where the plugin needs to handle rescheduling notifications when the phone is restarted

android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
DimiDeKerf: Thanks for your feedback! I instantiated the `dayInterval` upfront to make it more clear and to keep the calculation of the `repeatInterval` easier to digest. But its scope is indeed clearer when moved to the Daily state. I've updated both the Android and iOS implementation.

MaikuB:
Having taken a closer a look at this I don't think this is going to work. The iOS APIs require matching parts of the date/time but won't allow specifying the interval. In fact none of the iOS changes have actually done anything as it the notification is still shown daily

PR #311, opened 2019-09-07
Add support for delayed notification permission request on iOS

dluksza:
Currently notification permissions can only be requested when plugin is
initialized. In some cases this may result in suboptimal user
experience, when user will be asked for the notification permission
before he would event decide if he want to be notified by application.

This patch adds 'requestPermissions' method that may be used later on
for asking user to grant application notification permission. This way
all the notificaiton handlers can be initialized immediately after app
startup. Then later on, when user actually decide to use local
notifications 'requestPermissions' method will be used to as him for
required permissions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,42 @@ await flutterLocalNotificationsPlugin.show(
     3, 'Attention', 'Two new messages', platformChannelSpecifics);
 ```
 
+### [iOS only] Requesting notification permission
+
+By default this plugin will request notification permission when it is initialized. `IOSInitializationSettings` have three named parameters:
+1. `requestSoundPermission`,
+1. `requestBadgePermission`,
+1. `requestAlertPermission`
+that control this behaviour.
+
+If you want to delay permission request, set all of the above to false, then later call `requestPermissions` method.
+
+```dart
+FlutterLocalNotificationsPlugin flutterLocalNotificationsPlugin = new FlutterLocalNotificationsPlugin();
+var initializationSettingsAndroid =
+    new AndroidInitializationSettings('app_icon');
+var initializationSettingsIOS = new IOSInitializationSettings(
+        requestSoundPermission: false,
+        requestBadgePermission: false,
+        requestAlertPermission: false,
+        onDidReceiveLocalNotification: onDidReceiveLocalNotification,
+    );
+var initializationSettings = new InitializationSettings(
+    initializationSettingsAndroid, initializationSettingsIOS);
+flutterLocalNotificationsPlugin.initialize(initializationSettings,
+    onSelectNotification: onSelectNotification);
+
+/* some logic */
+
+var result = await flutterLocalNotificationsPlugin.requestPermissions(
+        requestSoundPermission: false,
+        requestBadgePermission: false,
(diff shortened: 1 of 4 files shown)


MaikuB:
Not really clear what problem this PR is trying to solve here but it seems like you want to separate the requesting of permissions from the initialisation. If that's the case, this isn't a change I'm looking to take in at this stage. The changes appear to involve having a platform-specific method that I'm intentionally trying to avoid with this plugin. Furthermore, it looks like `requestPermissions` needs to be call twice, which seems awkward with regards to developer experience. 

If you need to delay showing the permissions prompt then I would suggest you maintain your own fork or have your own code for this for reasons explained above. I should also clarify that I'm not trying to cover all possible scenarios that will work for everyone as that would be difficult to do whilst maintaining an API that avoids platform-specific methods. Also want to point out that Apple's own documentation states to that the permissions prompt is to be shown during launch https://developer.apple.com/documentation/usernotifications/asking_permission_to_use_notifications. I've yet to see apps that haven't followed this on iOS. If there are any, I suspect they're the exception not the norm

kaciula:
@MaikuB Please consider adding this feature to the plugin. In the link you've added, it is stated very clearly that Apple recommends requesting the permission at the right time, and NOT during launch. It's the opposite of what you affirmed. :)

Please re-think your decision. Adding a platform-specific method to request a permission is not the end of the world. :)

> Make the request in a context that helps the user to understand why your app needs authorization. For example, for a task-tracking app that sends reminder notifications, you could make the request after the user schedules a first task. This provides a better user experience than automatically requesting authorization on first launch, because the user can more easily see what purpose the notifications serve.

MaikuB:
@kaciula that's not what it said back when I posted that link so looks like the guidance has changed. You can submit a pull request to add the feature given there are now platform-specific implementations of the plugin

MaikuB:
This is the iOS implementation of the platform interface https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/IOSFlutterLocalNotificationsPlugin-class.html

If a PR is to be submitted, it can added to there. It can be accessed via [this property](https://pub.dev/documentation/flutter_local_notifications_platform_interface/latest/flutter_local_notifications_platform_interface/FlutterLocalNotificationsPlatform/instance.html) when running on iOS. This allows for platform-specific methods to be in platform-specific implementations of the plugin. The `FlutterLocalNotificationsPlugin` class will be more for cross-platform/common functionality. It may or may not need to have to `initialize` method tweaked so that it`s possible to avoid it triggering the permissions prompt

erolasan:
I need this as well. In my case, I am first showing the user the app onboarding then explain why the app needs permission and then ask for it. So this would be a very useful feature to have :)

dluksza:
@MaikuB I can reopen this PR (adjusted to your liking) if you are now open to merge this is.

I'd really like to get rid of my fork ;)

MaikuB:
Yep, happy to look at a PR that's adjusted based on what I mentioned above :)

dluksza:
I've opened new PR https://github.com/MaikuB/flutter_local_notifications/pull/488

Issue #307, opened 2019-09-04
!@Too many alarms (500) registered from uid


eugvb:
Hi, as known, max count of scheduled notifies is 500 (on samsung).
But this count is only incrementing while I cancel pending notifies and schedule new.

For example, if I run this code 5 times
```final list = await notifyPlugin.pendingNotificationRequests();
 for (int i = 0; i < list.length; i++) await notifyPlugin.cancel(list[i].id);
 for (int i = 0; i < 100; i++) {
        await notifyPlugin.schedule(...);
  }
```
I will get exception !@Too many alarms (500) registered from uid.
Also I've tried
```
 await notifyPlugin.cancelAll();
 for (int i = 0; i < 100; i++) {
        await notifyPlugin.schedule(...);
  }
```
same result

MaikuB:
I don't have a Samsung to try to reproduce this right now. If you or anyone in the community is able to do so and provide a solution then I'm happy to look at it

MaikuB:
I believe this issue should be addressed by the PR in #318 now. Will close this, feel free to re-open if this is still an issue

PR #318, opened 2019-09-14
changed FLAG_CANCEL_CURRENT to FLAG_UPDATE_CURRENT in cancelling alarms

jjs1015:
Changed `FLAG_CANCEL_CURRENT` to `FLAG_UPDATE_CURRENT` when creating PendingIntents in cancelling functions.  `FLAG_CANCEL_CURRENT ` will cancel the existing one and create a new PendingIntent, causing failure when cancel previous alarms which are created with the previous PendingIntents.

### DIFF
--- a/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -784,7 +784,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
     private void cancelNotification(Integer id) {
         Context context = registrar.context();
         Intent intent = new Intent(context, ScheduledNotificationReceiver.class);
-        PendingIntent pendingIntent = PendingIntent.getBroadcast(context, id, intent, PendingIntent.FLAG_CANCEL_CURRENT);
+        PendingIntent pendingIntent = PendingIntent.getBroadcast(context, id, intent, PendingIntent.FLAG_UPDATE_CURRENT);
         AlarmManager alarmManager = getAlarmManager(context);
         alarmManager.cancel(pendingIntent);
         NotificationManagerCompat notificationManager = getNotificationManager(context);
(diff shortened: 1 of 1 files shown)


MaikuB:
Thanks was looking to make this change as I suspect that is what is causing the issue at https://github.com/MaikuB/flutter_local_notifications/issues/307

You mentioned failures being caused though, how can I pick up on those like are there log entries to indicate there failures?

jjs1015:
Thanks for your prompt response.
You can use `dumpsys alarm` in the shell of device to check the info of existing alarms.

Step 1. Run `adb shell` to enter the shell of the device;
Step 2. Run `dumpsys alarm` to print the info of existing alarms on the device;

You can also use `dumpsys alarm |  grep {something like the package name}` to review info related to your app.

MaikuB:
Thanks for the info. Tried this with the current codebase and can confirm that with grep I see the alarm that I've scheduled. After I cancel it and take a look at the alarms, it looks like it's no longer there so I'm not sure exactly which scenario this PR is trying to solve. Are you able to provide a sample app or code to illustrate the problem that this PR would fix?

jjs1015:
Please check [sample app](https://github.com/jjs1015/flutter_local_notification_debug_sample) here.
BTW, I can reproduce this with Emulator provided by AVD (Android 8.1 with Google APIs, API 27).

MaikuB:
Ok thanks I was using the API 29 emulator. To confirm though, what's the behaviour you were expecting to be fixed here? With the sample you've written and using the changes you've made, if I click on the create button and then the cancel button, I can still see alarms being shown when I run `adb shell dumpsys alarm | grep "<package_id>"`

jjs1015:
Really sorry about my mistakes. I’ve updated my sample. Please pull the latest code. 
1. IDs should be different;
2. Extend the schedule by 1 day to avoid a notification which should be fired now.

Excepted behaviour:
1. After creation, output of  `dumpsys alarm | grep <package_id> | wc -l` is 300. 
2. After removal, output of `dumpsys alarm | grep <package_id> | wc -l` is still 300.
3. **Uninstall the Demo App**, rebuild with patch and run it.
4.  After creation, output of  `dumpsys alarm | grep <package_id> | wc -l` is 300. 
5. After removal, output of `dumpsys alarm | grep <package_id> | wc -l` is 0. (the correct behaviour, since all the alarms have been removed)

MaikuB:
Thanks, LGTM so will merge it in. However, I think your sample does highlight that there's an issue in the case updating a scheduled notifications that results in another alarm. Will look at address that outside of this PR and believe the fix required is exactly the same

PR #324, opened 2019-09-22
fix the bug related to NoSuchFieldException

NeoMindStd:
Fixes the issue where the app terminates due to NoSuchFieldException when alarm is recived or canceled on some devices (Not a fundamental solution)


MaikuB:
Unless you're able to provide steps to reproduce this issue then this won't be a PR I'll be looking to take in. Build previously failed as it states that exception is never thrown by the code that surrounds the catch block. Given that there's been no other reports on this issue, I would think your issue lies elsewhere

PR #379, opened 2019-11-27
Tidyups

domesticmouse:
- Re-generated `.gitignore` with `flutter create .` to capture the newer generated files
- Updated the layout of  `NotificationDetails` `@interface` declaration to be consistent


MaikuB:
Awesome, thanks!

PR #387, opened 2019-12-08
Added notification actions

gianlucaparadise:
With this pull request I want to add the ability to handle notification actions both on Android and iOS.


MaikuB:
I'm guessing you didn't follow [this issue](https://github.com/MaikuB/flutter_local_notifications/issues/17). Not looking at taking PRs due to issues with headless execution. Actions need to work even without the app running but there's a limitation in Flutter that prevents that from fully working

gianlucaparadise:
Ok, sorry, I didn't check it.

PR #388, opened 2019-12-09
Added notification visibility (Android only)

gianlucaparadise:
With this pull request I want to add the ability to configure the notification visibility on the lockscreen. This in an Android-only feature.

Official docs [here](https://developer.android.com/reference/android/app/Notification.html#visibility).

### DIFF
--- a/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -349,6 +349,11 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
         return icon;
     }
 
+    /**
+     * Sets the visibility property to the input Notification Builder
+     * @throws IllegalArgumentException If `notificationDetails.visibility` is not null but also
+     * not matches any known index.
+     */
     private static void setVisibility(NotificationDetails notificationDetails, NotificationCompat.Builder builder) {
         if (notificationDetails.visibility == null) {
             return;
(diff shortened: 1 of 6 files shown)


android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -348,6 +349,30 @@ private static IconCompat getIconFromSource(Context context, String iconPath, Ic
...
+                break;
+            case 1: // Public
+                visibility = NotificationCompat.VISIBILITY_PUBLIC;
+                break;
+            case 2: // Secret
+                visibility = NotificationCompat.VISIBILITY_SECRET;
+                break;
+
+            default:
+                System.out.println("FlutterLocalNotificationsPlugin.setVisibility - case not handled: " + notificationDetails.visibility);
```
MaikuB: I understand the reasoning behind adding this but would like see this removed. Logging to the console can confuse users of the plugin (has happened in the past). I would rather see us go with an IllegalArgumentException. I've seen a thread on the main Flutter repository on best practices for plugins in these scenarios but can't find it right now. At the least, there's an example in one of the official plugins [here](https://github.com/flutter/plugins/blob/017a57e1d3f8e3da7ac40ee35bbaaca059bc6cb6/packages/path_provider/android/src/main/java/io/flutter/plugins/pathprovider/StorageDirectoryMapper.java#L48)

gianlucaparadise:
As requested, I replaced the log with an exception.

MaikuB:
Thanks I'll give this and the other PR a spin on the weekend

PR #389, opened 2019-12-10
Added MediaStyle handling (Android only)

gianlucaparadise:
With this pull request I want to add the ability to set the Media Style to a notification. This in an Android-only feature.

While you cannot add actions due to [this issue](https://github.com/MaikuB/flutter_local_notifications/issues/17), with this feature you let Android treat the `largeIcon` Bitmap as album artwork.

Once actions are added to the library, with this feature you can show them as icon and you can promote up to 3 actions to be displayed in the standard view alongside the usual content.


README.md
```diff
@@ -34,6 +34,7 @@ A cross platform plugin for displaying local notifications.
     * Big text
     * Inbox
     * Messaging
+    * Media
```
MaikuB: I think a note should be left that it can only show pictures as album artwork i.e. no support for controlling media playback

android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -471,6 +477,28 @@ private static void setInboxStyle(NotificationDetails notificationDetails, Notif
         builder.setStyle(inboxStyle);
     }
 
+    private static void setMediaStyle(NotificationDetails notificationDetails, NotificationCompat.Builder builder) {
+        MediaStyleInformation mediaStyleInformation = (MediaStyleInformation) notificationDetails.styleInformation;
+        int[] showActionsInCompactView = convertIntegers(mediaStyleInformation.showActionsInCompactView);
+
+        androidx.media.app.NotificationCompat.MediaStyle mediaStyle = new androidx.media.app.NotificationCompat.MediaStyle();
+        mediaStyle.setShowActionsInCompactView(showActionsInCompactView);
```
MaikuB: From my own testing and looking at the doc that has a sample code snippet (see https://developer.android.com/reference/androidx/media/app/NotificationCompat.MediaStyle#summary), specifying actions doesn't appear to be needed for the media style notification to work. With that being the case, can you remove the code relating to the actions in both the Java and Dart side?

gianlucaparadise: Yes, it is correct that it is not needed. I added it because it will be a feature once the notification actions will be added to the plugin. Anyway I'll remove it.

README.md
```diff
@@ -34,6 +34,8 @@ A cross platform plugin for displaying local notifications.
...
+        * While media playback control using a `MediaSession.Token` is not supported, with this style you let Android treat the `largeIcon` Bitmap as album artwork
```
MaikuB: this should be bitmap instead of Bitmap

lib/src/platform_specifics/android/styles/media_style_information.dart
```diff
@@ -0,0 +1,14 @@
+import 'default_style_information.dart';
+
+class MediaStyleInformation extends DefaultStyleInformation {
+
+  MediaStyleInformation({
+    bool htmlFormatContent = false,
+    bool htmlFormatTitle = false,
+  }) : super(htmlFormatContent, htmlFormatTitle);
+
+  Map<String, dynamic> toMap() {
```
MaikuB: this method can be removed since it no longer has additional logic and just calls the base implementation

gianlucaparadise: I agree on the `toMap` method. Should I also remove the constructor?

MaikuB: unless i'm mistaken, this class doesn't seem to be needed anymore even. if so, let's remove it

android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
```diff
@@ -267,9 +268,16 @@ private static void readStyleInformation(NotificationDetails notificationDetails
             readInboxStyleInformation(notificationDetails, styleInformation, defaultStyleInformation);
         } else if (notificationDetails.style == NotificationStyle.Messaging) {
             readMessagingStyleInformation(notificationDetails, styleInformation, defaultStyleInformation);
+        } else if (notificationDetails.style == NotificationStyle.Media) {
+            readMediaStyleInformation(notificationDetails, styleInformation, defaultStyleInformation);
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private static void readMediaStyleInformation(NotificationDetails notificationDetails, Map<String, Object> styleInformation, DefaultStyleInformation defaultStyleInformation) {
```
MaikuB: `styleInformation` isn't used so can be removed as an argument

MaikuB:
I've changed the label for the button to show the media notification so makes more sense grammatically. Thanks for the PR!

PR #409, opened 2019-12-28
updated readme file to give info about ios not showing notification p…

Praveena0989:
…ermission popup


MaikuB:
Thanks for the PR but I don't see this as something that is necessary for the readme as it isn't specific to the plugin. This is how permission dialogs work on iOS in general and is applicable only for debugging purposes if you need want to see the dialog again. The comment is also a bit misleading as it is right now, it could interpreted as saying that the permission dialog won't show at all unless the app is re-installed

PR #447, opened 2020-01-20
missing_id

talmor-guy:


flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -728,8 +730,13 @@ private void schedule(MethodCall call, Result result) {
         Map<String, Object> arguments = call.arguments();
         NotificationDetails notificationDetails = extractNotificationDetails(result, arguments);
         if (notificationDetails != null) {
-            scheduleNotification(registrar.context(), notificationDetails, true);
-            result.success(null);
+            if (notificationDetails.id == null) {
```
MaikuB: I would suggest removing this code and moving the validation to be on the Dart side. This would then cover both Android and iOS. There's an existing method for validating the id. See https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/lib/src/helpers.dart

talmor-guy: ok, done

MaikuB: Thanks i'll be making some tweaks on the Dart side to use the different named constructors. Tried to push directly to here but that failed. Perhaps the permission to allow edits was disabled in your PR but that's fine

PR #488, opened 2020-02-20
[flutter_local_notifications] add support for delayed notification permission request on iOS

dluksza:
Currently notification permissions can only be requested when plugin is
initialized. In some cases this may result in suboptimal user
experience, when user will be asked for the notification permission
before he would event decide if he want to be notified by application.

This patch adds 'requestPermissions' method that may be used later on
for asking user to grant application notification permission. This way
all the notification handlers can be initialized immediately after app
startup. Then later on, when user actually decide to use local
notifications 'requestPermissions' method will be used to as him for
required permissions.

As this repository hosts two packages, please ensure the PR title starts with the name of the package that it relates to using square brackets (e.g. [flutter_local_notifications])

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -109,7 +109,7 @@ By default this plugin will request notification permission when it is initializ
 1. `requestAlertPermission`
 that control this behaviour.
 
-If you want to delay permission request, set all of the above to false, then later call `requestPermissions` method with desired permissions.
+If you want to delay permission request, set all of the above to false.
 
 ```dart
 FlutterLocalNotificationsPlugin flutterLocalNotificationsPlugin = new FlutterLocalNotificationsPlugin();
@@ -125,7 +125,11 @@ var initializationSettings = new InitializationSettings(
     initializationSettingsAndroid, initializationSettingsIOS);
 flutterLocalNotificationsPlugin.initialize(initializationSettings,
     onSelectNotification: onSelectNotification);
+```
 
+Then later call `requestPermissions` method with desired permissions.
+
+```dart
 var result = await IOSFlutterLocalNotificationsPlugin.instance?.requestPermissions(
         sound: true,
         badge: true,
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -39,6 +39,7 @@ A cross platform plugin for displaying local notifications.
 * [Android] Group notifications
 * [Android] Show progress notifications
 * [Android] Configure notification visibility on the lockscreen
+* [iOS] Delay requesting notification permission
 * [iOS] Customise the permissions to be requested around displaying notifications
 
 Note that this plugin aims to provide abstractions for all platforms as opposed to having methods that only work on specific platforms. However, each method allows passing in "platform-specifics" that contains data that is specific for customising notifications on each platform. This approach means that some scenarios may not be covered by the plugin. Developers can either fork or maintain their code for showing notifications in these situations. Note that the plugin still under development so expect the API surface to change over time.
@@ -99,6 +100,39 @@ Future onSelectNotification(String payload) async {
 In the real world, this payload could represent the id of the item you want to display the details of. Once the initialisation has been done, then you can manage the displaying of notifications.
 
 *Notes around initialisation*: if the app had been launched by tapping on a notification created by this plugin, calling `initialize` is what will trigger the `onSelectNotification` to trigger to handle the notification that the user tapped on. An alternative to handling the "launch notification" is to call the `getNotificationAppLaunchDetails` method that is available in the plugin. This could be used, for example, to change the home route of the app for deep-linking. Calling `initialize` will still cause the `onSelectNotification` callback to fire for the launch notification. It will be up to developers to ensure that they don't process the same notification twice (e.g. by storing and comparing the notification id).
+
+### [iOS only] Requesting notification permission
+
+By default this plugin will request notification permission when it is initialized. `IOSInitializationSettings` have three named parameters:
+1. `requestSoundPermission`,
+1. `requestBadgePermission`,
+1. `requestAlertPermission`
+that control this behaviour.
+
+If you want to delay permission request, set all of the above to false, then later call `requestPermissions` method with desired permissions.
+
+```dart
+FlutterLocalNotificationsPlugin flutterLocalNotificationsPlugin = new FlutterLocalNotificationsPlugin();
+var initializationSettingsAndroid =
+    new AndroidInitializationSettings('app_icon');
+var initializationSettingsIOS = new IOSInitializationSettings(
+        requestSoundPermission: false,
+        requestBadgePermission: false,
+        requestAlertPermission: false,
+        onDidReceiveLocalNotification: onDidReceiveLocalNotification,
+    );
+var initializationSettings = new InitializationSettings(
+    initializationSettingsAndroid, initializationSettingsIOS);
+flutterLocalNotificationsPlugin.initialize(initializationSettings,
+    onSelectNotification: onSelectNotification);
+/* some logic */
+var result = await flutterLocalNotificationsPlugin.requestPermissions(
+        requestSoundPermission: false,
+        requestBadgePermission: false,
+        requestAlertPermission: false,
+    );
+print('notification permissions were ${result ? '' : 'not '}granted');
+```
 
 ### Displaying a notification
 
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -127,9 +127,9 @@ flutterLocalNotificationsPlugin.initialize(initializationSettings,
     onSelectNotification: onSelectNotification);
 
 var result = await IOSFlutterLocalNotificationsPlugin.instance?.requestPermissions(
-        requestSoundPermission: false,
-        requestBadgePermission: false,
-        requestAlertPermission: false,
+        sound: true,
+        badge: true,
+        alert: true,
     );
 ```
 
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -125,13 +125,12 @@ var initializationSettings = new InitializationSettings(
     initializationSettingsAndroid, initializationSettingsIOS);
 flutterLocalNotificationsPlugin.initialize(initializationSettings,
     onSelectNotification: onSelectNotification);
-/* some logic */
-var result = await flutterLocalNotificationsPlugin.requestPermissions(
+
+var result = await IOSFlutterLocalNotificationsPlugin.instance?.requestPermissions(
         requestSoundPermission: false,
         requestBadgePermission: false,
         requestAlertPermission: false,
     );
-print('notification permissions were ${result ? '' : 'not '}granted');
 ```
 
 ### Displaying a notification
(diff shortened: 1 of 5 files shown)


flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
```diff
@@ -62,6 +64,23 @@ class FlutterLocalNotificationsPlugin {
     return true;
   }
 
+  Future<bool> requestPermissions({
```
MaikuB: what I meant [here](https://github.com/MaikuB/flutter_local_notifications/pull/311#issuecomment-585354185) was to not have any platform-specific methods in the cross-platform facing API i.e. in this class. Since this plugin now follows the [federated plugin approach](flutter.dev/go/federated-plugins), developers can access the platform-specific implementation of a plugin to access platform-specific functionality if needed.

Originally, I was thinking to have developers following the code similar to what is written below

```dart
 (FlutterLocalNotificationsPlatform.instance
             as IOSFlutterLocalNotificationsPlugin)
         ?.requestPermissions(...)
```

However, it's a bit sub-optimal as it requires developers to import the `flutter_local_notifications_platform_interface` package that is a transitive dependency some might not be aware of. As such my thoughts are to have a static `instance` property within the `IOSFlutterLocalNotificationsPlugin` and `AndroidFlutterLocalNotificationsPlugin` that points to `FlutterLocalNotificationsPlatform.instance` and cast it as the appropriate type. Can you make that change in this PR?

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -190,6 +190,18 @@ class IOSFlutterLocalNotificationsPlugin
         'initialize', initializationSettings.toMap());
   }
 
+  Future<bool> requestPermissions({
```
MaikuB: Will need to add API docs. I can add some in directly to this PR if you're ok with that

flutter_local_notifications/README.md
```diff
@@ -100,6 +101,39 @@ In the real world, this payload could represent the id of the item you want to d
...
+    initializationSettingsAndroid, initializationSettingsIOS);
+flutterLocalNotificationsPlugin.initialize(initializationSettings,
+    onSelectNotification: onSelectNotification);
+/* some logic */
+var result = await flutterLocalNotificationsPlugin.requestPermissions(
+        requestSoundPermission: false,
+        requestBadgePermission: false,
+        requestAlertPermission: false,
+    );
+print('notification permissions were ${result ? '' : 'not '}granted');
```
MaikuB: as this is a code snippet to demonstrate the API calls needed, this print statement and the `/* some logic */` line isn't needed

flutter_local_notifications/example/lib/main.dart
```diff
@@ -839,6 +849,32 @@ class _HomePageState extends State<HomePage> {
         payload: 'item x');
   }
 
+  Future<void> _requestPermissions(BuildContext context) async {
```
MaikuB: this will need to be updated based on the suggested changes below

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
dluksza: Sure, will add ios and android specific `instance` property to plugin class

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
dluksza: Will add basic docs, you can improve them if you feel like it.

dluksza:
Thank you for review, here is updated PR. I've decided to return `null` from `IOS/AndroidFlutterLocalNotificationsPlugin.instance`, then one doesn't need to use `if (Platform.isIOS)`, simple `?.` operator will do the trick ;)

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -54,6 +54,15 @@ class MethodChannelFlutterLocalNotificationsPlugin
 /// Android implementation of the local notifications plugin
 class AndroidFlutterLocalNotificationsPlugin
     extends MethodChannelFlutterLocalNotificationsPlugin {
+  /// Android specific plugin instance or null when not running in Android environment
+  static AndroidFlutterLocalNotificationsPlugin get instance {
+    if (FlutterLocalNotificationsPlatform.instance
```
MaikuB: This code block, including the if statement could be replaced by casting using the `as` keyword that was used in the earlier version of your PR. That is, it'll look like `FlutterLocalNotificationsPlatform.instance as AndroidFlutterLocalNotificationsPlugin`

If the type doesn't match it will return null as well :)

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -173,6 +182,15 @@ class AndroidFlutterLocalNotificationsPlugin
 /// iOS implementation of the local notifications plugin
 class IOSFlutterLocalNotificationsPlugin
     extends MethodChannelFlutterLocalNotificationsPlugin {
+  /// IOS specific instance of plugin or null when not running in iOS environment
+  static IOSFlutterLocalNotificationsPlugin get instance {
+    if (FlutterLocalNotificationsPlatform.instance
```
MaikuB: See comment on using `as` for casting above that can also be applied here

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -190,6 +208,19 @@ class IOSFlutterLocalNotificationsPlugin
         'initialize', initializationSettings.toMap());
   }
 
+  /// Requests given permission from user and returns current permission status.
```
MaikuB: "given permission" -> "the specified permission(s)"

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -190,6 +208,19 @@ class IOSFlutterLocalNotificationsPlugin
...
+  Future<bool> requestPermissions({
+    bool requestSoundPermission,
```
MaikuB: Sorry just realised this now but given the method itself is called `requestPermissions`, the name of the arguments can simplified to `sound`, `alert` and `badge`.

flutter_local_notifications/README.md
```diff
@@ -100,6 +101,38 @@ In the real world, this payload could represent the id of the item you want to d
...
+        requestAlertPermission: false,
+        onDidReceiveLocalNotification: onDidReceiveLocalNotification,
+    );
+var initializationSettings = new InitializationSettings(
+    initializationSettingsAndroid, initializationSettingsIOS);
+flutterLocalNotificationsPlugin.initialize(initializationSettings,
+    onSelectNotification: onSelectNotification);
+
+var result = await IOSFlutterLocalNotificationsPlugin.instance?.requestPermissions(
+        requestSoundPermission: false,
```
MaikuB: Shouldn't these parameters be set to true given that this section is to now demonstrate requesting permissions?

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
dluksza: oh, didn't know that `as` will return you `null`, was assuming that this would result in something like `ClassCastException` (yeah, can't hide my Java background ;))

MaikuB: Actually you might need to double check in case I'm wrong. Perhaps it doesn't work the same way as C#

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
dluksza: TBH, I would also remove permission parameters from `IOSInitializationSettings`, but this would be a breaking change.

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
MaikuB: Ok just tried it works fine :)
Reason I mentioned checking was I had used it when I knew the type was what I expected but can't remember if I had done this for when the type might not match as well. Perhaps the Dart docs around the [`as`](https://dart.dev/guides/language/language-tour#type-test-operators) keyword need updating hmm

MaikuB: Hmm this question would suggest it doesn't work the same way https://stackoverflow.com/questions/52494535/type-conversion-using-as-in-dart

Let's keep what you originally had then. Apologies for the confusion and sorry about that

flutter_local_notifications/README.md
dluksza: Good catch!

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
dluksza: For now I went with shorter parameter names

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
dluksza: Better be safe than sorry ;) I've reverted previous refactroing

dluksza: well, at least dartpat says that not checking types could backfire:
```
Uncaught Error: CastError: "asdf": type 'JSString' is not a subtype of type 'int'
```

MaikuB: Cool, looks like I learned something, I guess it implies it works for subtypes but better to be cautious as you said :)

flutter_local_notifications/README.md
```diff
@@ -100,6 +101,38 @@ In the real world, this payload could represent the id of the item you want to d
...
 
+### [iOS only] Requesting notification permission
+
+By default this plugin will request notification permission when it is initialized. `IOSInitializationSettings` have three named parameters:
+1. `requestSoundPermission`,
+1. `requestBadgePermission`,
+1. `requestAlertPermission`
+that control this behaviour.
+
+If you want to delay permission request, set all of the above to false, then later call `requestPermissions` method with desired permissions.
```
MaikuB: Hmm sorry I realised the intent you had went you had the "some logic" comment in the code further below now. How about breaking this sentence up. First would be "If you want to delay permission request, set all of the above to false..." that is followed by the code snippet for initialisation. After that is "then later call `requestPermissions` method with desired permissions" that then has the snippet for requesting permissions?

MaikuB:
I updated the example app as users may run the app and not hit the button to request permissions on iOS and jump straight to trying to show a notification. As a result they'll not see anything

MaikuB:
I made another update to fix an issue where `requestPermissions` would also process the launch notification that might result in it happening twice on iOS. Once build passes I'll merge it. I might make some changes to docs etc in a separate PR but everything else looks good. Thanks and much appreciated for the work you put into this!

dluksza:
Thank you for the review and for creating this plugin, without it I couldn't migrate my app ;)

MaikuB:
No worries. You may have noticed I did make tweaks to the wording in the docs to hopefully avoid misinterpretation. For example, delay the permissions request could imply the plugin allows specifying the duration of the delay before permissions are requested.

MaikuB:
I might look into a different of giving users access to the platform-specific implementation. Whilst the behaviour of the new static `instance` getters is documented, it may lead to others thinking that it's following the singleton pattern where it'll create an instance of the class if it doesn't exist

MaikuB:
I've made this change in #494 and removed `instance` to avoid confusion. You might want to update to 1.2.0 to get this change

PR #518, opened 2020-03-07
[flutter_local_notifications] add ability to toggle showing notification timestamp on Android

mojtabaghiasi:
now the developer can configure displaying the notification create or update time. default is true in android. just use **showWhen** attribute in AndroidNotificationDetails.

`  var androidPlatformChannelSpecifics = AndroidNotificationDetails(
      'your channel id', 'your channel name', 'your channel description',
      showWhen : false , enableVibration : false , autoCancel : false ,onlyAlertOnce: true, ongoing: true , playSound: false , icon: 'app_icon' );`

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -134,8 +134,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
                 .setContentIntent(pendingIntent)
                 .setPriority(notificationDetails.priority)
                 .setOngoing(BooleanUtils.getValue(notificationDetails.ongoing))
-                .setOnlyAlertOnce(BooleanUtils.getValue(notificationDetails.onlyAlertOnce))
-                .setShowWhen(BooleanUtils.getValue(notificationDetails.setShowWhen));
+                .setOnlyAlertOnce(BooleanUtils.getValue(notificationDetails.onlyAlertOnce));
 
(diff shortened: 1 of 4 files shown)


flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -134,7 +134,8 @@ private static Notification createNotification(Context context, NotificationDeta
                 .setContentIntent(pendingIntent)
                 .setPriority(notificationDetails.priority)
                 .setOngoing(BooleanUtils.getValue(notificationDetails.ongoing))
-                .setOnlyAlertOnce(BooleanUtils.getValue(notificationDetails.onlyAlertOnce));
+                .setOnlyAlertOnce(BooleanUtils.getValue(notificationDetails.onlyAlertOnce))
+                .setShowWhen(BooleanUtils.getValue(notificationDetails.setShowWhen));
```
MaikuB: I believe this needs amending. Any notification that was scheduled before this update was released would have a null value for the property when loaded from shared preferences

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -168,6 +171,7 @@ class AndroidNotificationDetails {
       this.largeIcon,
       this.largeIconBitmapSource,
       this.onlyAlertOnce,
+      this.showWhen,
```
MaikuB: Given the default behaviour is to show the timestamp, making the default value true will make this more obvious to users of the plugin

mojtabaghiasi:
Thanks for your response . I will fix these problems and commit them.

mojtabaghiasi:
I fixed all the problems in my last commit. thanks

Issue #563, opened 2020-04-06
custom sound doesn't work on android


JakaboosD:
I'm trying to play my own custom sound for my notifications on android. But on my android emulator, it just plays the default sound and on my personal device, it doesn't even make a sound (both run android 9).
```
FlutterLocalNotificationsPlugin _notifications = FlutterLocalNotificationsPlugin();

var initializationSettingsAndroid = AndroidInitializationSettings('app_icon');
var initializationSettingsIOS = IOSInitializationSettings();
var initializationSettings = InitializationSettings(initializationSettingsAndroid, initializationSettingsIOS);
await _notifications.initialize(initializationSettings);
var androidPlatformChannelSpecifics = AndroidNotificationDetails('test_channel', 'test', '', playSound: true, sound: RawResourceAndroidNotificationSound('my_sound'));
var iOSPlatformChannelSpecifics = IOSNotificationDetails();
var platformChannelSpecifics = NotificationDetails(androidPlatformChannelSpecifics, iOSPlatformChannelSpecifics);
await _notifications.show(0, 'test', 'this is a test', platformChannelSpecifics);
```

I don't get any errors. My file is in android/app/src/main/res/raw/my_sound.mp3

MaikuB:
Note that the sound is tied to the notification channel, which is created the first time you try to show a notification against a particular channel ID. The settings passed in will persist so won't change even if you changed the settings around the sound but use the same channel ID. So if you went through using the default sound (which is what happens by default when you don't specify anything) to specifying a custom sound, you'll need to uninstall and reinstall the app

MaikuB:
Regarding the sound not playing on your phone, you'll need to check the settings on your phone. The volume level is separate from getting notification sounds

JakaboosD:
Thanks! That did the trick! On my personal phone it was the same problem. I had just been updating the app. I uninstalled and reinstalled the app and now it plays my custom sound!

PR #564, opened 2020-04-07
 [flutter_local_notifications] add method for creating notification channels on Android

ZaarU:
This is necessary to receive the first message from the background when the channel is not initialized.
Before this PR first message from background received with default setting(default sound, vibration and etc.)

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -29,7 +29,7 @@ import androidx.core.graphics.drawable.IconCompat;
 import com.dexterous.flutterlocalnotifications.models.IconSource;
 import com.dexterous.flutterlocalnotifications.models.MessageDetails;
 import com.dexterous.flutterlocalnotifications.models.NotificationChannelAction;
-import com.dexterous.flutterlocalnotifications.models.NotificationChannelPlugin;
+import com.dexterous.flutterlocalnotifications.models.NotificationChannelDetails;
 import com.dexterous.flutterlocalnotifications.models.NotificationDetails;
 import com.dexterous.flutterlocalnotifications.models.PersonDetails;
 import com.dexterous.flutterlocalnotifications.models.styles.BigPictureStyleInformation;
@@ -126,7 +126,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
     }
 
     private static Notification createNotification(Context context, NotificationDetails notificationDetails) {
-        setupNotificationChannel(context, NotificationChannelPlugin.fromNotificationDetails(notificationDetails));
+        setupNotificationChannel(context, NotificationChannelDetails.fromNotificationDetails(notificationDetails));
         Intent intent = new Intent(context, getMainActivityClass(context));
         intent.setAction(SELECT_NOTIFICATION);
         intent.putExtra(PAYLOAD, notificationDetails.payload);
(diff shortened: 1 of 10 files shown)


MaikuB:
Thanks for the PR. Can you clarify what you mean in the scenario you described, I've yet to see this issue you've described happen or reported by others in the community. The plugin always will attempt to create a channel if needed prior to showing a notification. I've tested this using the example app and if I schedule the notification (note: I tweaked the code to show the notification in a minute), restart the device, it will show at the scheduled time with the custom sound etc. I'm wondering if perhaps your issue was more to do with the fact that you reused the same channel id that was already created, where the initial settings will remain unless you uninstalled the app. This is the expected behaviour and what happened to another developer recently (see https://github.com/MaikuB/flutter_local_notifications/issues/563)

The idea to add Android-specific methods for managing notification channels is good though and was something I was thinking could be added now that there's an Android-specific version of the plugin

ZaarU:
This problem occurs when you send a message to a device (for example, Firebase cloud), and the application is in the background (the message is not forcibly intercepted in flutter_local_notifications), and flutter_local_notifications has not yet created a channel, because the channel is registered in the "show" method.
 A message arrives, but its settings are set by default (sounds, vitration, etc.).

MaikuB:
In your example you referring to using Firebase Cloud Messaging for sending push notifications?

ZaarU:
> In your example are you referring to using Firebase Cloud Messaging for sending push notifications?
That's right.
But since the channel was not created (in my case, I have not yet called "show" method), then the notification came with the default settings.

ZaarU:
It is need on API >= 26

ZaarU:
> It is need on API >= 26
> https://developer.android.com/training/notify-user/build-notification#Priority

fghdfghdg

ZaarU:
Excuse me. I walked away, and my child began to have fun)))

MaikuB:
> That's right.
But since the channel was not created (in my case, I have not yet called "show" method), then the notification came with the default settings.

Ok thanks. I asked to double check as you originally said Firebase Cloud

> It is need on API >= 26
https://developer.android.com/training/notify-user/build-notification#Priority

Yep I know

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -75,6 +76,7 @@
     private static final String SELECT_NOTIFICATION = "SELECT_NOTIFICATION";
     private static final String SCHEDULED_NOTIFICATIONS = "scheduled_notifications";
     private static final String INITIALIZE_METHOD = "initialize";
+    private static final String CREATE_CHANNEL_METHOD = "create_notification_channel";
```
MaikuB, line 79: the name of the constant doesn't match existing convention, which would have resulted in this being named `CREATE_NOTIFICATION_CHANNEL_METHOD`

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationChannelPlugin.java
```diff
@@ -0,0 +1,89 @@
...
+        Integer b = (Integer) arguments.get(LED_COLOR_BLUE);
+        if (a != null && r != null && g != null && b != null) {
+            notificationChannel.ledColor = Color.argb(a, r, g, b);
+        }
+
+        notificationChannel.enableLights = (Boolean) arguments.get(ENABLE_LIGHTS);
+        return notificationChannel;
+    }
+
+    public static NotificationChannelPlugin fromNotificationDetails(NotificationDetails notificationDetails) {
```
MaikuB, line 72: This is a great idea as trying to think about how to refactor the existing code if I added in methods for channel management :)

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationChannelPlugin.java
```diff
@@ -0,0 +1,89 @@
...
+    private static final String ENABLE_VIBRATION = "enableVibration";
+    private static final String VIBRATION_PATTERN = "vibrationPattern";
+    private static final String CHANNEL_ACTION = "channelAction";
+    private static final String ENABLE_LIGHTS = "enableLights";
+    private static final String LED_COLOR_ALPHA = "ledColorAlpha";
+    private static final String LED_COLOR_RED = "ledColorRed";
+    private static final String LED_COLOR_GREEN = "ledColorGreen";
+    private static final String LED_COLOR_BLUE = "ledColorBlue";
+
+    public String channelId = "Default_Channel_Id";
```
MaikuB, line 27: I made some comments on the Flutter side (as that's what I looked at first) that also apply here where the `channel` prefix for the name of the fields aren't necessary. `channelAction` can be kept though and set to `NotificationChannelAction.CreateIfNotExists` when `NotificationChannelPlugin from(Map<String, Object> ...)` is called since this class is used when a channel is creating just before a notification is shown

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -940,6 +956,15 @@ private Boolean sendNotificationPayloadMessage(Intent intent) {
         }
         return false;
     }
+
+    private void initializeChannel(MethodCall call, Result result) {
```
MaikuB, line 960: can you rename this to `createNotificationChannel`?

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -940,6 +956,15 @@ private Boolean sendNotificationPayloadMessage(Intent intent) {
...
+        Map<String, Object> arguments = call.arguments();
+        NotificationChannelPlugin notificationChannelPlugin = NotificationChannelPlugin.from(arguments);
+        if (notificationChannelPlugin != null) {
```
MaikuB, line 963: From what I've seen, the null check is redundant

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -168,6 +169,12 @@ class AndroidFlutterLocalNotificationsPlugin
         return Future.error('method not defined');
     }
   }
+
+  /// Initializes channel for android API >= 26(from Oreo).This is necessary to 
+  /// receive the first message from the background when the channel is not initialized.
+  Future<bool> createNotificationChannel(AndroidNotificationChannel notificationDetails) async {
```
MaikuB, line 175: Can you update the example app and a button where it will only call this method to create a notification channel?

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -168,6 +169,13 @@ class AndroidFlutterLocalNotificationsPlugin
         return Future.error('method not defined');
     }
   }
+
+  /// Creates a notification channel.
+  ///  
+  /// Only applies to Android 8.0+
+  Future<bool> createNotificationChannel(AndroidNotificationChannel notificationDetails) async {
```
MaikuB, line 176: See comment made previously on making this return `Future<void>` as the Android code never returns false

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
```diff
@@ -247,4 +248,13 @@ class FlutterLocalNotificationsPlugin {
     return FlutterLocalNotificationsPlatform.instance
         ?.pendingNotificationRequests();
   }
+
+  Future<bool> createNotificationChannel(AndroidNotificationChannel notificationChannel) async {
```
MaikuB, line 252: See comment made previously on making this return Future<void> as the Android code never returns false. The return null statement will be redundant after making those changes

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
```diff
@@ -247,4 +248,13 @@ class FlutterLocalNotificationsPlugin {
     return FlutterLocalNotificationsPlatform.instance
         ?.pendingNotificationRequests();
   }
+
```
MaikuB, line 251: Realised API docs will be needed here but I can add to this PR once other changes are made and before I merge it in

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
```diff
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:flutter/foundation.dart';
+import 'package:flutter_local_notifications/src/platform_specifics/android/notification_channel.dart';
```
MaikuB, line 5: Relative imports are being used within the plugin and should be placed after package imports. See https://dart.dev/guides/language/effective-dart/usage#prefer-relative-paths-when-importing-libraries-within-your-own-packages-lib-directory

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -612,48 +614,56 @@ private static void setBigTextStyle(NotificationDetails notificationDetails, Not
...
+                    notificationChannel.setLightColor(notificationChannelDetails.ledColor);
                 }
-                notificationChannel.setShowBadge(BooleanUtils.getValue(notificationDetails.channelShowBadge));
+                notificationChannel.setShowBadge(BooleanUtils.getValue(notificationChannelDetails.showBadge));
                 notificationManager.createNotificationChannel(notificationChannel);
             }
         }
     }
 
     private static Uri retrieveSoundResourceUri(Context context, NotificationDetails notificationDetails) {
```
MaikuB, line 648: if these methods are redirecting to the new `retrieveSoundResourceUri(Context context, String sound, SoundSource soundSource)` method you've defined then we could get rid of the these two (i.e. `retrieveSoundResourceUri(Context context, NotificationDetails notificationDetails)` and `retrieveSoundResourceUri(Context context, NotificationChannelDetails notificationChannelDetails)`

flutter_local_notifications/example/lib/main.dart
```diff
@@ -404,6 +405,12 @@ class _HomePageState extends State<HomePage> {
                       await _showNotificationWithAttachment();
                     },
                   ),
+                  Platform.isAndroid ? PaddedRaisedButton(
```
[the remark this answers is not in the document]
ZaarU: I will remove the check altogether. No need for it

Issue #91, opened 2018-08-22
Add repeat interval for monthly notifications?


SuitMonkeyB:
Hi MaikuB,
would it be possible to add a repeatInterval for monthly alarms?

MaikuB:
Could you elaborate more with an example? I'm assuming you mean something have notification every x months? In any case, I'd be happy to look at a PR for this

SuitMonkeyB:
Oh, I mean specifically with the repeatInterval. Functions can be made to fire a notification once a minute, hour, day, week, but not monthly currently.

MaikuB:
Oh yep gotcha now. Can you explain the specific scenario you're trying to solve? I imagine it's to show a notification at a specific date and time every month.

SuitMonkeyB:
Yea that's exactly it. Say for example the user sets a reminder within my app to be reminded once a month at a certain date to buy dog food. 
My app then scuedules a reoccurring notification which fires at that date every month. 

I noticed the infrastructure based on repeatInterval was missing the ability to set repeat monthly.

MaikuB:
Note: sorry this will be long...main answer is in last paragraph

Sounds like you're looking at the periodicallyShow function, which wouldn't have been suitable to what you're doing as it's meant to more closely mirror [this](https://developer.apple.com/documentation/usernotifications/untimeintervalnotificationtrigger) in iOS and that isn't tied to a specific date and depends on when you've called it. Furthermore, the interval is based on the number of seconds so that's why the month is not available in the enum.

My current thinking is it won't be added or at least I don't have a timeframe for putting it. One of the reasons why it hasn't been done as well is that the expected behaviour of a monthly notification could potentially vary depending on each app intends it to do. One might define this to be every 30 days, if it's once a month, (how many days/seconds does that mean) or day of the month etc. If I look at when I schedule a event using the Google calendar app, if I did the a monthly event that started on the 31st of August, then it falls into the "day of the month" scenario as it the event on appears on months with 31 days.

IMO, as opposed to having monthly notifications, I think if and when #21 is done, that would give devs more flexibility to do monthly notifications that would fit their needs as they could schedule one notification, be notified when it appears and attempt to schedule another one in the future. I believe the headless Dart work that the Flutter team has done should be on the dev channel (0.5.8) now to see how it works. However, even if #21 was done I don't think I'd release an update until the changes done by the Flutter team has rolled over into the beta channel since more devs are likely to be on that channel/branch.

[...]

PR #582, opened 2020-04-14
Added monthly option

anoobbava:
Hi,
Currently, We have an option to set the repetition of flutter notifications in every minute, daily and weekly, it will be good if we have an option to send the notification monthly say 5ht of every month.
@MaikuB: please consider this requirement, and I have created a rough PR for the same, I am not sure it is correct changes for that, please discard if it is not good, But it will be good if we have this option, as well. If this option is already  achieved please update otherwise


MaikuB:
This has been raised before where I have given reasons why it's not supported in [this thread](https://github.com/MaikuB/flutter_local_notifications/issues/91) if you want to read in more depth. To summarise, for you, 30 days might be what you consider monthly but for others it may be to have a notification on the same date of each month. Developers should implement what's suited for their application. 

Furthermore, I have plans to deprecate some of the existing methods that schedule notifications as there are known issues when it comes to dealing with timezones so won't be taking a PR for this

PR #598, opened 2020-04-24
[flutter_local_notifications] add support for `setWhen` on Android notifications

### LINKED ISSUE #596: Android: support `setWhen`
It would be great if the plugin would support the [setWhen](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder?hl=en#setWhen(long)) method for the notification to set a custom timestamp for the notification.

This is especially important if you have multiple notifications displayed at the same time (e.g. one per download) and update them independently. Then, whichever notification gets updated last will be moved to the left-most side in the status bar. When the other notification updates, this one will be moved to the left and the notifications will swap places all the time, which looks quite weird.

To fix this, it should be possible to set a fixed timestamp for each of the notifications (e.g. the time of the first creation of the notification) in order to keep the notifications in the initial order (see https://stackoverflow.com/questions/31574474/stop-sorting-android-notifications).

nioncode:
Closes #596.


MaikuB:
Thanks a lot for the PR! I've updated the API doc for clarity and added a test :)

Issue #139, opened 2018-12-17
Loop sound for notification


anhhh11:
I think it should have something like customizing FLAG_INSISTENT flag (https://stackoverflow.com/questions/26631651/android-notification-builder-how-to-setsound-so-that-sound-plays-in-a-looper) to support loop notification sound.

MaikuB:
Sorry for the late response. If this is you something need, feel free to submit a PR as it appears you already have a fork

kodaizen:
Hello, 
I am also interested in repeating the sound

MaikuB:
Happy to look at a PR for this

MaikuB:
Closing this as it's not something I'm not looking to work on. As stated above, if someone in the community needs this then submit a PR

reshuashish:
how to get this feature for ios? (https://stackoverflow.com/questions/20063674/sound-notification-in-loop-in-for-alarm-application-in-ios)

MaikuB:
There is no such feature on iOS

reshuashish:
I am working on the alarm application in IOS. Where I want to send the notification when the application is in background mode, the sound notification would be like ALARM. i.e. sound will not stop until the user doesn't perform any interaction.

Is there any way to make it happen, I have seen some Alarm apps on AppStore  which use the same principle for alarm

PR #608, opened 2020-05-01
[flutter_local_notifications] Add extra flags for Android notifications

andylei:
Fixes https://github.com/MaikuB/flutter_local_notifications/issues/139

This commit lets users add any extra android notification flags not handled specifically by AndroidNotificationDetails`

### DIFF
--- a/flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
+++ b/flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
@@ -50,7 +50,7 @@ class AndroidNotificationDetails {
     this.visibility,
     this.timeoutAfter,
     this.category,
-    this.additionalFlags = 0,
+    this.additionalFlags,
   });
 
   /// The icon that should be used when displaying the notification.
@@ -121,7 +121,7 @@ class AndroidNotificationDetails {
   /// Specifies if this notification will function as the summary for grouped notifications.
   final bool setAsGroupSummary;
 
-  /// Sets the group alert behavior for this notification.
+  /// Specifies the group alert behavior for this notification.
   ///
   /// Default is AlertAll.
   /// See https://developer.android.com/reference/android/support/v4/app/NotificationCompat.Builder.html#setGroupAlertBehavior(int) for more details.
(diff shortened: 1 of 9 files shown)


flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -202,6 +203,12 @@ class AndroidNotificationDetails {
   /// Refer to Android notification API documentation at https://developer.android.com/reference/androidx/core/app/NotificationCompat.html#constants_2 for the available categories
   final String category;
 
+  /// Extra notification flags.
+  ///
+  /// These flags will get added to the Notification object's flags field: https://developer.android.com/reference/android/app/Notification#flags
+  /// Values are the constants that start with FLAG_: https://developer.android.com/reference/android/app/Notification#FLAG_AUTO_CANCEL
```
MaikuB, line 209: For extra clarity and to improve readability, can line be changed to

`For a list of a values, refer to the documented constants prefixed with "FLAG_" (without the quotes) at https://developer.android.com/reference/android/app/Notification.html#constants_1. For example, use a value of `4` to allow the audio to repeat as documented at https://developer.android.com/reference/android/app/Notification.html#FLAG_INSISTENT`

I assume you were trying to give an example of a flag but the wording doesn't make that clear. The flag I've given in the example would be more of use as well as per the related issue. Furthermore, auto-cancel is already supported by the plugin without the use of flags

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -202,6 +203,12 @@ class AndroidNotificationDetails {
...
+  final int extraFlags;
```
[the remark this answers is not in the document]
andylei: fixed

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
andylei: fixed

andylei:
Added this to the example app as well

MaikuB:
Thanks noticed something that I missed last time that I'll push directly to the PR

andylei:
thanks! any idea when this will get released?

MaikuB:
It's already out

PR #613, opened 2020-05-06
Adds support for broadcasting an intent with the notification in Android

vkammerer:
This adds support for broadcasting an [`Intent`](https://developer.android.com/reference/android/content/Intent) containing a notification in Android.

The current implementation is the one I am using in my own application and it has proven very useful, so I'm opening this PR here in the hope that it can benefit others too.

### Use case
It enables using the notification in other ways than via `notify`, such as in:
- [Foreground services](https://developer.android.com/guide/components/services#Foreground)
- [WorkManager](https://developer.android.com/reference/androidx/work/WorkManager)
- and maybe other Android APIs I'm unaware of

This can be very helpful for advanced Android Flutter applications, especially those who need to keep executing in the background. The PR includes an example of using it in a "dumb" foreground service which does nothing but start. 

The existing `notify` method can then be used against a `notificationId` previously used via an `Intent`, so it possible to keep using only this plugin to update a notification displaying in a foreground service or WorkManager.

It opens the door to other plugins using this one, in order to benefit from its good maintenance and good feature coverage, while focusing of their core functionality.

### Implementation
The proposed API is to:
- add a `broadcastTarget` String argument to the `AndroidNotificationDetails` constructor, which will be parsed by `Class.forName` in Android.
- let plugin authors or end users implement their own [`BroadcastReceiver`](https://developer.android.com/reference/android/content/BroadcastReceiver), whose `onReceive` method will be responsible for handling the notification.

I have tried to follow the existing architecture, but would happily change the implementation if needed.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -94,10 +94,12 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
     private static final String INVALID_BIG_PICTURE_ERROR_CODE = "INVALID_BIG_PICTURE";
     private static final String INVALID_SOUND_ERROR_CODE = "INVALID_SOUND";
     private static final String INVALID_LED_DETAILS_ERROR_CODE = "INVALID_LED_DETAILS";
+    private static final String INVALID_BROADCAST_TARGET_ERROR_CODE = "INVALID_BROADCAST_TARGET";
     private static final String INVALID_LED_DETAILS_ERROR_MESSAGE = "Must specify both ledOnMs and ledOffMs to configure the blink cycle on older versions of Android before Oreo";
     private static final String NOTIFICATION_LAUNCHED_APP = "notificationLaunchedApp";
     private static final String INVALID_DRAWABLE_RESOURCE_ERROR_MESSAGE = "The resource %s could not be found. Please make sure it has been added as a drawable resource to your Android head project.";
     private static final String INVALID_RAW_RESOURCE_ERROR_MESSAGE = "The resource %s could not be found. Please make sure it has been added as a raw resource to your Android head project.";
+    private static final String INVALID_BROADCAST_TARGET_ERROR_MESSAGE = "The specified class name doesn't match any existing class in the project";
     static String NOTIFICATION_ID = "notification_id";
     static String NOTIFICATION = "notification";
     static String NOTIFICATION_DETAILS = "notificationDetails";
@@ -686,8 +688,21 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
 
     static void showNotification(Context context, NotificationDetails notificationDetails) {
         Notification notification = createNotification(context, notificationDetails);
-        NotificationManagerCompat notificationManagerCompat = getNotificationManager(context);
-        notificationManagerCompat.notify(notificationDetails.id, notification);
+        if (notificationDetails.broadcastTarget == null) {
+            NotificationManagerCompat notificationManagerCompat = getNotificationManager(context);
+            notificationManagerCompat.notify(notificationDetails.id, notification);
+        }
+        else {
+            try {
+                Class broadcastTargetClass = Class.forName(notificationDetails.broadcastTarget);
+                Intent intent = new Intent(context, broadcastTargetClass);
+                intent.putExtra("notificationId", notificationDetails.id);
+                intent.putExtra("notification", notification);
+                context.sendBroadcast(intent);
+            } catch (ClassNotFoundException e) {
+                return;
+            }
+        }
     }
 
     private static NotificationManagerCompat getNotificationManager(Context context) {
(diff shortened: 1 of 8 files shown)


MaikuB:
I'll try to take a look more closely later in the week but can I clarify if you meant to say POC as I'm not familiar with the POF acronym and the C key is pretty close to F key on the keyboard

vkammerer:
Indeed I meant POC :) , and then I edited this part in the comment and removed that sentence.
I've also noticed the tests weren't passing at first so I added another commit to fix them. I look forward to your feedback

MaikuB:
I've taken a closer look and given the goal of this plugin is around getting access to the native notification APIs, I don't consider what's proposed here is not core functionality for the functionality. This looks to be more of advanced functionality that is up to application developers to implement as it's often the case the libraries that provide abstractions to a particular area of native APIs can't accommodate for everything.

The proposed changes aren't ideal either. For example, `show` is documented a method that shows a notification. Even without the documentation, it would reasonable to expect that's what it would do. However, the value of one parameter (`broadcastTarget`) can completely change what `show` does so it now ends up having more than one responsibility.

So based on what I've understand, this won't be something that will be merged in. I can see why it might be useful so you might want to consider sharing you work in some other form e.g. blog post, article etc :)

vkammerer:
Ok thank you for your reply.

Regarding the implementation, I understand the confusion about the method `show`: my first implementation added a method `broadcast`, which was more explicit, even if it meant more code. I could happily change it to that if you think it is appropriate.

However, I see that your reply also refutes the use case, so I'm willing to argue a bit more here :). The API that Android exposes for "showing" a notification, from my understanding, works in 3 main steps:
- creating the channel
- building the notification
- displaying the notification (either as a standalone notification via `notify` or as a required part of another component like via `startForeground`)

I totally understand your concern about constraining your plugin's API and making sure it doesn't start handling tasks that aren't its core functionality. But from what I understand, most of your plugin's work is related to the first two steps, and they can form a well defined responsibility for your plugin. The third step however (the way to use the notification channel and the built notification) is more succinct.

So while it would of course be possible for a user wishing to use a foreground service to reimplement the first two steps all by themselves, it would be a shame that they don't benefit from all the work you've done here, and it would also mean that their bundle size becomes larger with a lot of duplicate code if they still want to use your plugin for other tasks.

I guess that this raises more generally the question about how Flutter plugins should work together when they wrap APIs that are interdependent. The standard way to do this in Android is via Intents, so it seems natural that plugins expose them too.

My "dumb" foreground service may also have failed to show how useful they can be, especially in the context of a Flutter application. But they are the only way to keep executing reliably in the background in Android. Moreover, I have implemented a foreground service that executes Flutter / dart code, which I am going to release / propose soon (in issues of other repos), so it will make them even more accessible to Flutter developers.

I hope this helps understand the use case. If you are interested in including any way of broadcasting Intents, I'd be happy to modify the current implementation.

MaikuB:
> But from what I understand, most of your plugin's work is related to the first two steps, and they can form a well defined responsibility for your plugin

In terms of code, sure (note: much less code for first two steps on iOS) but the core value proposition of the plugin is to be able to display the notification so the 3rd step is what's crucial.

> So while it would of course be possible for a user wishing to use a foreground service to re-implement the first two steps all by themselves, it would be a shame that they don't benefit from all the work you've done here, and it would also mean that their bundle size becomes larger with a lot of duplicate code if they still want to use your plugin for other tasks.

> I guess that this raises more generally the question about how Flutter plugins should work together when they wrap APIs that are interdependent. The standard way to do this in Android is via Intents, so it seems natural that plugins expose them too.

This goes back to my point above that when it comes to advanced use cases, developers will need to write their own code. Sometimes it's possible to fork as library and use the work done as a base. Other times, it's requires starting from scratch.  

The proposed changes here is to broadcast an intent (it's core functionality) that holds a notification, which is a general case of broadcasting an arbitrary object. As such, I would be saying the approach should be to see if the general case can be achieved. It's possible it can't be and that application developers should look to write their own to achieve their needs. Instead, it seems like what's being done here would effectively be asking every plugin maintainer to expose an API that broadcast the objects they use/create on the native side. I've yet to see this done in other ecosystems that deal with cross-platform apps and isn't IMO a sustainable approach either

vkammerer:
Ok thank you for your quick reply.

I will try another approach, and will comment here again when I've made some progress.

vkammerer:
After wondering how I could keep benefiting from your plugin's good maintenance and feature coverage, while still adding the broadcasting functionality proposed above, I came to the following solution: I created another plugin that "depends" on yours: https://github.com/vkammerer/flutter_local_notifications_broadcast

The dependency is achieved by defining it in `pubspec.yaml` and in `build.gradle`, as can be seen in [this commit](https://github.com/vkammerer/flutter_local_notifications_broadcast/commit/642190ce392a5b617261d7fe65fedee0bb5cbe4e).

As you can see in the [Android plugin class](https://github.com/vkammerer/flutter_local_notifications_broadcast/blob/master/android/src/main/java/com/vincentkammerer/flutter_local_notifications_broadcast/FlutterLocalNotificationsBroadcastPlugin.java), I extend your plugin's class to achieve the desired functionality:
```java
public class FlutterLocalNotificationsBroadcastPlugin extends FlutterLocalNotificationsPlugin
```

The only change I've had to do in your project to enable this is to make the methods `extractNotificationDetails` and `createNotification` public instead of private.

I've also added an example similar to the one I was proposing in the PR, and it works well in the tests I've done.

MaikuB:
Sounds good :)

That was the initial thought I had if it's not taken to the general case though wasn't sure about how feasible it would be as it depends on what APIs are currently public

vkammerer:
Cool, good to hear that you like that solution :)

Do you think that you could make `extractNotificationDetails` and `createNotification` public in your plugin then?

MaikuB:
Sorry but not looking to do so as it's really internal logic and adds more burden when it comes to maintanence in looking at what's considered a breaking change

vkammerer:
Ok, so I'll have to keep a version of your plugin and reapply changes on top to keep my functionality working then.

> Edit: forgot to add I'm actually going to review the scopes of some of the native code and reduce accessibility where possible as well

Does that mean there may be other changes than making these two methods public for me to keep using it?

MaikuB:
> Does that mean there may be other changes than making these two methods public for me to keep using it?

Yes there may more changes but I didn't say I would make the two methods public. What I'm saying is there may be more things in the code that are public that could be made private that I'll be looking at

PR #618, opened 2020-05-10
[flutter_local_notifications] Adding comment to grouping example

cgalan12:
Adding comment to grouping example to remark that the summary notification is required


MaikuB:
Thanks for attempting to add more info some of the info needs updating in general, which I'll be doing soon so going to close this

Issue #92, opened 2018-08-23
Custom intervals for periodicallyShow


Zil85:
Are there any thoughts on adding custom intervals for showing repeating notifications?

Currently it supports the repeat intervals (EveryMinute, Hourly, Daily, Weekly).

But for my case I would be interested in having the users select the interval themselves.

Without going to much into the businesslogic.
Basically the user sets a reminder and chooses a time interval e.g. every 5min, 20min, 1.5hours, etc.
It would most likely always be in the range of 5min - 4hours.

MaikuB:
Have been thinking about it. The reason why it's not there was that there are devs that are also targeting iOS 8 and 9 and custom intervals like that aren't supported. However, the fraction of services on those versions are very small so that's why I'm considering this ATM. This will mean it will be a function (possibly an overload) that only works for Android and iOS 10+ so will need to be clearly documented

mccadi:
+1 ...I would need this as well! Hoping to set intervals in terms of every week, every other week, every three weeks, every four weeks...

MaikuB:
I have seen your fork and I'm always open to look at a PR :)

mccadi:
I did, I just tried changing the static values for minute, hour, week, etc but I am not quite sure if this will work or repeat properly until testing, but iOS notifications seem to work. Also having trouble with android getting notifications to show reliably atm...currently a work in progress

LSmint:
Is it possible to combine every other week with specific weekday (e.g.: Monday and Friday)?

cesarureno:
Any updates?

MaikuB:
@cesarureno I'm actually looking to close this now as no one has looked into contributing a PR despite how long this issue has been around. I would say if you're after this feature that you look into implementing yourself and submit a PR if possible

cesarureno:
Oh, I see @MaikuB I will start to try make some awesome, thanks for your time and your package

Bryanx:
You can do this by combining this package with [flutter_background_fetch](https://github.com/transistorsoft/flutter_background_fetch).

- Just use schedule() to schedule the first notification and store the required interval and starttime in sqlite or shared prefs.
- Create a headless task with flutter_background_fetch that runs in the background every 15 minutes.
- Make the task check if the notification has been sent. 
- If it has not been sent schedule a new one based on the sent time of the last notification.

Unfortunately the interval has to be > 15min and checking every 15 min will cost some battery. You can increase the fetch time for Android but not for iOS.

[...]

PR #632, opened 2020-05-19
Add support of intervals quantity for periodic notifications

BlueLightFirefly:
Added number of intervals for repeating notifications. For ios lower than ios 10 works as before. 

Works as: if you set `repeatInterval` as EveryMinute and set `repeatIntervalQuantity` to 5, that means "Repeat every 5 minutes". Same for other `repeatInterval`

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -301,7 +301,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
             default:
                 break;
         }
-
+        repeatInterval = repeatInterval * notificationDetails.repeatIntervalQuantity;
         long startTimeMilliseconds = notificationDetails.calledAt;
         if (notificationDetails.repeatTime != null) {
             Calendar calendar = Calendar.getInstance();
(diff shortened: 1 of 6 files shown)


MaikuB:
Thanks for the PR but if the behaviour isn't going to apply to older versions of iOS (which I understand because it doesn't support it) then having it as a separate method would be better IMO. I'm also currently working on fixing the issue with daylight savings for the scheduled notifications. As part of this, I have a solution for `periodicallyShow` to trigger notifications with exact timings on Android. I'm hoping to release a development version of the plugin soon. This could form the basis for another method where another method exists for custom intervals e.g. specifying the interval based on the number of seconds. 

As such I'd say that at this stage I won't be looking at merging this PR. I hope the reasoning makes sense

BlueLightFirefly:
I've added such possibility only because I need this functionality right now. This solution resolves my problems, and I thought that it can be helpful for other devs who need it now. This solution can solve issue https://github.com/MaikuB/flutter_local_notifications/issues/92 . Until you will implement your issue it can be helpful for others. But the decision is yours.

About ios lower ios 10: in my opinion, not too much apps support ios 9, and it is not necessary. Support of low Android versions is critical, but ios not so much. Most devices which support ios 9 can be updated to ios 10 (except iphone 4s, ipod 5th gen, all devices you can see at https://en.wikipedia.org/wiki/IOS_10#Supported_devices). But of course, will be perfect if your plugin will support as much as possible OS. And solution with another method call is suitable.

MaikuB:
> About ios lower ios 10: in my opinion, not too much apps support ios 9, and it is not necessary...

Note that i wasn't arguing for supporting older versions of iOS as it's the old notification APIs don't support them to begin with and made a point that there's the older versions don't have much marketshare in #92.

Closing this now and can be revisited later as I'm targeting a new development release really soon. Developers can refer to the changes in this PR to maintain their own forks with said changes if it's something they need urgently

PR #654, opened 2020-06-01
Fixed missing emulator-headless

fkorotkov:
Apparently there is no more `emulator-headless`: https://stackoverflow.com/a/59250326


MaikuB:
Thanks a lot again @fkorotkov :)

PR #661, opened 2020-06-07
[flutter_local_notifications] Fixes notification details if null

Vanethos:
- Adds try/catch to json deserializing
- Adds check if object is null

This is due to the fact that the notification may be of incorrect json type or we may receive an empty object `{}` which will lead to nullpointer exceptions down the line

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/ScheduledNotificationReceiver.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/ScheduledNotificationReceiver.java
@@ -19,7 +19,6 @@ import java.lang.reflect.Type;
 
 public class ScheduledNotificationReceiver extends BroadcastReceiver {
 
-
     @Override
     public void onReceive(final Context context, Intent intent) {
         NotificationManagerCompat notificationManager = NotificationManagerCompat.from(context);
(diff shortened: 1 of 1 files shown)


Vanethos:
This is due to:

```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.setupNotificationChannel(:46)
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.createNotification(:4)
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.showNotification()
       at com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver.onReceive(:77)
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:3255)
       at android.app.ActivityThread.-wrap17()
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1680)
       at android.os.Handler.dispatchMessage(Handler.java:105)
       at android.os.Looper.loop(Looper.java:164)
       at android.app.ActivityThread.main(ActivityThread.java:6565)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
```

MaikuB:
Whilst there is an explanation on what happens from deserialising, what are the steps to reproduce this? I want to make sure the reason behind this is known. This fix would still prevent a notification from being shown

MaikuB:
To elaborate more on what I mean, whilst you've provided a stack trace and said the following

> This is due to the fact that the notification may be of incorrect json type or we may receive an empty object {} which will lead to nullpointer exceptions down the line

What I'd like to know is how and why this happens as this PR may not be the appropriate fix

Lima21:
Hi @MaikuB, I have this problem for a while and I've check all the example and use as a reference, I thought that can be related with permissions but after testing that in release mode I didn't receive any error from my phone, so I asked @Vanethos for help :)
I've some users using my app and this only happens in 0.0011% of users.

I just set the notifications with the showDailyAtTime.
Thank you! :slight_smile:

MaikuB:
@Lima21 I suspected that might be the case but at the moment this PR looks like a "reaction" to a stack trace when the problem highly likely from not having configured the all of the rules for a release build at some point. That would therefore affect how data is deserialised. If you kept your app in source control then this something you can confirm by checking the history and test as well.

At the moment I don't see much reason to accept the PR (sidenote: it fails to build by the way). Currently the approach taken is a bit similar to an app that has logic that passes null values around that leads to errors and rather than preventing that from happening, a try/catch block is put in. The explanations given imply that the Gson library can have random behaviour without details on why. Given it's a widely used library, I doubt that is the case too as it would erode trust in the API.

MaikuB:
Closing due to lack of further updates and explanation

Lima21:
This fix my problem and I don't get any complaint from users

Issue #459, opened 2020-01-28
Scheduling multiple notifications is slow on Android


hurrba:
## Problem
Scheduling multiple notifications has quite an overhead on Android, and can take several seconds on older phones. 
This doesn't seem to be an issue on iOS devices, possible due to lack of bridge overhead and optimized data (de-)serialization. 

### Use case
We have business-logic that requires us to calculate custom intervals for scheduling notifications (e.g. monthly interval, birth-control cycle). Because of this, we need to use the `schedule` function to schedule each notification individually. 

We often call the `schedule` function with 60 notifications (we've set this as a max since iOS has a limit of 64), resulting in a noticeable lag on newer phones, and occasional ANRs on older phones. 

## Possible solution
Create the option to bulk/batch schedule notifications. 

I'll create a pull request with proposed changes to allow for bulk/batch scheduling.

## Related issues
#242

MaikuB:
I remember that issue being raised. I see what you describe being a more specialised case that developers should implement on their own should they need it. Using platform channels is more expensive in general that apps that are concerned about the performance to extent would likely need to look at batch operations for all plugins. This leads to an awkward API IMO and I wasn't thinking this plugin is going to help solve every use case. As such I'm not looking at a PR.

I think it would be worth looking more if the performance is strictly around `schedule` or if it applies to `show` as well. If it's just the former then perhaps changing how the notifications are saved to shared preferences would help. Commenting out the Java code on saving to shared preferences would be a way to check as well to see if it makes a significant difference to the `schedule` method

hurrba:
Thanks for the quick reply. 

I understand your concerns. However, this is a crucial feature for us, so we're going ahead and implementing it in a fork for now. We don't think it makes the API as awkward as one could fear, and I'll make a PR that I hope you will take a look at anyway.

We've looked into the performance, and there are basically two main culprits: communication over the bridge and instantiation of the `GsonBuilder` [here](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L168) (being called multiple times).

MaikuB:
Sorry should clarify what I meant earlier is that not looking at merging one at this stage. But yes do open a PR, be interested to see what you come up with as well. Ignoring the batch scheduling, you raise a good point that perhaps `GsonBuilder` could just be created once. That could be a improvement done on it's own

[...]

PR #683, opened 2020-06-26
[flutter_local_notifications] Bulk scheduling

hurrba:
Adds functionality to bulk schedule notifications.

**Related issue:**
#459


MaikuB:
Looks like this is failing a number of checks. Can you take a look at addressing the reported issues?

hurrba:
I was a bit quick on the trigger. will fix the linting and tests.

MaikuB:
Thanks for the fixes. I'll likely not get to take a closer look at the PR until later in the week

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -282,7 +282,15 @@ private static void scheduleNotification(Context context, final NotificationDeta
         }
     }
 
-    private static void zonedScheduleNotification(Context context, final NotificationDetails notificationDetails, Boolean updateScheduledNotificationsCache) {
+    private static void scheduleNotificationsAndUpdateCache(Context context, final List<NotificationDetails> notificationDetails) {
+        for (NotificationDetails details : notificationDetails) {
```
MaikuB, line 286: IMO, saving should happen after a scheduling a single notification so it's treated like an atomic transaction. For example, if scheduling was to fail halfway through the collection for some reason, the ones that were successfully scheduled should have been saved to shared preferences

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -349,42 +325,12 @@ class IOSFlutterLocalNotificationsPlugin
...
-              uiLocalNotificationDateInterpretation.index,
-        }
-          ..addAll(scheduledDate.toMap())
-          ..addAll(scheduledNotificationRepeatFrequency == null
-              ? <String, Object>{}
-              : <String, Object>{
-                  'scheduledNotificationRepeatFrequency':
-                      scheduledNotificationRepeatFrequency.index
-                }));
+  Future<void> zonedSchedule(List<NotificationData> notifications) async {
```
MaikuB, line 328: Can you make bulk scheduling a method on it's own? Currently it's done as a breaking change when it's more of a helper/extension and this change would force everyone that schedules a single notification to pass a list

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -542,35 +488,12 @@ class MacOSFlutterLocalNotificationsPlugin
...
-          'payload': payload ?? '',
-        }
-          ..addAll(scheduledDate.toMap())
-          ..addAll(scheduledNotificationRepeatFrequency == null
-              ? <String, Object>{}
-              : <String, Object>{
-                  'scheduledNotificationRepeatFrequency':
-                      scheduledNotificationRepeatFrequency.index
-                }));
+  Future<void> zonedSchedule(List<NotificationData> notifications) async {
```
MaikuB, line 491: Concerned about the naming of `NotificationData` as it's very similar to `NotificationDetails`. Do you have other ideas for the name?

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -1,11 +1,11 @@
 import 'dart:async';
 
-import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_local_notifications_platform_interface/flutter_local_notifications_platform_interface.dart';
-import 'package:timezone/timezone.dart';
 
+import '../flutter_local_notifications.dart';
```
MaikuB, line 6: What was this import needed for? Can you make this import the specific file that has the type that is needed as opposed to the main library for the plugin itself

flutter_local_notifications/lib/src/notification_details.dart
```diff
@@ -19,3 +31,73 @@ class NotificationDetails {
   /// Notification details for macOS.
   final MacOSNotificationDetails macOS;
 }
+
+class NotificationData {
```
MaikuB, line 35: Not sure what this might be renamed to but can you move this to another file?

flutter_local_notifications/lib/src/notification_details.dart
```diff
@@ -19,3 +31,73 @@ class NotificationDetails {
...
+    if (platform.isIOS) {
+      _specs = platformSpecifics?.iOS?.toMap();
+    }
+    if (platform.isMacOS) {
+      _specs = platformSpecifics?.macOS?.toMap();
+    }
+    return _specs ?? <String, dynamic>{};
+  }
+
+  Map<String, dynamic> toMap() {
```
MaikuB, line 78: All mapping code has been moved to be extension methods (e.g. in `method_channel_mappers.dart`. Can you move this code as well? The reason for doing this was to not expose `toMap()` in the public API surface as it's only used internally

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
hurrba: good point. however, the saving was one of the main time-consumers, and iirc it can take up to 50ms. 
a solution could be to add each successfully scheduled notification to a temporary list, and then save this in one go.

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
hurrba: yeah I wasn't completely happy about that one either. `Notification` is already taken in Flutter, so users of this lib would have to import with a prefix, which (I assumed) isn't desired. 

how about `LocalNotification`?

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
hurrba: isn't this class abstracted away from users? 

I've kept the signature as-is for `FlutterLocalNotificationsPlugin.zonedSchedule` for scheduling a single notification, and added [`FlutterLocalNotificationsPlugin.scheduleNotifications`](https://github.com/MaikuB/flutter_local_notifications/pull/683/files#diff-cda68032e4b223cfb99a263bd79ecf37R254) for multiple notifications.

MaikuB: These classes aren't abstracted away by the cross-platform facing API if that's what you mean. Developers can choose to use the platform-specific implementation of the plugin to access methods that are specific to a platform. At a later point `zonedSchedule` will be part of the platform interface as well. Haven't done so yet as the beta release to get some feedback on this

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
MaikuB: Hmm I might have to think more about this. One option is to perhaps rename the existing `NotificationDetails`

MaikuB: Unfortunately I can't think of a name and the more I think I about this and the PR, the more I wonder about if this should be PR should be part of the plugin.

The options for the name so far are generic that imply it could be used for other methods but it can't be.

The introduction of the class itself brings a bit inconsistency on how notification content is passed to be displayed. If a new method is introduced as part of the cross-platform facing API for displaying notifications, it should be something that can be added to the platform interface as well (note: `schedule` wasn't added as it was going to deprecated). However, this can't really be because the object it requires contains knowledge of other platforms. This may well be to do with how the reason this method exists right now is because it's a "workaround" for a performance issue using platform channels on Android. It would not otherwise have been part of the "core" APIs for the plugin.

With all the above in mind and unless you have ideas to address what I mentioned without breaking changes, I'm currently leaning towards leaving this out and including a link to this PR in the readme for those that would need this level of performance optimisation

MaikuB:
Closing this as I've decided not to merge this based on some of the points I'd raised and in the related thread, particularly how it's not a core method and only exists due to the cost of using platform channels on Android. Thanks for your efforts though. I'll add a link to this PR in the readme so others can refer it if you they need this level of optimisation for their applications

Issue #690, opened 2020-06-30
Interested in a PR to support getActiveNotifications?


vkammerer:
Hello @MaikuB ,

Would you be interested in a PR that adds support for [Android's getActiveNotifications](https://developer.android.com/reference/android/app/NotificationManager#getActiveNotifications()) ?
The implementation could check the API Level and return an error message if the user is under version 23.

MaikuB:
If there's a way for developers to check the API level and not just rely on the error then yeah I'm happy to take a look at it. Haven't use it for checking API level but it looks like the `device_info` plugin can do this. If so, updating the example app to make use of that would be awesome :)

MaikuB:
Forgot to add, if you open a PR can you do that against the `timezone_macos_support` branch? That's the branch I'm using for the beta release

vkammerer:
Ok great, so just to clarify, what do you think if I:
- add a new method in [flutter_local_notifications_plugin.dart](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart) that would look like
```dart
Future<List<ActiveNotification>> getActiveNotifications() async {
    if (_platform.isAndroid) {
      await resolvePlatformSpecificImplementation<
              AndroidFlutterLocalNotificationsPlugin>()
          ?.getActiveNotifications();
    } else {
      return null;
    }
}
```
- implement the method `getActiveNotifications` that would return the following values:
```dart
[Instance of ActiveNotification] // with the active notifications if any are present 
[] // if there is no active notification
null // if the API level is under 23
```
- implement a new button in the example app, which would:
  - first query the Android API Level with the `device_info` plugin
  - then either:
    - display a message "ActiveNotifications is available only for Android API Level 23 +
and above" if the user is under 23
    - or display a message with the notifications data

MaikuB:
With the federated plugin implementation (see this [doc](https://docs.google.com/document/d/1LD7QjmzJZLCopUrFAAE98wOUQpjmguyGTN2wd_89Srs/edit) for more info if you're interested), the shift is to move away from having no-op methods in the cross-platform facing API to have it being a monolith. As such,  `getActiveNotifications()` should be left out of the `FlutterLocalNotificationsPlugin` class. Developers do have the ability to get the underlying platform-specific implementation to access platform-specific methods. Everything else sounds good to me.  Wasn't mentioned happy to have the readme updated to include mention of the feature as well :)

[...]

PR #704, opened 2020-07-09
[flutter_local_notifications] Adds support for Android's getActiveNotifications

vkammerer:
This PR follows up on the discussion initiated in https://github.com/MaikuB/flutter_local_notifications/issues/690.

@MaikuB I followed your [last comment](https://github.com/MaikuB/flutter_local_notifications/issues/690#issuecomment-653266829) suggestion and a `PlatformException` is thrown when the Android API level is under 23.

I'm a bit unsure which fields should be part of the `ActiveNotification` class: I have currently only implemented `id`, `channelId`, `title` and `body`, but would be happy to add more if you think it is relevant. Indeed, since we have access to the `Notification` instance on the Android's side, we could get more data.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -106,7 +106,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
     private static final String INVALID_SOUND_ERROR_CODE = "INVALID_SOUND";
     private static final String INVALID_LED_DETAILS_ERROR_CODE = "INVALID_LED_DETAILS";
     private static final String GET_ACTIVE_NOTIFICATIONS_ERROR_CODE = "GET_ACTIVE_NOTIFICATIONS_ERROR_CODE";
-    private static final String GET_ACTIVE_NOTIFICATIONS_ERROR_MESSAGE = "Android version must be at least 23 to use getActiveNotifications";
+    private static final String GET_ACTIVE_NOTIFICATIONS_ERROR_MESSAGE = "Android version must be at least 6.0 to use getActiveNotifications";
     private static final String INVALID_LED_DETAILS_ERROR_MESSAGE = "Must specify both ledOnMs and ledOffMs to configure the blink cycle on older versions of Android before Oreo";
     private static final String NOTIFICATION_LAUNCHED_APP = "notificationLaunchedApp";
     private static final String INVALID_DRAWABLE_RESOURCE_ERROR_MESSAGE = "The resource %s could not be found. Please make sure it has been added as a drawable resource to your Android head project.";
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -16,6 +16,9 @@ import android.media.AudioAttributes;
 import android.media.RingtoneManager;
 import android.net.Uri;
 import android.os.Build;
+import android.os.Build.VERSION;
+import android.os.Build.VERSION_CODES;
+import android.service.notification.StatusBarNotification;
 import android.text.Html;
 import android.text.Spanned;
 
@@ -84,6 +87,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
     private static final String INITIALIZE_METHOD = "initialize";
     private static final String CREATE_NOTIFICATION_CHANNEL_METHOD = "createNotificationChannel";
     private static final String DELETE_NOTIFICATION_CHANNEL_METHOD = "deleteNotificationChannel";
+    private static final String GET_ACTIVE_NOTIFICATIONS_METHOD = "getActiveNotifications";
     private static final String PENDING_NOTIFICATION_REQUESTS_METHOD = "pendingNotificationRequests";
     private static final String SHOW_METHOD = "show";
     private static final String CANCEL_METHOD = "cancel";
@@ -101,6 +105,8 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
     private static final String INVALID_BIG_PICTURE_ERROR_CODE = "INVALID_BIG_PICTURE";
     private static final String INVALID_SOUND_ERROR_CODE = "INVALID_SOUND";
     private static final String INVALID_LED_DETAILS_ERROR_CODE = "INVALID_LED_DETAILS";
+    private static final String GET_ACTIVE_NOTIFICATIONS_ERROR_CODE = "GET_ACTIVE_NOTIFICATIONS_ERROR_CODE";
+    private static final String GET_ACTIVE_NOTIFICATIONS_ERROR_MESSAGE = "Android version must be at least 23 to use getActiveNotifications";
     private static final String INVALID_LED_DETAILS_ERROR_MESSAGE = "Must specify both ledOnMs and ledOffMs to configure the blink cycle on older versions of Android before Oreo";
     private static final String NOTIFICATION_LAUNCHED_APP = "notificationLaunchedApp";
     private static final String INVALID_DRAWABLE_RESOURCE_ERROR_MESSAGE = "The resource %s could not be found. Please make sure it has been added as a drawable resource to your Android head project.";
@@ -871,6 +877,9 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
             case DELETE_NOTIFICATION_CHANNEL_METHOD:
                 deleteNotificationChannel(call, result);
                 break;
+            case GET_ACTIVE_NOTIFICATIONS_METHOD:
+                getActiveNotifications(result);
+                break;
             default:
                 result.notImplemented();
                 break;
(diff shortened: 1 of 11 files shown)


MaikuB:
Thanks for the PR. Would you be able to change the target branch to `timezone_macos_support` instead of `master` so I can put this feature into the current beta? Will go through the PR more when it's done.

Regarding adding more info, do you know if it's possible to get the payload, as defined by the plugin, and include that as part of the info that is returned?

vkammerer:
I have just changed the target branch to `timezone_macos_support`, and the diff is identical.

I had tried to retrieve the payload, which would be great to have. But I didn't manage to, since the notification's `PendingIntent` doesn't allow access to its `Intent` which contains the payload data.. Can you think of another way to retrieve it?

MaikuB:
There are some merge conflicts that need to be resolved though.

Regarding the payload, I admittedly forgot about the intent part so don't know of another way at the moment.

vkammerer:
Oops I hadn't seen the merge conflicts.. I have just resolved them pushed an updated version.

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -101,6 +105,8 @@
     private static final String INVALID_BIG_PICTURE_ERROR_CODE = "INVALID_BIG_PICTURE";
     private static final String INVALID_SOUND_ERROR_CODE = "INVALID_SOUND";
     private static final String INVALID_LED_DETAILS_ERROR_CODE = "INVALID_LED_DETAILS";
+    private static final String GET_ACTIVE_NOTIFICATIONS_ERROR_CODE = "GET_ACTIVE_NOTIFICATIONS_ERROR_CODE";
+    private static final String GET_ACTIVE_NOTIFICATIONS_ERROR_MESSAGE = "Android version must be at least 23 to use getActiveNotifications";
```
MaikuB, line 109: 23 is the API level, 6.0 is the Android version. Could you update this error to refer to version number as that's what I've been using in the plugin as there will be developers who haven't worked on Android directly to be familiar with API levels

flutter_local_notifications/lib/src/platform_specifics/android/status_bar_notification.dart
```diff
@@ -0,0 +1,22 @@
+class ActiveNotification {
```
MaikuB, line 1: Can you rename the file to match the name of the class?

flutter_local_notifications/lib/src/platform_specifics/android/status_bar_notification.dart
```diff
@@ -0,0 +1,22 @@
...
+    this.title,
+    this.body,
+  );
+
+  final int id;
+  final String channelId;
+  final String title;
+  final String body;
+
+  toString() {
```
MaikuB, line 14: Can this be removed? Haven't implemented this for any other class in the plugin and removing it would allow users of the plugin to define their own implementation of `toString` e.g. via extension methods?

flutter_local_notifications/lib/src/platform_specifics/android/status_bar_notification.dart
```diff
@@ -0,0 +1,22 @@
+class ActiveNotification {
+  const ActiveNotification(
+    this.id,
+    this.channelId,
+    this.title,
+    this.body,
+  );
+
+  final int id;
```
MaikuB, line 9: Missing API docs for the property. Can refer to the existing public API docs to copy and paste them if it makes sense to do so. `channelId` will need special mention that it is only returned on Android 8.0+

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
vkammerer: Sure, maybe we can discuss the exact wording here before I push changes?

I'm not sure I understood which number you prefer: the API level, or the Android version? I've searched for "Android" in your repository and the only occurence I could find was the README that says 
```markdown
**Android API 16+** (4.1+, the minimum version supported by Flutter).
```
so should I replace it with (version 1):
```java
"Android API Level must be at least 23 to use getActiveNotifications"
```
or with (version 2)
```java
"Android version must be at least 6.0 to use getActiveNotifications"
```

flutter_local_notifications/lib/src/platform_specifics/android/status_bar_notification.dart
vkammerer: sure

flutter_local_notifications/lib/src/platform_specifics/android/status_bar_notification.dart
vkammerer: sure

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
MaikuB: Sorry for the confusing. You're right that the readme has both in there though for public API docs I've simplified it refer to Android version i.e. go with version 2

flutter_local_notifications/lib/src/platform_specifics/android/status_bar_notification.dart
vkammerer: Ok I added documentation comments to all the properties. Let me know if it's ok like that

vkammerer:
I just pushed https://github.com/MaikuB/flutter_local_notifications/pull/704/commits/deea969f67bbfa3913bc31ceace8dac4688b1ec6 , which also updates the README with a reference and usage example of the new feature, as you had mentioned in https://github.com/MaikuB/flutter_local_notifications/issues/690#issuecomment-653258905

MaikuB:
Hmm the analyse task is failing is a weird way, maybe I'll need to remove it until I figure out what's going on

vkammerer:
By the way I didn't know someone else than me could add commits to my PRs. I've seen that your last commits have been added to my repo in the branch `android_get_active_notifications`.
Which git command did you use to do that?

flutter_local_notifications/example/lib/main.dart
```diff
@@ -1178,6 +1185,84 @@ class _HomePageState extends State<HomePage> {
...
+          Text(
+            'Active Notifications',
+            style: TextStyle(fontWeight: FontWeight.bold),
+          ),
+          Divider(color: Colors.black),
+          if (activeNotifications.isEmpty) Text('No active notification'),
+          if (activeNotifications.isNotEmpty)
+            ...activeNotifications
+                .map(
+                  (ActiveNotification aN) => Column(
```
MaikuB, line 1242: nit: Convention being used in the existing codebase is to stick with a single letter for variables used in an anonymous function/lambda expression. Having said that, should be able to use the `collection for` that is idiomatic to Dart and removes the need for spread, map and toList operations.

See this article for an example if you haven't come across it before https://medium.com/dartlang/announcing-dart-2-3-optimized-for-building-user-interfaces-e84919ca1dff

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -255,6 +256,19 @@ class AndroidFlutterLocalNotificationsPlugin
   Future<void> deleteNotificationChannel(String channelId) =>
       _channel.invokeMethod('deleteNotificationChannel', channelId);
 
+  Future<List<ActiveNotification>> getActiveNotifications() async {
+    final List<Map<dynamic, dynamic>> activeNotifications =
+        await _channel.invokeListMethod('getActiveNotifications');
+    return activeNotifications
+        .map((activeNotification) => ActiveNotification(
```
MaikuB, line 263: Same comment as above on variable name. Also running `flutter analyze` shows an error where the type hasn't been specified

flutter_local_notifications/lib/src/platform_specifics/android/active_notification.dart
```diff
@@ -0,0 +1,20 @@
...
+    this.id,
+    this.channelId,
+    this.title,
+    this.body,
+  );
+
+  /// The notification's id.
+  final int id;
+
+  /// The notification channel's id - Returned only on Android 8.0+
```
MaikuB, line 12: nit: I've been putting things similar _Returned only on Android 8.0+_ in a separate paragraph as part of following these guidelines https://dart.dev/guides/language/effective-dart/documentation#do-start-doc-comments-with-a-single-sentence-summary

flutter_local_notifications/example/lib/main.dart
```diff
@@ -1178,6 +1185,84 @@ class _HomePageState extends State<HomePage> {
...
+                    crossAxisAlignment: CrossAxisAlignment.start,
+                    children: <Widget>[
+                      Text(
+                        'id: ${aN.id.toString()}\n'
```
MaikuB, line 1246: `toString()` is redundant when doing string interpolation

flutter_local_notifications/example/lib/main.dart
```diff
@@ -1178,6 +1185,84 @@ class _HomePageState extends State<HomePage> {
...
+                        'channelId: "${aN.channelId}"\n'
```
MaikuB, line 1247: I'd suggest removing the double quotation marks as it might confuse users into thinking they are part of the value itself. There isn't a need to distinguish if a value is a string when it's present in the app if that was the reason why they were placed there

flutter_local_notifications/example/lib/main.dart
```diff
@@ -1178,6 +1185,84 @@ class _HomePageState extends State<HomePage> {
...
+                    ],
+                  ),
+                )
+                .toList()
+        ],
+      );
+    } on PlatformException catch (error) {
+      return Text(
+        'Error calling "getActiveNotifications"\n'
+        'code: "${error.code}"\n'
```
MaikuB, line 1261: See comment above on the double quotation marks

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -255,6 +256,19 @@ class AndroidFlutterLocalNotificationsPlugin
   Future<void> deleteNotificationChannel(String channelId) =>
       _channel.invokeMethod('deleteNotificationChannel', channelId);
 
+  Future<List<ActiveNotification>> getActiveNotifications() async {
+    final List<Map<dynamic, dynamic>> activeNotifications =
```
MaikuB, line 260: Sorry missed this last time but does this work if `dynamic` was replaced with `Object`?

MaikuB:
Regarding pushing changes to the PR, no special git commands are needed for that. When you create a PR, there's a checkbox on if you allow edits to be done by maintainers. If that's enabled, I can use the GitHub client to download your fork/branch

image

flutter_local_notifications/example/lib/main.dart
vkammerer: I wasn't aware of `collection for` which I'll definitely use from now on! I've implemented it in this PR

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
vkammerer: I've added `// ignore: always_specify_types` before that line as adding the type conflicts with another rule

flutter_local_notifications/lib/src/platform_specifics/android/active_notification.dart
vkammerer: ok sure

flutter_local_notifications/example/lib/main.dart
vkammerer: ok

flutter_local_notifications/example/lib/main.dart
vkammerer: ok

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
vkammerer: Trying to cast to another type throws an Exception at runtime so this signature should be kept for now

flutter_local_notifications/example/lib/main.dart
vkammerer: ok

vkammerer:
Ok good to know about the PR updates you can make.
I've added a basic test that looks like the others in the list, and ran them and they all pass.

MaikuB:
Once the build passes, I'll merge this in. Will likely do some general cleanup separately before releasing this out. Thanks again :)

Issue #74, opened 2018-07-08
Callback when notification is dismissed by user


radzish:
If this applies to android only, please do it for at least android.

MaikuB:
This is dependent on the Flutter team finishing the work on headless Dart execution. Also, I don't see this feature adding a lot of value to a lot of other devs (unless your original post gets upvoted by others though I have other priorities at the moment) so I would suggest once the Flutter team has finished their work (at which point I'll work on custom notifications actions) that you submit a PR for this

ciw1973:
Surely the primary use case for local notifications is to allow triggering of functionality in the app which scheduled the notification in the first place?

MaikuB:
I think you've misunderstood what I was getting it. What I meant is that this functionality is of lower priority compared to other features that are used more often e.g. custom actions. I don't see many apps that do something when a notification is dismissed, not to say they don't exist.

Bear in mind i'm trying to prioritise what I can work on given this is something I'm working on in my spare time. Currently I've already run into issues doing #21 that would help set the framework for #17 and this ticket. Hence why I'm asking if other devs in the community could assist in implementing features.

ciw1973:
Yes, I'd misread your previous comment, which I why I deleted mine shortly after I posted it.

Thanks for the clarification.

matanshukry:
I needed it so I implemented it ( #722 ).
Like I said in there I don't have a Mac so it's an Android only PR, but hopefully someone else can do the other part ?

matanshukry:
@MaikuB My use case for this feature is to create a grouping for message notifications.

From what I could gather, I must have a summary for all notifications I want to group together.
Also, if one of the notifications is dismissed, I should update the summary (since it contains the text for the dismissed notification, which isn't longer in the group).

Is there a way around it / another way you recommend?
or perhaps one of the details I understood is incorrect?

MaikuB:
If the changes you have in your PR work for your scenarios (i.e. you don't care that it won't work when the app is killed) then I'd suggest to continue using your own fork. If you do need it working even when the app isn't running then the only way is likely what I've suggested [here](https://github.com/MaikuB/flutter_local_notifications/issues/17#issuecomment-660717981), where upi move more of the notification logic to be on the Android side i.e. handle what happens when a notification is dismissed without trying to run any Dart code

matanshukry:
I went with the fork side.

For anyone else that's interested:
[...]

PR #722, opened 2020-07-20
[flutter_local_notifications] Added callback for dismissing a notification

matanshukry:
Notes:
- I don't have a Mac so this is an Android only version. I hope the iOS version will be not be complex either
- I couldn't think of a way to show the dismiss in the UI, so right now you can just see it in the console
  - I thought about using a Snackbar but there isn't a scaffold, and I wasn't sure it's worth adding

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -74,6 +74,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
     private static final String DRAWABLE = "drawable";
     private static final String DEFAULT_ICON = "defaultIcon";
     private static final String SELECT_NOTIFICATION = "SELECT_NOTIFICATION";
+    private static final String DISMISS_NOTIFICATION = "DISMISS_NOTIFICATION";
     private static final String SCHEDULED_NOTIFICATIONS = "scheduled_notifications";
     private static final String INITIALIZE_METHOD = "initialize";
     private static final String CREATE_NOTIFICATION_CHANNEL_METHOD = "createNotificationChannel";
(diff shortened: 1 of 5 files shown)


MaikuB:
Your PR only works when the app is actually running and is why I haven't implemented it myself. The reason why #74 was marked as blocked and is mentioned in the thread that it should also when the app isn't running via headless Dart execution.  This is something that has issues at the moment and I'm not looking to merge such changes until it works in that scenario unless the Flutter team decides not to support headless execution anymore. This is something that was being discussed amongst the team to from what I know

matanshukry:
@MaikuB One could implement it in a service, much like FCM does.
Or the issues you're talking about also apply there? If so, what issues are those?

MaikuB:
Also applies to FCM and same reason why #17 is blocked, which goes to this issue I raised on the Flutter repo itself https://github.com/flutter/flutter/issues/23904

matanshukry:
@MaikuB Would you care to explain how does it apply to FCM exactly?
A quick test on my app that uses FCM seems to work:

1. Sending message when app is in the foreground (works)
2. Terminating the app
3. Sending message when app is terminated (works, and launches the app)
4. Sending message when app is in the foreground (works)

The repo you provided in the flutter issue isn't valid anymore, so at least that is what I understood from your explanation.
Correct me if I'm wrong.

MaikuB:
Actually, I forgot FCM has different callbacks for different scenarios. In that case it _might_ not be an issue as they appear to detect if the app is running or not. Though if you look at the [last comment](https://github.com/flutter/flutter/issues/23904#issuecomment-652148410), you can see another user being impacted by the issue where once background execution works, it stops normal callbacks from working

> The repo you provided in the flutter issue isn't valid anymore, so at least that is what I understood from your explanation.

I'm not sure what point you were trying to make here. What I meant there is that I accidentally deleted the repository that demonstrates the problem that still exists. Perhaps you confused repo with repro steps? The issue

matanshukry:
> I'm not sure what point you were trying to make here. What I meant there is that I accidentally deleted the repository that demonstrates the problem that still exists. Perhaps you confused repo with repro steps? The issue

I meant I didn't fully understand the context since I couldn't read the code, so I might of missed something.

> Though if you look at the last comment, you can see another user being impacted by the issue where once background execution works, it stops normal callbacks from working

I did see it, but I'm using FCM with many other plugins (and channels) and never encountered such an issue, nor have I read about it somewhere. If it is a bug it seems rare enough that it could be ignored (at least in my case).

> In that case it might not be an issue as they appear to detect if the app is running or not
Assuming I implement it like they did then, and it would work in the background - will you accept it into this plugin?

Keep in mind I can't implement the iOS side.

MaikuB:
> Assuming I implement it like they did then, and it would work in the background - will you accept it into this plugin?

Not looking to at this stage as the related issue is still an issue. It also adds maintenance burden where I need to consider what the migration path would be when general background execution land as callbacks need to be saved in shared preferences at the moment. PR would need an iOS and macOS implementation (I added support to macOS in a prerelease version)

If you what you have works for you then I'd suggest you keep your own fork going. Would be interested in seeing how it goes, particularly to see if you run into issues

matanshukry:
@MaikuB I completed the fork and the changes and it's now working correctly. Feel free to take a look and comment, would love to hear your opinion.

Also, regarding the last comment in the Dart PR - I've encountered the issue as well, but it seems to be an issue with FCM specifically. there is also a PR that fixes it, although it's not in. Thought you might be interested to know that:
https://github.com/FirebaseExtended/flutterfire/pull/1917

pranavo72bex:
Any Update regarding this PR?

PR #730, opened 2020-07-24
[flutter_local_notifications] Android Full-Screen Notification

### LINKED ISSUE #651: Full screen notification
Please consider working for Full screen notification (Full screen intent in android) works in Android

nadavfima:
This should take care of functionality.
Let's see how the tests pan out.

Closes #651


nadavfima:
Seems like that CodeFactor check is not related to anything I changed.

flutter_local_notifications/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Sat Jul 04 21:26:33 IDT 2020
+distributionBase=GRADLE_USER_HOME
```
MaikuB, line 2: Can you revert adding this in? My understanding is that this would start locking the plugin with specific Gradle distribution versions. A number of the first-party plugins leave this out too

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -216,6 +217,9 @@ class AndroidNotificationDetails {
   /// Refer to Android notification API documentation at https://developer.android.com/reference/androidx/core/app/NotificationCompat.html#constants_2 for the available categories
   final String category;
 
+  /// Specifies whether the notification should try to launch the intent as soon as it triggeres
```
MaikuB, line 220: Spelling mistake and sentence should end with a full stop

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -44,7 +44,8 @@ class AndroidNotificationDetails {
     this.visibility,
     this.timeoutAfter,
     this.category,
-    this.additionalFlags,
+    this.isFullScreen = false,
```
MaikuB, line 47: Rename to more closely match the property on Android side i.e. `fullScreenIntent`. The parameter passed via the method channel should be renamed to match it as well

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -179,6 +183,19 @@ private static Notification createNotification(Context context, NotificationDeta
             builder.setWhen(notificationDetails.when);
         }
 
+        if (notificationDetails.isFullScreen){
+            builder.setFullScreenIntent(pendingIntent, true);
+
+            // wake device
+            PowerManager powerManager = (PowerManager) context.getSystemService(Context.POWER_SERVICE);
```
MaikuB, line 190: Is the rest of the code here actually needed? From what I'm reading in the [docs](https://developer.android.com/training/notify-user/time-sensitive), there's no mention of it being needed

MaikuB:
Fixed the macOS issues that got reported in #734 so you'll need to update your fork. There was one issue that came from this PR though

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
nadavfima: > Is the rest of the code here actually needed? From what I'm reading in the [docs](https://developer.android.com/training/notify-user/time-sensitive), there's no mention of it being needed

I'll verify, but I think it is required for the screen to be able to turn-on and stay on.

flutter_local_notifications/android/gradle/wrapper/gradle-wrapper.properties
nadavfima: 100% right

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
nadavfima: I think you're right.

nadavfima:
Adding to example now.

nadavfima:
I think I'm done

flutter_local_notifications/lib/src/platform_specifics/android/method_channel_mappers.dart
```diff
@@ -234,7 +234,8 @@ extension AndroidNotificationDetailsMapper on AndroidNotificationDetails {
         'visibility': visibility?.index,
         'timeoutAfter': timeoutAfter,
         'category': category,
-        'additionalFlags': additionalFlags
+        'fullScreen': fullScreenIntent,
```
MaikuB, line 237: missed renaming the string to `fullScreenIntent` so both the Dart and Android side match

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
```diff
@@ -108,7 +108,7 @@
     private static final String SCHEDULED_DATE_TIME = "scheduledDateTime";
     private static final String TIME_ZONE_NAME = "timeZoneName";
     private static final String SCHEDULED_NOTIFICATION_REPEAT_FREQUENCY = "scheduledNotificationRepeatFrequency";
-
+    private static final String FULL_SCREEN = "fullScreen";
```
MaikuB, line 111: This will need renaming since the Dart prop has been renamed

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -15,12 +15,15 @@
 import android.media.AudioAttributes;
 import android.media.RingtoneManager;
 import android.net.Uri;
+import android.os.Build;
```
MaikuB, line 18: Remove any imports that aren't needed

flutter_local_notifications/example/lib/main.dart
```diff
@@ -499,22 +505,51 @@ class _HomePageState extends State<HomePage> {
...
+            FlatButton(
+              child: Text('Cancel'),
+              onPressed: () {
+                Navigator.pop(context);
+              },
+            ),
+            FlatButton(
+              child: Text('OK'),
+              onPressed: () async {
+                Navigator.pop(context);
```
MaikuB, line 529: Popping first and then doing the work after it is a bit counter-intuitive. Can you instead use `zonedSchedule` to schedule a full-screen intent notification to take place in 3 seconds and pop after scheduling is done? That would help to guarantee the notification runs regardless of what state the app is in

flutter_local_notifications/example/lib/main.dart
```diff
@@ -499,22 +505,51 @@ class _HomePageState extends State<HomePage> {
...
   Future<void> _showNotificationWithNoBody() async {
     var androidPlatformChannelSpecifics = AndroidNotificationDetails(
         'your channel id', 'your channel name', 'your channel description',
         importance: Importance.max, priority: Priority.high, ticker: 'ticker');
     var platformChannelSpecifics = NotificationDetails(
       android: androidPlatformChannelSpecifics,
     );
-    await flutterLocalNotificationsPlugin.show(
-        0, 'plain title', null, platformChannelSpecifics,
+    await flutterLocalNotificationsPlugin.show(0, 'plain title', null, platformChannelSpecifics,
```
MaikuB, line 552: Hmm didn't expect the alignment to change as part of this PR. Can you run the formatter to make sure the code is formatted properly and there aren't any analysis issues?

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -216,6 +217,10 @@ class AndroidNotificationDetails {
   /// Refer to Android notification API documentation at https://developer.android.com/reference/androidx/core/app/NotificationCompat.html#constants_2 for the available categories
   final String category;
 
+  /// Specifies whether the notification should launch a full-screen intent as soon as it triggers.
+  /// Note: The system UI may choose to display a heads-up notification, instead of launching your full-screen intent, while the user is using the device.
```
MaikuB, line 221: A blank line should separate the two sentences like the rest of the API docs to follow [these guidelines](https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph)

flutter_local_notifications/README.md
```diff
@@ -132,6 +132,23 @@ If the vibration pattern of an Android notification will be customised then add
 <uses-permission android:name="android.permission.VIBRATE" />
 ```
 
+
+#### Full-Screen Intent Notifications
```
MaikuB, line 136: nit: use sentence casing instead of title casing

MaikuB:
Out of curiosity, are you able to let me know what your use case is for needing a full-screen intent?

flutter_local_notifications/example/lib/main.dart
nadavfima: I reverted this change. not sure why IntelliJ did that.
Also fixed a couple analysis issue with docs.

nadavfima:
> Out of curiosity, are you able to let me know what your use case is for needing a full-screen intent?

Sure. I'm working on a timer app, and I want the timer screen to pop-up when the timer is done - if the screen is off.

I do something similar with a Native Android alarm clock app, which also uses the `AlarmManager` along with FullScreen Intent.

That's where I got the WakeLock thingy from, but in the Alarm Clock it's actually used for a `Service` that runs in the background and plays the ringtone in an infinite loop, even if the activity was not started.

flutter_local_notifications/example/lib/main.dart
```diff
@@ -499,13 +505,48 @@ class _HomePageState extends State<HomePage> {
...
+            FlatButton(
+              child: Text('Cancel'),
+              onPressed: () {
+                Navigator.pop(context);
+              },
+            ),
+            FlatButton(
+              child: Text('OK'),
+              onPressed: () async {
+
```
MaikuB, line 529: nit: remove blank lines in this method

MaikuB:
PR looks good to me, though when I run it on my Pixel 4 XL there doesn't look like there's much difference as I keep getting the heads-up notification. I lock the screen after pressing OK to dismiss the dialog. I know that the OS may decide to show a standard notification but do you happen to have a recording of the example app to show what it's like when it's a full-screen intent/notification?

MaikuB:
FYI just pushed some changes after running the formatter (`flutter format .`). Guessing the changes that got applied before was due to a result of the formatter not being run at some point e.g. from another PR that code submitted. I'll update the PR template later to remind others to run it before submitting a PR

nadavfima:
> PR looks good to me, though when I run it on my Pixel 4 XL there doesn't look like there's much difference as I keep getting the heads-up notification. I lock the screen after pressing OK to dismiss the dialog. I know that the OS may decide to show a standard notification but do you happen to have a recording of the example app to show what it's like when it's a full-screen intent/notification?

That's interesting. I ran it on my Pixel 2 XL and it's running fine.
I'll try to record a video tomorrow.

So you lock (turn-off) the phone's screen, and then the notification would turn the screen on - but not launch the intent?

MaikuB:
That's right, I do have face unlock on my phone but I doubt that makes a difference

MaikuB:
Any updates on being able to capture a recording?

MaikuB:
Was able to reproduce this now after configuring the priority and importance of the notification. Not sure if this was something you had before and then forgot

niteshgrg:
@MaikuB This looks good. Is there some guide on how to use this properly i can see the example which looks good but i can gather by that it is also bounded by limited styling it will offer. Ideally in full screen i would want to style the complete screen right. Is there a way to pass complete screen style ?

Techno-Disaster:
@nadavfima @MaikuB Quick question, just adding `android:showWhenLocked="true"` to the .MainActivity results in the app showing on the lock screen all the time (reproducible with the example app), is it possible to only show the app (or full screen notification) on the lock screen when we get a notification?

Techno-Disaster:
@nadavfima @MaikuB Quick question, just adding `android:showWhenLocked="true"` to the .MainActivity results in the app showing on the lock screen all the time (reproducible with the example app), is it possible to only show the app (or full screen notification) on the lock screen when we get a notification?

PR #742, opened 2020-08-04
Added method to check if a channel is disabled on Android

vegidio:
Use case: the user disables the notification for a certain channel on Android settings. This method that I created can be used by the developer to detect that and let the user know that he needs to enable notifications for that channel before notifications can be sent again.


MaikuB:
This looks too specific to be included in the plugin given it relies on a specific property of the notification channel. A more generic approach is to return the details of a notification channel as it would avoid having to create separate methods for retrieving different properties. The other thing is the documentation [here](https://developer.android.com/reference/android/app/NotificationChannel#getImportance()) states it's possible that the importance level is "higher" than none and notifications won't be posted to the channel either

PR #748, opened 2020-08-10
Add monthly & yearly notifications

antoine92190:
Added support for monthly and yearly repeat notifications on both iOS and Android.

I tested the scheduled dates which are correct, however I could not confirm through tests that notifications are actually repeated every month/year.
Testing it on real device would take a long time (a month or a year), so if you have any guidance on how I could test and make sure this works, I will be happy to try it out.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -771,6 +771,12 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
             } else if(notificationDetails.scheduledNotificationRepeatFrequency == ScheduledNotificationRepeatFrequency.Weekly) {
                 LocalDateTime localDateTime = LocalDateTime.parse(notificationDetails.scheduledDateTime).plusWeeks(1);
                 return DateTimeFormatter.ISO_LOCAL_DATE_TIME.format(localDateTime);
+            } else if(notificationDetails.scheduledNotificationRepeatFrequency == ScheduledNotificationRepeatFrequency.Monthly) {
+                LocalDateTime localDateTime = LocalDateTime.parse(notificationDetails.scheduledDateTime).plusMonths(1);
+                return DateTimeFormatter.ISO_LOCAL_DATE_TIME.format(localDateTime);
+            } else if(notificationDetails.scheduledNotificationRepeatFrequency == ScheduledNotificationRepeatFrequency.Yearly) {
+                LocalDateTime localDateTime = LocalDateTime.parse(notificationDetails.scheduledDateTime).plusYears(1);
+                return DateTimeFormatter.ISO_LOCAL_DATE_TIME.format(localDateTime);
             }
         } else {
             if(notificationDetails.scheduledNotificationRepeatFrequency == ScheduledNotificationRepeatFrequency.Daily) {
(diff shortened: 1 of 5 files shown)


MaikuB:
Unfortunately I don't have any guidance on testing it on a real device. I'll take a look into the PR but what cases did your testing of dates cover and did it include making sure the dates calculated on both platforms match?

antoine92190:
I added 2 buttons in the `main` example file:
- `Schedule monthly 10:00:00 am notification in your local time zone`
- `Schedule yearly 10:00:00 am notification in your local time zone`
I tested using those buttons and played around with different dates.

Then I used the `Check pending notifications` button to get the scheduled notification and added some logs:
- iOS: `requests` in `pendingUserNotificationRequests` method
- Android: `scheduledNotification` in `pendingUserNotificationRequests` method

Note: `pendingUserNotificationRequests` method in both platforms is not returning the notification date. Maybe it could be useful to add that in the result.

MaikuB:
> Note: pendingUserNotificationRequests method in both platforms is not returning the notification date. Maybe it could be useful to add that in the result.

Dates aren't included as they can't be retrieved, especially on Android since the notification APIs don't support scheduling and requires using another Android API to schedule a notification

antoine92190:
On Android, I managed to retrieved scheduled date using `scheduledNotification.scheduledDateTime` in the below snippet. Am I missing something?

https://github.com/MaikuB/flutter_local_notifications/blob/dbe1e65429f647611f60f157c9177d3e46f8dcae/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L793

MaikuB:
Yes, my point was that the dates don't get returned by native APIs. The field you're referring to is custom data set by the plugin and there may be a  difference in when it fires due to battery optimisation.

There's also the matter of dealing with notifications scheduled with the old APIs exposed by the plugin that had issue with daylight savings. Perhaps it could be the decision is that dates won't be returned for notifications scheduled via the deprecated APIs. Even then, the new approach to schedule notifications requires timezones to be specified. The older iOS APIs don't support this well. Unless someone else can contribute a nice solution to cover these issues, I'd rather leave it out as it adding it in will most likely lead to more maintenance burden with bug reports etc

MaikuB:
I had to merge a fix I did on master, do you mind updating your branch? After that I'll play around with the PR bit more, will need to check things like how it behaves with boundary cases like when a recurring notification first triggers on February 29th and finding out when the next notification will be

antoine92190:
Done

MaikuB:
Thanks, I took a look at the PR and it looks like both platforms don't behave the same way. If the initial date and time is 10am on the 29th of February 2020,  the next time the notification fires on Android would be 29th of March. When I look at the next trigger date on iOS, it's showing up as 29th August 2020.

antoine92190:
Isn't it the same behaviour for the daily and weekly methods you implemented?

antoine92190:
If it is, then we can check for all repeat notifications if date is scheduled in the past.
Then we schedule the next notification from the current day (today), hence the next scheduled date will always be in the future.
I guess this check only needs to be implemented on Android, as it seems this is the default behaviour on iOS.

MaikuB:
> Isn't it the same behaviour for the daily and weekly methods you implemented?

Monthly is a different interval to both of those to begin it shouldn't behave the same way. Also ignore how I specified a date in the past. My point here is that the next trigger dates aren't the same

MaikuB:
Actually I see what you mean now, iOS will pick a date that is current. I wonder what if there's a way to calculate the next instance of a date using just the components i.e. without the calendar trigger

MaikuB:
I ended up turning back the date and time to be able to test with some dates in the past and there's still an issue where both platforms behave differently. If 31st of January 2020 is used for a monthly notification, the next instance on Android is 29th of February. On iOS, it's 1st of March. If it's a yearly notification for 29th of February 2020, next instance on Android is 28th of Feburary 2021 but on iOS it's 1st of February 2021

antoine92190:
> If it's a yearly notification for 29th of February 2020, next instance on Android is 28th of Feburary 2021 but on iOS it's 1st of February 2021

Did you mean that it's 1st of **March** 2021 on iOS?

MaikuB:
No it was February. It isn't that odd because in the iOS you have for yearly notifications you've only specified it to match on the month and time components. If you added day as well then off-memory it would go 1st of March 2021. Still, there are inconsistencies across both platforms. Do you know how this resolved this as I haven't come across a way to do so

antoine92190:
Ok got it. I'm not sure how to resolve this without some dirty checks and calculations.
I will dig into more when I have some time.
And I guess you are not confortable with saying "it behaves differently but that because of platform specifics", right? :)

MaikuB:
> And I guess you are not confortable with saying "it behaves differently but that because of platform specifics", right?

No as a cross-platform API is like an abstraction/refactoring of functionality that does the same thing across all platforms. Therefore, there's an expectation that it doesn't behave differently unless it's been told to do so e.g. similar to have there are optional arguments per platform in this plugin's APIs

PR #758, opened 2020-08-18
[flutter_local_notifications] Added an option for Person to display a circle icon.

firatcetiner:
I added a simple `Bitmap` implementation of converting the original image to its rounded circle version, on the `Android` side. The change is in the `Person` class'. The new `circleIcon` parameter will decide whether show `Person.icon` as a circle or not when `icon` is a `FlutterBitmapAssetAndroidIcon` or `BitmapFilePathAndroidIcon`, for all other types of `AndroidIcon` class, the `circleIcon` parameter will be ignored. (e.g always assigned false).

I should state that I'm not an Android expert on these topic, I needed to show the Person icon as a circle and I took the opportunity to implement this, it might not be the best solution but it works for me. I hope it can be useful for others as well.

Also, notice that I haven't added an sample usage of this new `circleIcon` on the `main.dart` in the example project because I didn't want to change the existing example code as it could have been produced some merge conflicts after the changes. So the best thing is to let contributors decide how to properly change the example app.

I tested it on a real device running the example application. Here is the result, Lunch bot's avatar changed into a circle avatar: 

<a href="https://ibb.co/ynLMB7b">example-sceenshot</a>

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -12,6 +12,11 @@ import android.content.SharedPreferences;
 import android.content.res.AssetFileDescriptor;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
+import android.graphics.Canvas;
+import android.graphics.Paint;
+import android.graphics.PorterDuff;
+import android.graphics.PorterDuffXfermode;
+import android.graphics.Rect;
 import android.media.AudioAttributes;
 import android.media.RingtoneManager;
 import android.net.Uri;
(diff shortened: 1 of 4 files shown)


MaikuB:
Thanks for submitting the PR. However, from what I can see this is custom logic that is specific to individual use cases. It's not part of the core responsibilities for a notifications plugin as applications could decide to use an arbitrary shape. Also I would've thought that the image itself could be made circular with transparency applied.

elkhalifte:
You actually don't need native code for this.

just by using this [plugin](https://pub.dev/packages/image) you can achieve the results.


`class Utils {
  static Future<String> downloadFile(String url, String filename) async {
    final dir = await getApplicationDocumentsDirectory();
    final filePath = '${dir.path}/$filename';
    final response = await http.get(Uri.parse(url));
    final file = File(filePath);
    final srcFile = await file.writeAsBytes(response.bodyBytes);
    final srcImg = decodeImage(srcFile.readAsBytesSync());

    var img = Image.from(srcImg!);

    img = copyCropCircle(
      img,
      // radius: img.width ~/ 1.5,
    );
     if (img != null) {

    final png = encodePng(img);
    await file.writeAsBytes(png);
  }
    return filePath;
  }
}`

elkhalifte:
You actually don't need native code for this.

just by using this [plugin](https://pub.dev/packages/image) you can achieve the results.

alfalcon90:
If anyone is looking to do this without third-party packages, this is how we got it working:

https://gist.github.com/alfalcon90/711f981780751ef81134c484ec559a75#file-crop_image-dart

PR #763, opened 2020-08-24
[flutter_local_notifications] null checking mainActivity before .getIntent()

bornold:
Calling 
```java 
 FlutterLocalNotificationsPlugin.registerWith(...
```
from background throws null pointer at
```java
launchIntent = mainActivity.getIntent();
```

Demo app repo here where we FirebaseMessaging will try to initialize  flutter_local_notifications from background (just run app):
https://github.com/bornold/firebase_messaging_local_notifications_npe_example
and example of using this fix:
https://github.com/bornold/firebase_messaging_local_notifications_npe_example/tree/using-fix


MaikuB:
Thanks!

Issue #298, opened 2019-08-23
NullPointerException in setSmallIcon


kinex:
This is probably caused by not providing correct icon name on init, but anyway it should not crash.

E/AndroidRuntime( 4210): java.lang.RuntimeException: Unable to start receiver com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
E/AndroidRuntime( 4210): 	at android.app.ActivityThread.handleReceiver(ActivityThread.java:3399)
E/AndroidRuntime( 4210): 	at android.app.ActivityThread.-wrap18(Unknown Source:0)
E/AndroidRuntime( 4210): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1780)
E/AndroidRuntime( 4210): 	at android.os.Handler.dispatchMessage(Handler.java:105)
E/AndroidRuntime( 4210): 	at android.os.Looper.loop(Looper.java:164)
E/AndroidRuntime( 4210): 	at android.app.ActivityThread.main(ActivityThread.java:6944)
E/AndroidRuntime( 4210): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime( 4210): 	at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:327)
E/AndroidRuntime( 4210): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1374)
E/AndroidRuntime( 4210): Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
E/AndroidRuntime( 4210): 	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.setSmallIcon(FlutterLocalNotificationsPlugin.java:156)
E/AndroidRuntime( 4210): 	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.createNotification(FlutterLocalNotificationsPlugin.java:131)
E/AndroidRuntime( 4210): 	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.showNotification(FlutterLocalNotificationsPlugin.java:598)
[...]

PR #768, opened 2020-08-25
[flutter_local_notifications] setSmallIcon NullPointerException

mgonzalezc:
This is a proposed fix for 'Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference' in setSmallIcon method, described here https://github.com/MaikuB/flutter_local_notifications/issues/298

My thought is that, even if the small icon is indicated when a notification is being scheduled, some devices are clearing the data stored in SharedPreferences. This means that when a scheduled notification is going to be shown, it cannot find the small icon, so it crashes.

I think using the app icon in those cases is not the most elegant solution, but it has helped our team to reduce crash rate (this bug was number 1 in our list). I hope it helps.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -183,12 +183,15 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
         } else {
             SharedPreferences sharedPreferences = context.getSharedPreferences(SHARED_PREFERENCES_KEY, Context.MODE_PRIVATE);
             String defaultIcon = sharedPreferences.getString(DEFAULT_ICON, null);
-            if (StringUtils.isNullOrEmpty(defaultIcon)) {
+            
+            if (!StringUtils.isNullOrEmpty(defaultIcon)) {
+                builder.setSmallIcon(getDrawableResourceId(context, defaultIcon));
+            } else if (notificationDetails.iconResourceId != null) {
                 // for backwards compatibility: this is for handling the old way references to the icon used to be kept but should be removed in future
                 builder.setSmallIcon(notificationDetails.iconResourceId);
-
             } else {
-                builder.setSmallIcon(getDrawableResourceId(context, defaultIcon));
+                // for errors where small icon is null
+                builder.setSmallIcon(context.getApplicationInfo().icon);
             }
         }
     }


MaikuB:
Thanks for the PR and you might be onto something here even if the PR isn't merged in. Have you been able to reproduce this by clearing the data? I'll need to see if I can reproduce this too, if so then I'll need to consider removing the ability to set a default icon or add a caveat to the docs. Note that there should already be an ability to specify an icon for each individual notification so that can be used to get around the issue if your hypothesis turns out to be true.

The other thing to note is that scheduled notifications are also saved to shared preferences so they can remain scheduled after a phone is rebooted. If your users are clearing out shared preferences then they may find that there are notificactions that don't appear

mgonzalezc:
Thank you for your comment!

No, sorry, I haven't tried to reproduce the issue myself. I just tried to guess where the problem was by looking into the code.

I will explain my experience, in case it helps:

First, we were using only the default icon specified on plugin initialization. After exploring the code, I saw that I missed the ability to specify the icon for each scheduled notification so I tried to use this instead. I also wrote a function that removed the old scheduled notifications and created the new ones with the icon for the users. I was pretty sure this would help but it didn't.  

Ie, both the default icon and the specified icon were null so the plugin executed the last else in the setSmallIcon method:

`builder.setSmallIcon(notificationDetails.iconResourceId);
`
which crashes because `notificationDetails.iconResourceId` is null. 

That's why I added a new "else" that puts the app icon instead if both default and specified icon are null. The crash has not happened anymore.

I didn't check how the scheduled notifications were stored. If they are also stored in the SharedPreferences as you mentioned then I don't understand why the plugin is still trying to show the notification but the icon is not there anymore... I will try to go through this one more time to see if I can find something

MaikuB:
Have you been able to find something or a way to reproduce the issue you've encountered?

MaikuB:
Unless you've found more details on how to reproduce this, I'll likely close this PR soon. I even tried using the `shared_preferences` plugin to clear everything after the notification was scheduled and it appeared fine. One of the issues with merging this plugin is it would hide when an icon hasn't been configured properly. If you're not aware, it's also possible to set an icon for each notification.

Another potential cause for this issue is when an app doesn't await the `initialize` call as it's asynchronous. Linter rules can be setup to pick issues like in the code so I would suggest double checking the code for your app to see if that is the case. If it is then it'd be much appreciated if you would let me know

leonardocustodio:
I think this is the same issue that I’ve just reported: #867 
I can’t reproduce it on my devices but I keep getting crashes every day. Although it happens just in a few devices.

MaikuB:
Closing to lack of updates and @leonardocustodio confirmed that awaiting the call to `initialize` fixed the issue

Issue #780, opened 2020-09-01
FlutterLocalNotificationsPlugin.java line 25 - 'boolean java.lang.Boolean.booleanValue()' on a null object reference


Fasust:
First of all, thanks a lot @MaikuB for maintaining this Package. We have been using it in production and it worked amazingly for us so far.

**Describe the bug**
I have been receiving some crash reports through _Crashlytics_ over the Weekend. Multiple users with Android devices encountered crashes caused by the `FlutterLocalNotificationsPlugin.java`. This the Stack trace _Crashlytics_ provided:

````
Fatal Exception: java.lang.RuntimeException: Unable to start receiver com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.Boolean.booleanValue()' on a null object reference
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:4238)
       at android.app.ActivityThread.access$1700(ActivityThread.java:274)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2113)
       at android.os.Handler.dispatchMessage(Handler.java:107)
       at android.os.Looper.loop(Looper.java:237)
       at android.app.ActivityThread.main(ActivityThread.java:8154)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:496)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1100)
````
Cause by: 
````
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.Boolean.booleanValue()' on a null object reference
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.createNotification(FlutterLocalNotificationsPlugin.java:25)
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.showNotification(FlutterLocalNotificationsPlugin.java:1)
       at com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver.onReceive(ScheduledNotificationReceiver.java:14)
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:4229)
       at android.app.ActivityThread.access$1700(ActivityThread.java:274)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2113)
[...]

PR #781, opened 2020-09-01
[flutter_local_notifications] Add Null-Safety to FullScreen Intent Check

Fasust:
Related Issue: #780


MaikuB:
I've tweaked the code a bit as I wasn't sure how long it might take to do for the change to be done if I were left a comment. Will merge this after all the checks have passed

Issue #782, opened 2020-09-01
Because every version of flutter_driver from sdk depends on platform 3.0.0-nullsafety.1 and flutter_local_notifications depends on platform ^2.2.1, flutter_driver from sdk is forbidden. So, because flutter_local_notifications depends on flutter_driver any from sdk, version solving failed


acoutts:
**Describe the bug**

Flutter 1.22 (dev/master channel) breaks flutter_local_notifications due to the platform package dependency.

**To Reproduce**

1. `flutter channel dev && flutter upgrade`
2. `cd example/`
3. `flutter pub get`
4. See error below:
```
Because every version of flutter_driver from sdk depends on platform 3.0.0-nullsafety.1 and flutter_local_notifications depends on platform ^2.2.1, flutter_driver from sdk is forbidden.
So, because flutter_local_notifications depends on flutter_driver any from sdk, version solving failed.
```

**Expected behavior**

The package should work with all Flutter upgrades.

**Sample code to reproduce the problem**

You can reproduce this with the example app included in the library.

PR #783, opened 2020-09-01
[flutter_local_notifications] Fix pub versioning error in Flutter 1.22

acoutts:
This fixes the issue here: https://github.com/MaikuB/flutter_local_notifications/issues/782

### DIFF
--- a/flutter_local_notifications/CHANGELOG.md
+++ b/flutter_local_notifications/CHANGELOG.md
@@ -1,3 +1,7 @@
+# [1.4.4+5]
+
+* [Android] Fix issue [782](https://github.com/MaikuB/flutter_local_notifications/issues/782) by removing unused dependency to the platform lib which fixes the pub versioning error in Flutter 1.22.
+
 # [1.4.4+4]
 
 * [Android] Fix issue [759](https://github.com/MaikuB/flutter_local_notifications/issues/759) by guarding code on getting the intent from the activity in case there isn't an activity that could cause `initialize()` and `getNotificationAppLaunchDetails()` to fail when called from the background
--- a/flutter_local_notifications/pubspec.yaml
+++ b/flutter_local_notifications/pubspec.yaml
@@ -1,12 +1,11 @@
 name: flutter_local_notifications
 description: A cross platform plugin for displaying and scheduling local notifications for Flutter applications with the ability to customise for each platform.
-version: 1.4.4+4
+version: 1.4.4+5
 homepage: https://github.com/MaikuB/flutter_local_notifications/tree/master/flutter_local_notifications
 
 dependencies:
   flutter:
     sdk: flutter
-  platform: ^2.2.1
   flutter_local_notifications_platform_interface: ^1.0.1
 
 dev_dependencies:


flutter_local_notifications/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# [1.4.4+5]
+
+* [Android] Fix issue [782](https://github.com/MaikuB/flutter_local_notifications/issues/782) by removing unused dependency to the platform lib which fixes the pub versioning error in Flutter 1.22.
```
MaikuB, line 3: Changelog entry will need tweaking as the dependency is actually used (it's imported in `flutter_local_notifications_plugin.dart`) but looks like it's now pulled as a transitive dependency. I'm assuming that it's the Flutter SDK that is pulling it in

MaikuB (changes requested):
Thanks for the PR. Looks good to me but the changelog entry will need a little tweaking

flutter_local_notifications/CHANGELOG.md
acoutts: In that case do we need to mark this a new major version as trying to install it prior to Flutter 1.22 might not work then?

MaikuB: Good question, I've only tried it with Flutter 1.2.0. I'll need to be tested on 1.12, if it still works then should be good to merge the PR in after amending the changelog. However, it a version bump is needed then I would suggest that the PR goes to the other branch of the repo that is already being used for the 1.5.0 beta that I intend to release soon. I'm waiting to test an upcoming daylight savings changes that takes on this weekend before I make it a stable release

acoutts: Ah! I think it makes sense to just make this part of 1.5.0 then. It's just a simple one line fix to remove the dependency.

What's the best way to make this part of that release?

MaikuB: I believe the best way is close this PR and make it that you do your PR based off the `time_zone_macos_support` branch. I see you've been doing PRs from your own master branch. One tip I learnt recently to create your own branches on your fork when submitting a PR. That can make it easier to keep your fork of master up to date so it exactly matches the original repo's master branch

acoutts: Thanks for that! Alright i'll do that.

acoutts: So to confirm you want me to fork off of `time_zone_macos_support` to my own github, create a branch called `fix-782` to reference my issue number, and create a PR to merge my branch into the `time_zone_macos_support` one here?

MaikuB: Only if it's necessary. If the PR works with Flutter 1.12 then it's not needed. If it doesn't then the PR would also need to bump the minimum supported version of Flutter.

You already have a fork of this repo and that should already have the `time_zone_macos_support` for you to branch off from provided you've kept your fork up to date.

MaikuB: Should I add that I would suggest testing with Flutter 1.12 first if you can. The Flutter website does have an archive of older SDKs up on there

acoutts: I just checked out stable branch and notice it's on 1.20.2 - curious why target 1.12 still if the stable is 1.20 now?

acoutts: Ah right - minimum supported version, I see.

acoutts: I tried it out on 1.12.13+hotfix.9 (same as in the min version), and it seems that platform is pulled in transitively here too, so I think it's safe to do this on the current version / sdk support.

acoutts:
Closing as I moved branches.

flutter_local_notifications/CHANGELOG.md
acoutts: I'm actually not able to run the example app from master branch on 1.12.13:
```
2020-09-02 08:25:24.000368-0400 Runner[78869:5970246] Failed to find assets path for "Frameworks/App.framework/flutter_assets"
2020-09-02 08:25:24.066855-0400 Runner[78869:5970804] [VERBOSE-2:engine.cc(127)] Engine run configuration was invalid.
2020-09-02 08:25:24.067021-0400 Runner[78869:5970804] [VERBOSE-2:shell.cc(437)] Could not launch engine with configuration.
2020-09-02 08:25:24.096888-0400 Runner[78869:5970821] flutter: Observatory listening on http://127.0.0.1:61872/Rr3T8tYt9GI=/
```

It does work fine on the latest stable 1.20 though. It appears the minimum supported SDK needs bumped regardless of the change here.

MaikuB:
Did you check Android? I believe the issue you're seeing is more to do with the Xcode project having been migrated as per https://flutter.dev/docs/development/ios-project-migration

acoutts:
Ah android looks fine, I understand. I just opened https://github.com/MaikuB/flutter_local_notifications/pull/785 to try this again.

PR #828, opened 2020-10-02
[flutter_local_notifications] Create notification helper

rrifafauzikomara:
Create `notification helper` with implementation `singleton pattern` to grouping all notification types including permission and initialize.
With this class, we separated the `UI` (main.dart) and the `logic of notification` (notification_helper.dart). And our code on main.dart is not too many lines of code.

### DIFF
--- a/flutter_local_notifications/example/lib/main.dart
+++ b/flutter_local_notifications/example/lib/main.dart
@@ -1,41 +1,17 @@
 import 'dart:async';
-import 'dart:io';
-import 'dart:typed_data';
 import 'dart:ui';
 
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/services.dart';
 import 'package:flutter_local_notifications/flutter_local_notifications.dart';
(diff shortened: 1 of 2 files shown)


MaikuB:
Sorry but I'm not taking this PR in. These are the reasons why

- I get that the `main.dart` file is long and one of the reasons is that doing so allows all of the Dart code to be shown in the example tab.  Your PR would prevent this from happening
- The attempt at implementing a singleton offers very little benefit when you rely on an instance of the plugin to passed through each method. 
- The point about how UI is separated isn't true as there's still UI code in the helper class.
- Helper/utility classes typically have common code his PR has really only cut and pasted the code into another class
- Your helper class isn't not self-contained and imports the `main.dart` file even
- Probably the most important reason though, is that your restructuring adds more indirection. The example is structured as it is right now that when developers look at the button for the various kinds of notification and see one of interest, they can immediately jump to the code that has nearly all the code they need to show the particular kind of notification. This makes it easier to cut and paste. Note that I say "nearly" here as things like initialisation is done elsewhere

PR #850, opened 2020-10-12
[flutter_local_notifications] Support notification reschedule for android during restart / quickboot

nghenglim:
So I am testing the reschedule functionality on my Huawei device, the local notification is not being rescheduled when I restart my device.

The following enhancement (plus making sure the startup manager don't disable launch on startup) fixes the issue.


MaikuB:
Thanks!

Issue #492, opened 2020-02-22
Clearing badge number?


jaykovski:
Is there a better/proper way of clearing badge number?
At the moment I am firing silent, empty notification (which user can't see) with a `badgeNumber: 0` to remove badge number, any better way?

MaikuB:
Try https://pub.dev/packages/flutter_app_badger

edwin-alvarez:
should this be part of the same plugin?

MaikuB:
That kind of functionality uses a different set of APIs that has no relation to notifications so won't be included in this plugin.

greymag:
Hello. I'm understand why you are not want to include this functionality in the plugin, but I would ask to consider it again. 

Here is why:
- Plugin provider an ability to set counter when we schedule a notification, but without resetting it - it's useless. 
- You suggest to use `flutter_app_badger` - and it is a solution, but it's not good one. I need one more dependency, which may not be supported in the future and which adds much more than needed (what affects the size of the application).
- Implementation of the feature - is one line and not affect any other parts of the plugin. So it does no harm those who don't need a feature, but it's makes life way more easier for those who needs it.

I'd make little PR #876 with adding this feature, hope you accept it. 

Regards!

muarachmann:
@MaikuB I get your point but I think it will be harmless and more or less, a plus to this plugin. bwt the #876 works cool

SalahAdDin:
@MaikuB could you provider an example about how to do this?

We can't found a event listener to update it when the notifications are comming.

PR #876, opened 2020-10-28
[flutter_local_notifications] add ability to set app icon badge number on iOS

greymag:
## Problem

Plugin provider an ability to set counter when we schedule a notification on iOS, but after that we can't change (or reset it) without schedule another notification.

## Solution

Added method `setBadgeNumber` In the platform implementation `IOSFlutterLocalNotificationsPlugin`

See #492 for more information.

### DIFF
--- a/flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
+++ b/flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
@@ -14,6 +14,7 @@
 }
 
 NSString *const INITIALIZE_METHOD = @"initialize";
+NSString *const SET_BADGE_NUMBER_METHOD = @"setBadgeNumber";
 NSString *const SHOW_METHOD = @"show";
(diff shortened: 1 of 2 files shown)


MaikuB:
At this stage my stance on this remains the same and won't be looking at taking PRs for this.  Given Flutter is a relatively new community, this is the opportunity for the community to promote existing work to avoid what's happens in other communities where there are many packages that overlap with each other or solve the same problem. This is similar to how the rework for `firebase_messaging` has documentation for using `flutter_local_notifications` to display a local notification upon receiving a push notification. 

I understand this adds an extra dependency but when I compare size it adds to the release binary, it adds a such a miniscule amount that it is neglible. If such an extremely small difference is a concern for one's context that it is they likely should be looking at implementing everything on their own.

Given that what `flutter_app_badger` does, I doubt it needs to be maintained. Should there be a case where it no longer works then I'd be happy to take a look at this again

SalahAdDin:
@MaikuB could you provider an example about how to do this?

We can't found a event listener to update it when the notifications are comming.

PR #880, opened 2020-10-29
WIP: [flutter_local_notifications ] Notification actions for iOS, macOS, Android

ened:
This PR addresses #17 - and includes a very basic implementation for iOS & Android.

A click on the notification will launch (if not already running) a separate Flutter engine, which initializes a callback dispatcher. 
The dispatcher function then sets up platform communication for other plugins and starts listening to a event channel for existing or additional notification click events.

### Screenshots

**Android**:

Simple actions:

screen

Text input actions to collect user entered text:

screen

**iOS**:

Simulator Screen Shot - iPod touch (7th generation) - 2020-11-01 at 09 23 40

**macOS**:

Simple actions: 

Screen Shot 2021-12-29 at 20 50 59

Text input actions to collect user entered text:

Screen Shot 2022-01-02 at 22 28 54

The PR is in early stage to gather feedback on the basic use case by others.

### Feedback requested

Please support this PR by testing it in your own App and comment with missing features / use cases.

To use it, add the following to your `pubspec.yaml` file:

``` yaml
dependency_overrides:
  flutter_local_notifications:
    git:
      url: https://github.com/ened/flutter_local_notifications.git
      path: flutter_local_notifications
      ref: 17-notification-actions
  flutter_local_notifications_platform_interface:
    git:
      url: https://github.com/ened/flutter_local_notifications.git
      path: flutter_local_notifications_platform_interface
      ref: 17-notification-actions
```

The documentation is still very incomplete, therefore please refer to the example project of this branch.

### Open tasks

- [x] Text input & predefined action (Android) support
- [x] Android: Notification Actions text color: https://stackoverflow.com/questions/41073294/how-to-change-notification-action-text-color-in-android-n
- [ ] Documentation
- [ ] Test coverage
- [x] macOS support
- [x] metadata/payload accessible in action handler
- [x] autoCancel property for android notification action handler that dismisses the notification (feature parity with iOS)


ened:
@MaikuB what code style does this project follow? Any particular setting I could enable/import in Android Studio?

ghenry:
Awesome!

MaikuB:
@ened I've done very little of pure native Android development so don't know how style settings could be configured so admittedly there isn't a particular style used at the moment. Only thing I try to do is avoid hungarian notification, naming constants with the `k` prefix and try to use early returns where possible.

FYI that there were other PRs before. The most comprehensive one was probably https://github.com/MaikuB/flutter_local_notifications/pull/338.

There were also issues with background execution that I mentioned in #17 and reported on the Flutter repo. From memory, I saw discussions on the main Flutter repo that looking at background execution was on the horizon.

Note: I say pure native Android development as my previous experience doing Android development was using Xamarin

ened:
@MaikuB ok - thx for the link to the other PR & I will extract what is needed. The background execution support is included in Flutter, see flutter_workmanager or flutter_uploader plugins as a reference. I think we can get this sorted, but it will be a bit foreign to some developers, because the callback function runs in a separate thread. On the other hand, if you are using firebase_messaging and the `onBackgroundMessage` handler there, you'll feel right at home.

For the code formatting, I normally enable the "google-java-formatter" plugin in Android studio and let it do it's magic on ALL java files. That's what the Flutter team is using as well.

I can not include formatting changes in this PR though, so would you be willing to tackle this or should I submit a separate PR to reformat _all_ of the Android java code using that plugin? Obviously it would include some notes in the README.

MaikuB:
@ened I know it's supported but I ran what seemed like rather severe issue when I tried to implement it in this plugin and within the code of one of the Flutter team member's own samples that I reported in https://github.com/flutter/flutter/issues/23904. Note that I accidentally deleted the fork of the geofencing sample but it was acknowledged as an issue

With regards to formatting, happy to take guidance on this and a PR on this. I have had to format Java code for PRs I've submitted to the Flutter and Firebase plugins that I would assume use the same formatter but used the script they had created for it as I wasn't aware that there's an Android Studio plugin for it

ghenry:
> I think we can get this sorted, but it will be a bit foreign to some developers, because the callback function runs in a separate thread. On the other hand, if you are using firebase_messaging and the `onBackgroundMessage` handler there, you'll feel right at home.
> 

Sorry to ask something unrelated to this thread, but the above caught my eye.

I'm stuck on background messaging and thought it is exactly something to do with threading and my .Application class and .MainActivity class. Do you have a link to something I can read more about on what you've just highlighted?

Thanks,
Gavin.

ened:
The PR has been now been updated with a iOS implementation - from basic tests in emulator it works the same way as Android, but does require a a bit more preparation work by the user of this plugin:

1. `AppDelegate` needs to register the plugin registrant callback 
2. iOS defines actions per notification category, so the developer needs to configure them during the `initialize` call. Internally, the categories need to be configured before the permissions are requested; which is reflected in the code.

I've updated the example project with this - but documentation is still a bit sparse; this is probably the biggest task now.

It would be great if the community could support testing this and submit feedback. Will update the PR description with details.

Also, we should define how much of the APIs available is to be implemented:

https://developer.android.com/training/notify-user/build-notification#Actions
https://developer.apple.com/documentation/usernotifications/declaring_your_actionable_notification_types

@MaikuB For the code formatting, unfortunately https://github.com/google/google-java-format/issues/536 is blocking a proper solution. Will follow this up when there are news.

@ghenry I'm not sure what the problem is exactly. The example project in this fork uses the same mechanism, so I'd encourage you to test it locally. It's also a lot smaller than the firebase plugin, so should be easier to navigate around. :)

ghenry:
>
> @ghenry <https://github.com/ghenry> I'm not sure what the problem is
> exactly. The example project in this fork uses the same mechanism, so I'd
> encourage you to test it locally. It's also a lot smaller than the firebase
> plugin, so should be easier to navigate around. :)
>

Thanks. After reading all the firebase code last night and callkeep code, I
think my issue is something being run in one isolate (background message
handler) and callkeep running in another isolate/activity with no access to
each other and callbacks in callkeep acting on empty states as the states
are in the other thread. Still researching. I may just bring the lot to the
foreground :)

Looking forward to playing with this code with GetX and xmpp_stone

Great work!

Gavin.

>

ened:
> Thanks. After reading all the firebase code last night and callkeep code, I think my issue is something being run in one isolate (background message handler) and callkeep running in another isolate/activity with no access to each other and callbacks in callkeep acting on empty states as the states are in the other thread. Still researching. I may just bring the lot to the foreground :) Looking forward to playing with this code with GetX and xmpp_stone Great work! Gavin.

Transporting messages across isolates could be done in two ways:

1. You could try registering a named [ReceivePort](https://api.flutter.dev/flutter/dart-isolate/SendPort-class.html) using [IsolateNameServer](https://api.flutter.dev/flutter/dart-ui/IsolateNameServer-class.html).
2. You could use a plugin with a static field to keep/publish state messages between the isolates. Having it done this way would allow you to customize cache behavior when one of the isolates disappeared. This will require both isolates to have access to plugins though.

ghenry:
>
> Thanks. After reading all the firebase code last night and callkeep code,
> I think my issue is something being run in one isolate (background message
> handler) and callkeep running in another isolate/activity with no access to
> each other and callbacks in callkeep acting on empty states as the states
> are in the other thread. Still researching. I may just bring the lot to the
> foreground :) Looking forward to playing with this code with GetX and
> xmpp_stone Great work! Gavin.
>
> Transporting messages across isolates could be done in two ways:
>
>    1. You could try registering a named ReceivePort
>    <https://api.flutter.dev/flutter/dart-isolate/SendPort-class.html>
>    using IsolateNameServer
>    <https://api.flutter.dev/flutter/dart-ui/IsolateNameServer-class.html>.
>    2. You could use a plugin with a static field to keep/publish state
>    messages between the isolates. Having it done this way would allow you to
>    customize cache behavior when one of the isolates disappeared. This will
>    require both isolates to have access to plugins though.
>
>
Thanks Sebastian. I think I'll initialise some things in the background
message handler to test. That's enough of me polluting this issue, but
thanks for reading and replying.

Gavin.

Barry0501:
Woa, awesome! This is exactly what I need in this plugin

yaakov-lightricks:
Hi, this is great! Exactly what I miss in this great plugin for long time!
If help is needed I can try help.

postacik:
Glad to see this coming finally. I'll test and report...

postacik:
I need help for using this feature.

The expected behavior of clicking an action button is similar to clicking the notification body except for an additional parameter which holds the clicked button id.

However, in your solution, a function in a separate isolate is called and even the notification is not dismissed when the action button is clicked. (I tested it on Android)

Is it not possible to trigger the `onSelectNotification` event when an action button is clicked?

Or is it possible to wake the application up in the `backgroundHandler` ?

ened:
Following answer is for the current implementation:

> However, in your solution, a function in a separate isolate is called and even the notification is not dismissed when the action button is clicked. (I tested it on Android)

It's a good question whether should the notification action be dismissed automatically. I tend to agree it should be. Will look at implementing the code.

> Is it not possible to trigger the onSelectNotification event when an action button is clicked?

Precisely because it's in a separate isolate, the `onSelectNotification` event may not be called. 
Consider that the actions can be simple buttons like in the demo, or in the future perhaps input / message fields (like for message apps). When the notification arrives, the App may be terminated & the only part of the App handling the notification action is the background handler.

The handler has full app functionality available (plugins, etc), just not the App itself.

This is a design decision in the Android system itself, as evidenced by how notification actions are handled (in a "BroadcastReceiver" which has a separate lifecycle).

> Or is it possible to wake the application up in the backgroundHandler ?

Based on above, you could use plugins like `android_intent` to start the main activity back.

There may be other issues by starting the activity this way, e.g. it will probably need to get a flag called "start new task", which means the App starts from 0 and you'd see the notification in a different place then `onSelectNotification`.

enesoren:
The feature I've been waiting for so long, Is there any prediction about when it will be merged?

ened:
@enesoren it needs more testing.. I can resolve issues but am currently very time-constrained rgd. testing.

MadhaviMadineni:
@ened Hi, this is great. any predictions, as to when will this get merged.

ened:
@MadhaviMadineni the PR needs community testing, please see the instructions on the top. 

While at it, the plugin should probably also support arbitrary notification actions like text input (so that you can build a chat message app); will look at that next.

enesoren:
Hi, I need help to test it. I get the following error when I make the change in pubspec:

Could not find a file named "pubspec.yaml" in https://github.com/ened/flutter_local_notifications.git 6954ace17b999094d0b7e1e78f974702862bf478.
Running "flutter pub get" in app...
pub get failed (1; Could not find a file named "pubspec.yaml" in https://github.com/ened/flutter_local_notifications.git
6954ace17b999094d0b7e1e78f974702862bf478.)

How can I solve this problem?

ened:
@enesoren ah - I have written the wrong instructions on the top, please use this instead:

```
dependency_overrides:
  flutter_local_notifications:
    git:
      url: https://github.com/ened/flutter_local_notifications.git
      path: flutter_local_notifications
      ref: 17-notification-actions
  flutter_local_notifications_platform_interface:
    git:
      url: https://github.com/ened/flutter_local_notifications.git
      path: flutter_local_notifications_platform_interface
      ref: 17-notification-actions
```

enesoren:
@ened Yep, it works. I can test it now, thank you :)

Barry0501:
Hi @ened, your PR is awesome. I have been waiting for this feature for a long time. How do you deal with the problem when user dismiss Notification

ened:
> Hi @ened, your PR is awesome. I have been waiting for this feature for a long time. How do you deal with the problem when user dismiss Notification

What is the problem with that? When a user dismisses the notification there is nothing an App can do, really. :)

hardik-weenggs:
Can you add Notification Data(Map<String, dynamic> message) with id on **backgroundHandler**  callback?

ened:
@hardik-weenggs Good idea. Yes, will add that next.

hardik-weenggs:
And also add color change feature to ActionButton text.

joankabello:
> > Hi @ened, your PR is awesome. I have been waiting for this feature for a long time. How do you deal with the problem when user dismiss Notification
> 
> What is the problem with that? When a user dismisses the notification there is nothing an App can do, really. :)

Hello @ened,

I think what @Barry0501 meant is that when you want to show a foreground notification you should pass the parameter `ongoing: true`. But if the notification is ongoing you can not slide to dismiss. Is there any way to make this type of notification dismiss-able ? (At least this is a problem that i occurred, which i think is not related much with the feature but i would be good if there was an option)

mehrdad-shokri:
@MaikuB @ened any estimated on when this pr will be merged?

MaikuB:
@mehrdad-shokri as @ened had mentioned already, this needs more community testing and feedback around use cases etc. Currently, there's been very little feedback/discussion from the community and a macOS implementation would also be needed. If you the way the PR works now covers your scenarios, you can still point to the fork as you don't need to wait for this to be merged to use it

marcqtan:
> How do I get notification data if I click on the Action button directly?
> Can you add Notification Data(Map<String, dynamic> message) with id on **backgroundHandler** callback?

Any updates on this? I need this feature too. Thanks!

marcqtan:
I'm trying to open a URL when clicking the first Action item. using url_launcher. but I'm having this error below:
Screen Shot 2020-12-21 at 2 06 18 PM



opening URL via onSelectNotification() works fine.

```
await flutterLocalNotificationsPlugin.initialize(initializationSettings,
        backgroundHandler: notificationActionCallback,
        onSelectNotification: (String payload) async {
      if (payload != null) {
        print('notification payload: $payload');
        if (await canLaunch('https://www.google.com')) {
          await launch('https://www.google.com');
        } else {
          throw 'Could not launch';
        }
      }
});
```


```
static void notificationActionCallback(String id) async {
    if (await canLaunch(id)) {
      await launch(id);
    } else {
      throw 'Could not launch $id';
    }
 }
```


Any Idea on how to solve this?

ened:
@marcqtan the only android `Context` available to url_launcher when running in the isolate is the Application context itself, therefore calling launch() will not work. I suggest you try using android_intent and search for the "new activity" flag. Remember that notifications actions can be processed at any time even when the application is not running.

@hardik-weenggs Support for text input has been added.

sante85:
please merge, because is a very cool feature!

enesoren:
Hi, I'm trying to post to a URL with message-id when I click to action button. Is there any way to get the message-id or notification information when I click the action button?

ened:
> Hi, I'm trying to post to a URL with message-id when I click to action button. Is there any way to get the message-id or notification information when I click the action button?

You try to attach metadata / extra information to the action itself?

enesoren:
> > Hi, I'm trying to post to a URL with message-id when I click to action button. Is there any way to get the message-id or notification information when I click the action button?
> 
> You try to attach metadata / extra information to the action itself?

I try to access the notification payload in the notificationTapBackground function.

ened:
> > You try to attach metadata / extra information to the action itself?
> 
> I try to access the notification payload in the notificationTapBackground function.

From Android point of view, the notification itself is not available, but there is a way to add "extras" to the notification action. I'll take a look how that works. You may need to specify the payload on each action.

ened:
@enesoren this has been added, please take a look at the latest version :)

mk-dev-1:
Hi guys,

I noticed that the payload I am getting remains the same, even if I schedule a new notification from within the app (I have confirmed that I am setting a new payload). So it looks like something is not updating right, or do I have to reset/clean something upon finishing my background actions?

ened:
@mk-dev-1 I noticed something similar but didn't get into much detail yet. As a test, try reinstalling the App and schedule another notification - this time press the other action button (or the other notification). Did the payload change?

firatcetiner:
This really is a good feature. @ened Thanks for your contribution. I am going to test this on my Android device and hopefully on an iOS device too. I hope more people can test this feature and you will be able to merge it asap.

ReniDelonzek:
Hello, I'm testing the feature here and it works with charm. I would like to know how I can make the notification go away after an action is triggered?

ened:
@ReniDelonzek can you try to use the standard `cancelNotification` API in the plugin?

ReniDelonzek:
@ened @firatcetiner 
How can I get the notification id to cancel it? Since the callback of the action sends only the id of the action itself, not the notification

firatcetiner:
@ReniDelonzek You should store the id of your notification in somewhere. For example I store them as static int fields in my NotificationUtil class.

ReniDelonzek:
@firatcetiner Even if I keep their ids in a static field, how can I know which one the action was selected in?

firatcetiner:
@ReniDelonzek Well you can add the id to the payload. @ened Maybe the action callback should expose the notification's id?

ReniDelonzek:
I believe that exposing the notification id would be the most correct way

ReniDelonzek:
Hi, I did some tests adding the id to the payload as suggested by @firatcetiner, but the notification remains there even after calling ```flutterLocalNotificationsPlugin.cancel(id)```
Tested on android

hawkinsjb1:
I'm attempting this solution for an app which requires me to click "Stop Timer" from the navigation tray. 

If I'm understanding correctly, it seems as though I can use an intent to "restart" my app, with specific params

```
void notificationTapBackground(String x, String y, String z) {
  if (Platform.isAndroid) {
    AndroidIntent intent = AndroidIntent(
      action: 'action_edit',
      data: null,
      arguments: {'action': 'stop_timer'},
    );
    intent.launch();
  }
}
```
My hope is that I can use the above code to silently restart my app service, check for any intent-based launch params, and call special functions if params are present (e.g. 'stop_timer' => stopTimer API call @ app start)

Am I correct in assuming this can all be done silently and in the background, even though the app is technically being started up?

Finally, if the above implementation is correct.. now all I should need is figuring out how exactly to use MainActivity.kt to register incoming intents, I've seen lots of examples for MainActivity.java but I'm not an Android dev so I'm not really sure how it translates to Kotlin. Any insights would be appreciated

ReniDelonzek:
Just completing the reports of my tests, in IOS, when clicking on the notification to open the app (without any action at all), the app is not opened and a crash is generated.

Tested on real devices, on various versions of IOS

Miller9921:
HOLA, Alguien me podria dar informacion sobre este error al intentar agregar las librerias.
ERROR:
[localnotifications] flutter pub get
Running "flutter pub get" in localnotifications...              
Because localnotifications depends on flutter_local_notifications from git which depends on flutter_local_notifications_platform_interface ^2.0.0+1, flutter_local_notifications_platform_interface from hosted is required.

So, because localnotifications depends on flutter_local_notifications_platform_interface from git, version solving failed.
pub get failed (1; So, because localnotifications depends on flutter_local_notifications_platform_interface from git, version solving failed.)
exit code 1

Miller9921:
Hola, como puedo realizar una accion desde los botones de notificacion, ya estan implementados pero aun no se como puedo ponerlos a funcionar para que realicen algo.

lokeshofficial:
how can I dismiss the notification once I clicked on any action ?

lokeshofficial:
HI @ened ,

I'm glad you implemented the actions. I have been waiting for this feature so long.

Although I noticed a bug where you cannot provide payload for an action. 
Every time the payload is 0 whenever a action is triggered.

Is there anything I'm missing apart from example I got from your repo.

Thanks

paul678:
Hi @ened annd @MaikuB any updates on the status of this PR? I have an app (Android + iOS) undergoing development so I may be able to help with testing if needed (I need to support 2 action buttons on zonedScheduled notifications)
Many thanks!

Viraakkp:
This is really cool!!!!!

DesmondFox:
very useful feature!

jonas-zebari:
On iOS, will it be possible to show different actions based on the data payload contents from a Firebase Messaging notification?

ali321-hub:
Hello, we can't try this feature because it is not null-safety. Do you have any guesses about when it will be updated?

FatihTirek:
Hi @ened, I'm looking forward to use this feature but there is a problem with payload. It stays same even though I changed their content. And I'm sure that this issue didn't cause from my code because I saw payload ids in the console. Their content were exactly same. I hope this issue will be fixed as soon as possible because I have an app in production and some of users requested this feature. I tried to fix that but I failed cause I don't have enough Java experiences.

ened:
> Hi @ened, I'm looking forward to use this feature but there is a problem with payload. It stays same even though I changed their content. And I'm sure that this issue didn't cause from my code because I saw payload ids in the console. Their content were exactly same. I hope this issue will be fixed as soon as possible because I have an app in production and some of users requested this feature. I tried to fix that but I failed cause I don't have enough Java experiences.
> 
> 
> 
> Edit: Hi again @ened, I found the issue. It seems wrong intent comes to the ActionBroadcastReceiver. I guess the actionPendingIntent's request code overlapping another request code. Anyway I just changed request code and flag in FlutterLocalNotificationPlugin.java at line 199 like this: 
> 
> 
> 
> PendingIntent actionPendingIntent = PendingIntent.getBroadcast(context, requestCode++, actionIntent, PendingIntent.FLAG_UPDATE_CURRENT);
> 
> 
> 
> and line 188: 
> 
> 
> 
> int requestCode = 915612566; ( Random number )
> 
> 
> 
> It works like a charm now.

Thank you so much.

I'll reactivate the PR ASAP!

ali321-hub:
Hi, thanks for the plugin, it's a very useful feature.
 When you change the id parameter in Android notificationActions, the incoming id does not change.

 Is it possible to make changes on the notification without closing the current notification? ike the feature of adding a new message on top of other messages when a new message is received in Whatsapp.

Finally, can you make a function assignable to the place that says chrome like in this picture?
https://docs.google.com/document/d/1epoSlZZvWIRt9Xxhfqh_aoyO_xENSTnHnBFdHoAqB3M/edit?usp=sharing

jonas-zebari:
This looks super promising :) Is there any estimate as to when this might make its way to release/pre-release?

Bringoff:
This PR was opened literally a year ago. Any plans on finishing and merging it? 🙂

abinemmanuel:
An awesome package [flutter_local_notifications](https://pub.dev/packages/flutter_local_notifications) and an excellent [enhancement](https://github.com/MaikuB/flutter_local_notifications/pull/880) for that!
I am already using  [flutter_local_notifications](https://pub.dev/packages/flutter_local_notifications) for local and FCM notifications in my app and is looking for this exact solution for supporting _Notification Actions_  for some use cases.
 Would be happy to take part in testing this enhancement and making it part of this package!

@ened , since it is being fine-tuned for over an year now, how far it is from getting `pre-release` / `release` state?
Really looking forward to it. Thanks!

Kavantix:
I just made a PR to @ened 's branch that merges in the current master and migrates to nullsafety.
The actions seem to work on iOS and Android

ened:
@Kavantix You're a superstar! :-) Thank you.

MaikuB:
@ened let me know if you need assistance with something besides reviewing the code. I'll give it another spin to test since it's up to date now

ened:
> @ened let me know if you need assistance with something besides reviewing the code. I'll give it another spin to test since it's up to date now

I think we need to confirm the documentation required for library users to have a successful integration.

And of course the APIs should function as you'd expect.

.github/workflows/format.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           find . -name '*.m' -exec clang-format -i {} \;
           find . -path '*/ios/**/*.h' -exec clang-format -i {} \;
           find . -path '*/macos/**/*.h' -exec clang-format -i {} \;
-          git diff --exit-code
+          git diff --exit-code || (git commit --all -m "Clang Format" && git push)
```
MaikuB, line 25: Nice one :)

This and the Swift one look useful enough on their that i'll try to get this on the `master` branch separately

alexursul:
Hello! I've noticed a bug. It seems that an `id` from `NotificationActionDetails` returns an id of the last shown notification. Also, when 2 or more notifications remain in tray, pressing on an action button make this action button "pressed" on another notifications (impossible to press and fire an action more than once).

Kavantix:
@alexursul That's on android right?

alexursul:
@Kavantix Yes. Haven't tested on iOS.

lucasribolli:
Is there a way to prevent the Android device to kill the isolate when the user taps the action button? I mean, would it be a foreground service that requires a notification to give a feedback to the user? Or the Flutter Engine handle this by itself?

I am talking about the **Required notifications** section of this [link](https://material.io/design/platform-guidance/android-notifications.html#usage)

flutter_local_notifications/example/lib/main.dart
```diff
@@ -806,7 +880,126 @@ class _HomePageState extends State<HomePage> {
     const NotificationDetails platformChannelSpecifics =
         NotificationDetails(android: androidPlatformChannelSpecifics);
     await flutterLocalNotificationsPlugin.show(
-        0, 'plain title', 'plain body', platformChannelSpecifics,
+        id++, 'plain title', 'plain body', platformChannelSpecifics,
```
MaikuB, line 883: @Kavantix @ened nice idea to increment but one thing to note on why  I had originally set the id to zero in all cases was to allow for `cancel()` to dismiss the notification on the screen. It looks like the `_cancelNotification()` method still invokes logic to cancel the notification with an id of zero. This may potentially look like it's not cancelling anything if multiple notification requests are made. If changes are going to be made around the id is set then a way to cancel a notification with a specific id is required and allows users of the example to know exactly what is being cancelled/deleted

Kavantix: Good point, we can make cancel also increment so they are also cancelled in order?

Kavantix: Or we could even show a list of notifications that were scheduled and allow them to cancel a specific one

ened: Let's think who should keep track of notification IDs. 

I think it's the client that's responsible, so I would prefer to store the notification ids in a list, then let the "cancel" function in the example iterate through the list and cancel them one by one, then clear the list.

In a real world App you'd have specific notification ids as constants.

ened: > Or we could even show a list of notifications that were scheduled and allow them to cancel a specific one

Yes, essentially that based on what I just posted, if we want things to be that complex in the example app.

Kavantix: No in a real world app they are not per se constants just known values

ened: > No in a real world app they are not per se constants just known values

Ok agree. ;-)

MaikuB: This one's tricky as there could be arguments for decrementing instead so it cancels the last notification that was shown. A really naive approach I can think of is to continue incrementing and  have more changes done so the text within the notification will have the id. With that done, a text field can be added to users specify the ID to cancel

Kavantix: But yes keeping a list of ids would be simple enough, and then the button can say `cancel all notifications`

Kavantix: Ah last notification first, interesting that does make sense. We could use a stack perhaps

ened: `I++`,
`I--`

Maybe?

Kavantix: @MaikuB with that idea, you don't know the id of a notification that is scheduled in the future right, that only works for already delivered notifications.
Adding the list should be doable IMO (it is flutter after all)

Kavantix: Or we could just do cancel all and actually call the cancel all method of the package?
That's by far the simplest right

MaikuB: yeah the  i++/i-- approach could work

I was going to say if a list was going to be used then note that the app already has a button that calls https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/FlutterLocalNotificationsPlugin/cancelAll.html and could result in some confusion in the button text isn't clear enough. That is, instead of cancel all, perhaps it's more accurate to say something along the lines of "cancel each (individual) notification by id". 

Note: in the original code I had for the example app, given every notification had the same ID, cancel all led to same result. Its purpose was more to highlight that there's a different API in case you're wondering why I bothered putting that there :)

MaikuB:
Am I missing something in trying to get the actions appear on iOS? Was running the example app and only see notifications without actions. This was on iPhone 13 simulator running iOS 15.2

flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
```diff
@@ -234,6 +253,100 @@ - (void)pendingNotificationRequests:(FlutterResult _Nonnull)result {
...
+  return result;
+}
+
+/// Extracts notification categories from [arguments] and configures them as
+/// appropriate.
+///
+/// This code will simply return the `completionHandler` if not running on a
+/// compatible iOS version or when no categories were specified in [arguments].
+- (void)configureNotificationCategories:(NSDictionary *_Nonnull)arguments
+                                 result:(FlutterResult _Nonnull)result {
```
MaikuB, line 284: Was there a reason to include the `result` object here? From what I could tell, it shouldn't be needed here and using here could lead to issues as calling `result(...)` signals to Flutter that the `Future` related to `initialize()` method has completed. This is actually done as part of the `requestPermissionsImpl()` method in Objective-C

cc @Kavantix in case this was done of a PR you did to update @ened's fork and may have an answer

Kavantix: Now that you mention it I did also wonder the same thing when I was updating the PR.
This should probably be refactored such that neither method calls the result.
It should just be called in initialize

MaikuB: The reason it's called in `requestPermissionsImpl()` is because it is meant to return the result of answering the permissions prompt on iOS/macOS

ened: This was adjusted in the latest MR. Please re-open a comment when needed.

ened:
> Am I missing something in trying to get the actions appear on iOS? Was running the example app and only see notifications without actions. This was on iPhone 13 simulator running iOS 15.2

The actions will only appear once you long-press the notification.

flutter_local_notifications/example/lib/main.dart
ened: This was addressed in the latest update, please take a look.

flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
MaikuB: @ened to clarify, did you mean there was a comment that I've missed I could've been reopened and replied to or to re-open this one if there's something to follow-up on?

ened: The latter one, please reopen if further changes are needed. :-)

MaikuB:
Thanks @ened. Been a while since I used iOS. Ran into a weird problem though, the actions were appearing for me before and from the notification centre, I could swipe and select "View" to see the actions. However, after some point, I no longer see the actions. See below to see what I mean

https://user-images.githubusercontent.com/25263378/147644439-07e4b2d4-aa88-41df-8299-4e68ee011402.mov

Any idea what be the cause of the issue?

A few thoughts upon looking at the PR and seeing what's there. Let me know what you think

- Some applications will want actions to trigger the app to navigate to a specific page/route when
    - the app is running. This requires communication between isolates. Be good add an example app this via [IsolateNameServer](https://api.flutter.dev/flutter/dart-ui/IsolateNameServer-class.html). This is something I could help with it and this try out from your PR. In doing so however, this led me to see an issue where calling `initialize()` would need to be done in a `StatefulWidget` (e.g. as part of `initState`) e.g. to get a `BuildContext` to use the imperative Navigator APIs. This makes me think it would be better to change the plugin to use streams instead of callbacks like the `firebase_messaging` plugin. Any thoughts on this?
    - when the action launches the app when it's in a terminated state. From what I could see, it doesn't seem like it's possible. On iOS, I believe it should've been possible via [UNNotificationActionOptionForeground](https://developer.apple.com/documentation/usernotifications/unnotificationactionoptions/unnotificationactionoptionforeground?changes=__7&language=objc). I suspect this isn't working as expected as if you check the numeric value in the documentation page I linked, it's `1 << 2` and that ends up with `4`. From what I could tell, the mapping logic within the PR takes the numeric `index` value of the enum and adds one to it so it ends with a value of 3 instead. I suspect `IOSNotificationCategoryOption` may have a similar issue given bitwise operations are involved as opposed to being a straightforward integer value. Putting that issue aside, is there a way to determine which action (id) launched the app? This would possible involve changes to the logic related to [getNotificationAppLaunchDetails()](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/FlutterLocalNotificationsPlugin/getNotificationAppLaunchDetails.html)
- Would it be possible to show how an action could trigger another plugin like the `url_launcher` one mentioned in the updates in the readme? I don't know what would actually be involved myself. This could be done by updating the example app
- On iOS, tapping on an action dismisses a notification by default. Although it's not the default behaviour on Android, I believe most apps would also want this behaviour on Android. Is there a way to support this? It looks like the example app tried to demonstrate this by calling the `cancel()` method of the plugin . This could cause an issue for applications dealing with scheduled/recurring notifications that have actions as calling `cancel()` would also unschedule it so to speak. Perhaps this calls for having a boolean `autoCancel` property on the `AndroidNotificationAction` that defaults to true?
- Was the plan to add macOS support to this PR too? Asking as the PR introduces a number of classes/enum with the `IOS` prefix. iOS and macOS share the same notification APIs though so following this convention could lead to some duplication where there'd be macOS classes and enums. Perhaps replacing `IOS` with `Apple` would be better in cases like this? I may need to see if this can be done throughout other places in the existing code of the plugin first

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -211,6 +220,58 @@ protected static Notification createNotification(
             .setOngoing(BooleanUtils.getValue(notificationDetails.ongoing))
             .setOnlyAlertOnce(BooleanUtils.getValue(notificationDetails.onlyAlertOnce));
 
+    if (notificationDetails.actions != null) {
+      // Space out request codes by 16 so even with 16 actions they won't clash
+      int requestCode = notificationDetails.id * 16;
```
MaikuB, line 225: Is multiplying by 16 required? From what I could tell from the issue that had been reported before, the problem was to do with using the same request code (999) for all of the pending intents associated with an action. From what I read [here](https://stackoverflow.com/a/38759473) that also references the official docs, so long as the combination of receiver and request code are unique then there shouldn't be clashes. If users of the plugin provide unique action IDs, this should mean that multiplying by 16 is redundant 

cc @Kavantix

Kavantix: Ah I see what you mean, I'm not 100% sure I tried to use the notification id directly, if that would also solve the problem then the times 16 should indeed not be needed

MaikuB: Oops apologies I misread and thought it was a numeric action id being passed. In that case a solution that I reckon would be better is to remove this logic and then update the line further below so instead of

```
 PendingIntent.getBroadcast(
                context, requestCode++, actionIntent, PendingIntent.FLAG_ONE_SHOT)
```

it becomes

```
```
 PendingIntent.getBroadcast(
                context, action.id.hashCode(), actionIntent, PendingIntent.FLAG_ONE_SHOT)
```

to allow for using the hash code of the action id, which should be unique given action id is expected to be a unique identifier. Thoughts?

Comment (Kavantix) [reply flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java:225]: Hashcodes are not guaranteed to be unique, even worse, this will make the request code of two different notifications the same since the action id will be the same and thus the hashcode

Comment (MaikuB) [reply flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java:225]: Ah good point, been a while since I looked at them so was going by how a lot of equality checks I've seen of late includes it as part of the calculation. Regarding the second point, appending the notification id to the string could help but it's a moot point if the hash codes aren't unique

**Comment by MaikuB** [comment]
@ened when you got a moment could you let me know what your thoughts about what I wrote in my previous message? Apologies for the long wall of text there.

Another thing I tried to create a branch based on the latest changes to update the example app. This was to show how to use the `IsolateNameServer` so that one of the actions will trigger the app to navigate to a different page when it's in the foreground. The app failed to compile on iOS though as it couldn't find "ActionEventSink.h". I assume this might happened when you tried to update to refactor classes common to iOS and macOS but couldn't actually do so as macOS is using Swift?

**Comment by ened** [comment]
> Another thing I tried to create a branch based on the latest changes to update the example app. This was to show how to use the `IsolateNameServer` so that one of the actions will trigger the app to navigate to a different page when it's in the foreground. The app failed to compile on iOS though as it couldn't find "ActionEventSink.h". I assume this might happened when you tried to update to refactor classes common to iOS and macOS but couldn't actually do so as macOS is using Swift?

ActionEventSink.h is not used on macOS anymore, we don't do isolates there for now. See https://github.com/flutter/flutter/issues/65222.

It's no problem to use ObjC code from Swift. The error you had is due to a missing `pod install` invocation in the example/macos (or example/ios) folder after pulling the latest source code.

**Comment by ened** [comment]
> > Another thing I tried to create a branch based on the latest changes to update the example app. This was to show how to use the `IsolateNameServer` so that one of the actions will trigger the app to navigate to a different page when it's in the foreground. The app failed to compile on iOS though as it couldn't find "ActionEventSink.h". I assume this might happened when you tried to update to refactor classes common to iOS and macOS but couldn't actually do so as macOS is using Swift?
> 
> ActionEventSink.h is not used on macOS anymore, we don't do isolates there for now. See [flutter/flutter#65222](https://github.com/flutter/flutter/issues/65222).
> 
> It's no problem to use ObjC code from Swift. The error you had is due to a missing `pod install` invocation in the example/macos (or example/ios) folder after pulling the latest source code.

Oh, I see the error is related to some other issue as well.. the files are not listed in the iOS project. Will resolve now.

File: flutter_local_notifications/darwin/Classes/Converters.h [hunk 5]
```diff
@@ -10,8 +10,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface Converters : NSObject
 
-+ (UNNotificationCategoryOptions)parseNotificationCategoryOptions:(NSArray *)options API_AVAILABLE(ios(10.0));
-+ (UNNotificationActionOptions)parseNotificationActionOptions:(NSArray *)options API_AVAILABLE(ios(10.0));
++ (UNNotificationCategoryOptions)parseNotificationCategoryOptions:
```
MaikuB, line 13: Do correct me I'm wrong but as this is being shared, won't some tweaks need to be done check if there's a compatibile macOS version too? i.e. it'd become `API_AVAILABLE(ios(10), macosx(10.14))`

MaikuB: @ened example reference https://github.com/FirebaseExtended/flutterfire/blob/e3ff56d540db0a5c7511be5b1dfd8c0d1b86d192/packages/firebase_messaging/firebase_messaging/ios/Classes/FLTFirebaseMessagingPlugin.m#L528

ened: OK!

ened:
> Thanks @ened. Been a while since I used iOS. Ran into a weird problem though, the actions were appearing for me before and from the notification centre, I could swipe and select "View" to see the actions. However, after some point, I no longer see the actions. See below to see what I mean
> 
>  Screen.Recording.2021-12-29.at.4.34.02.pm.mov 
> 
> Any idea what might be the cause of the issue?

No idea yet - this seems to work well at the moment. I have not explored 

> A few thoughts upon looking at the PR and seeing what's there. Let me know what you think
> 
> * Some applications will want actions to trigger the app to navigate to a specific page/route when
>   
>   * when the action launches the app when it's in a terminated state. From what I could see, it doesn't seem like it's possible. On iOS, I believe it should've been possible via [UNNotificationActionOptionForeground](https://developer.apple.com/documentation/usernotifications/unnotificationactionoptions/unnotificationactionoptionforeground?changes=__7&language=objc). I suspect this isn't working as expected as if you check the numeric value in the documentation page I linked, it's `1 << 2` and that ends up with `4`. From what I could tell, the mapping logic within the PR takes the numeric `index` value of the enum and adds one to it so it ends with a value of 3 instead. I suspect `IOSNotificationCategoryOption` may have a similar issue given bitwise operations are involved as opposed to being a straightforward integer value. Putting that issue aside, is there a way to determine which action (id) launched the app? This would possible involve changes to the logic related to [getNotificationAppLaunchDetails()](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/FlutterLocalNotificationsPlugin/getNotificationAppLaunchDetails.html)

Fixed!

> * Would it be possible to show how an action could trigger another plugin like the `url_launcher` one mentioned in the updates in the readme? I don't know what would actually be involved myself. This could be done by updating the example app

Have added a basic version for this.

> * On iOS, tapping on an action dismisses a notification by default. Although it's not the default behaviour on Android, I believe most apps would also want this behaviour on Android. Is there a way to support this? It looks like the example app tried to demonstrate this by calling the `cancel()` method of the plugin . This could cause an issue for applications dealing with scheduled/recurring notifications that have actions as calling `cancel()` would also unschedule it so to speak. Perhaps this calls for having a boolean `autoCancel` property on the `AndroidNotificationAction` that defaults to true?

OK seems reasonable - will add to the todo list.

> * Was the plan to add macOS support to this PR too? Asking as the PR introduces a number of classes/enum with the `IOS` prefix. iOS and macOS share the same notification APIs though so following this convention could lead to some duplication where there'd be macOS classes and enums. Perhaps replacing `IOS` with `Apple` would be better in cases like this? I may need to see if this can be done throughout other places in the existing code of the plugin first

Yes I was hoping to get macOS support in as well, at least in the basic fashion. There are some limitations outlined in the code now.

MaikuB:
> Fixed!

Thanks I'll see if it'll cooperate again on iOS for me to try this out. Is there a way for an action to trigger the app to be launched on Android though? I'd imagine that a launch intent should be used instead but then wonder how that would play out with callbacks. Not sure if that means a separate callback is required as the current one for notification actions would be for "background" actions

ened:
Hi everyone subscribed. Please give this PR a try and support by reviewing code, writing tests and/or documentation. I think we are pretty close to a releasable version but I need your support. My time will be quite limited and I prefer not to keep this PR open for another year. ;-)
Thank you.

MaikuB:
@ened one option I can think of is to merge this in and do a pre-release. Would allow for changes to be consumed and tested more easily and reduce the need for you to deal with merge conflicts.  This would also be a good basis for the plugin to have more refactoring done to create the Darwin classes for iOS and macOS. Thanks for your efforts on this

To everyone else, I know there's a lot of subscribers on this but can I get a better indication of how many having been using @ened's fork so far? There are comments with questions or those describe issue but I wonder how many of you have successfully been using it and what feedback you have. If you could perhaps upvote on this message as indicator and optionally reply with feedback then that would be much appreciated

ened:
> @ened one option I can think of is to merge this in and do a pre-release. Would allow for changes to be consumed and tested more easily and reduce the need for you to deal with merge conflicts. This would also be a good basis for the plugin to have more refactoring done to create the Darwin classes for iOS and macOS. Thanks for your efforts on this

Yes that sounds Good. You can mark the feature as experimental somewhere in the public API. And the next major version of the plug-in could remove the flag for example.

MaikuB:
I'll do that then :)

AdamBridges:
> @ened one option I can think of is to merge this in and do a pre-release. Would allow for changes to be consumed and tested more easily and reduce the need for you to deal with merge conflicts. This would also be a good basis for the plugin to have more refactoring done to create the Darwin classes for iOS and macOS. Thanks for your efforts on this
> 
> To everyone else, I know there's a lot of subscribers on this but can I get a better indication of how many having been using @ened's fork so far? There are comments with questions or those describe issue but I wonder how many of you have successfully been using it and what feedback you have. If you could perhaps upvote on this message as indicator and optionally reply with feedback then that would be much appreciated

I was using ened's fork in September 2021 but it was having compatibility issues with other plugins as this fork's dependencies were outdated then and assumed this wasn't being worked on anymore, so I copied ened's fork and updated things on my own and have been using it ever since. Anyway, I suspect it's all fixed now so I'll try this current version going forward.

MaikuB:
@AdamBridges thanks for the input. What kind of use cases did you have in your app and was the actions and the way they were handled in this PR sufficient to cover them?

MaikuB:
> Is there a way to prevent the Android device to kill the isolate when the user taps the action button? I mean, would it be a 
> foreground service that requires a notification to give a feedback to the user? Or we don't need to do nothing because
>Flutter Engine handle this by itself?

> I am talking about the Required notifications section of this link

@lucasribolli late response but I suspect you might be confusing some things. An isolate is a concept specific to Flutter and a background isolate is spun when the user taps on action. That's not related to foreground services that is mentioned in the link. Furtherrmore, the plugin already supports foreground services. Clone the repository and run the example app on Android to take a look

AdamBridges:
> @AdamBridges thanks for the input. What kind of use cases did you have in your app and was the actions and the way they were handled in this PR sufficient to cover them?

The use case included 3 actions that would either (1) store data to local/remote databases without opening the app; (2) open the app to a specific page; or (3) would reschedule the notification to fire again in 1-3 hours without opening the app. All of these actions work on Android except (2) and I had to resort to using an Android Intent package.

MaikuB:
@AdamBridges how did you handle (2) on iOS and application lifecycle i.e. navigating to a page when it app is running or terminated?

AdamBridges:
@MaikuB I had issues with getting it to play nice on iOS so I used another package to handle notifications on iOS: The other package is capable of launching the app from an action and then there's logic in place to navigate the user from there. Messy, but it worked. 

I'll have an opportunity to test the notifications on iOS and Android with only this package in a couple weeks from now if you want to ask me about it again then. Otherwise, I'll try to remember to post my experiences then.

lucasribolli:
> To everyone else, I know there's a lot of subscribers on this but can I get a better indication of how many having been using @ened's fork so far? There are comments with questions or those describe issue but I wonder how many of you have successfully been using it and what feedback you have. If you could perhaps upvote on this message as indicator and optionally reply with feedback then that would be much appreciated

I have been used this fork for some months and it seems to be good. But I have one issue to report that I cannot identify the cause. In iOS it seems not possible to use the own `FlutterLocalNotificationsPlugin` to schedule notifications inside the background isolate. I am trying to schedule new notifications using the same methods used in the main isolate, with `initialize` etc. The requisite is, when user press the action button, a notification are scheduled. 
It works well on Android, no issues.

ened:
@AdamBridges @lucasribolli really appreciate you took the time and wrote some feedback. We can look into getting the component improved in the future.

@MaikuB let's lock this conversation and add a new label/tag (perhaps issue template) for the issue tracker so that future bug reports can be found quickly.

MaikuB:
Yep makes sense. I've created a "notification actions" labels and will close and lock this. If you have some feedback or issue, relating to the functionality that's currently in the pre-release, please file a new issue with the label

flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
```diff
@@ -234,6 +252,78 @@ - (void)pendingNotificationRequests:(FlutterResult _Nonnull)result {
...
+                           options:[Converters parseNotificationCategoryOptions:
+                                                   category[@"options"]]];
+
+        [newCategories addObject:newCategory];
+      }
+
+      if (newCategories.count > 0) {
+        UNUserNotificationCenter *center =
+            [UNUserNotificationCenter currentNotificationCenter];
+        [center getNotificationCategoriesWithCompletionHandler:^(
```
MaikuB, line 311: @ened I figured out what was the cause behind why notification actions stopped appearing for me on iOS and narrowed it to here. removing the logic to query the existing categories and then appending to that list allowed me to see the notification actions again. was there a reason for querying this? from my own tests and what i've seen in docs, should be able to just call `[center setNotificationCategories:newCategories];`

flutter_local_notifications/ios/Classes/FlutterEngineManager.m
```diff
@@ -0,0 +1,92 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+static FlutterEngine *backgroundEngine;
+
+@implementation FlutterEngineManager {
+  NSUserDefaults *_persistentState;
+}
+
++ (BOOL)shouldAddAppDelegateToRegistrar:
```
MaikuB, line 23: Hi @ened, I know it's been a long time now since this PR but do you remember the reason for this method? I was looking to see if this could be removed. A while back, the plugin was bumped that it no longer supports old notification APIs that in turn required hooking into the app delegate. I removed the methods tied to the delegate but realised it's still registered. Looking to see if the registration can be removed. I believe so but wanted to check with you in case I've missed something

MaikuB: Ah sorry I spoke too soon. I missed that removing the registration of the plugin as the app delegate had an effect on the `UNUserNotificationCenterDelegate`  methods from being fired. I forgot tjay the Flutter engine is detecting if a plugin has registered as an application delegate to then try to fire the appropriate `UNUserNotificationCenterDelegate` methods

Issue #912, opened 2020-11-19
activity-alias support


crazecoder:
```xml
<activity-alias
            android:name="DefaultAlias"
            android:enabled="true"
            android:targetActivity=".MainActivity">
```
```
java.lang.ClassNotFoundException: com.flutter.exanple.DefaultAlias
at java.lang.Class.classForName(Native Method)
at java.lang.Class.forName(Class.java:454)
at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.getMainActivityClass(FlutterLocalNotificationsPlugin.java:550)
```

MaikuB:
Any chance you could submit a PR to resolve this issue? As Flutter's standard application template doesn't use aliases (at least last time I checked), this looks like a very specific use case. Furthermore,  this is new concept for me as I don't do "pure" native Android development

MaikuB:
Closing as PR was merged and released. Thanks

PR #919, opened 2020-11-21
fix ClassNotFoundException when use activity-alias

crazecoder:
#912

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -160,7 +160,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
 
     private static Notification createNotification(Context context, NotificationDetails notificationDetails) {
         setupNotificationChannel(context, NotificationChannelDetails.fromNotificationDetails(notificationDetails));
-        Intent intent = new Intent(context, getMainActivityClass(context));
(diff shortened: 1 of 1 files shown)


flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -553,11 +557,18 @@ private static void setTimeoutAfter(NotificationDetails notificationDetails, Not
 
     private static Class getMainActivityClass(Context context) {
         String packageName = context.getPackageName();
-        Intent launchIntent = context.getPackageManager().getLaunchIntentForPackage(packageName);
-        String className = launchIntent.getComponent().getClassName();
+        PackageManager packageManager = context.getPackageManager();
```
MaikuB, line 560: Thanks for the PR. I just realised that given `getMainActivityClass` was only really to specify the intent relating to the launch activity, I suspect this entire method could be removed. That is, the intent from calling `packageManager.getLaunchIntentForPackage(packageName)` could be passed to the notification itself (i.e. [here](https://github.com/MaikuB/flutter_local_notifications/blob/88ff541da3bfe5af877724a472967467b731f6d6/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L162). I imagine that should still solve the issue you had with activity aliases. If so, could you make the change I described in this PR?

MaikuB:
Thanks for making the changes. Will find some time later in the week to give this a spin before it's merged in. You can ignore the macOS build errors as it's an issue with Flutter on macOS where they have a fix that is only on master channel and the build script uses the dev channel

Issue #924, opened 2020-11-24
HTTP not allowed error in example app


iqfareez:
Tried to show big picture notification. 
The notification didn't show.

The log:

```
D/ViewRootImpl@e5e2f46[MainActivity](31169): ViewPostIme pointer 0
D/ViewRootImpl@e5e2f46[MainActivity](31169): ViewPostIme pointer 1
E/flutter (31169): [ERROR:flutter/lib/ui/ui_dart_state.cc(177)] Unhandled Exception: Bad state: Insecure HTTP is not allowed by platform: http://via.placeholder.com/48x48
E/flutter (31169): #0      _HttpClient._openUrl (dart:_http/http_impl.dart:2435:7)
E/flutter (31169): #1      _HttpClient.openUrl (dart:_http/http_impl.dart:2342:7)
E/flutter (31169): #2      IOClient.send (package:http/src/io_client.dart:31:37)
E/flutter (31169): #3      BaseClient._sendUnstreamed (package:http/src/base_client.dart:91:38)
E/flutter (31169): #4      BaseClient.get (package:http/src/base_client.dart:27:7)
E/flutter (31169): #5      get.<anonymous closure> (package:http/http.dart:46:36)
E/flutter (31169): #6      _withClient (package:http/http.dart:166:20)
E/flutter (31169): #7      get (package:http/http.dart:46:5)
E/flutter (31169): #8      _HomePageState._downloadAndSaveFile (package:flutter_local_notifications_example/main.dart:781:42)
E/flutter (31169): <asynchronous suspension>
E/flutter (31169): #9      _HomePageState._showBigPictureNotification (package:flutter_local_notifications_example/main.dart:788:40)
E/flutter (31169): #10     _HomePageState.build.<anonymous closure> (package:flutter_local_notifications_example/main.dart:410:33)
E/flutter (31169): #11     _InkResponseState._handleTap (package:flutter/src/material/ink_well.dart:985:20)
E/flutter (31169): #12     _InkResponseState.build.<anonymous closure> (package:flutter/src/material/ink_well.dart:1101:38)
E/flutter (31169): #13     GestureRecognizer.invokeCallback (package:flutter/src/gestures/recognizer.dart:183:24)
E/flutter (31169): #14     TapGestureRecognizer.handleTapUp (package:flutter/src/gestures/tap.dart:598:11)
[...]

PR #925, opened 2020-11-24
[flutter_local_notifications] update example app to point https placeholder image URLs

iqfareez:
Fix the said issue #924 in `flutter_local_notifications\example` app. 

Tested on Android 9.0 Pie. Run on debug build. Works as expected.

Since I didn't have any iOS device, I didn't make any changes in iOS side. 

https://flutter.dev/docs/release/breaking-changes/network-policy-ios-android
https://flutter.dev/docs/development/add-to-app/ios/project-setup#local-network-privacy-permissions


MaikuB (changes requested):
Thanks for the PR. I believe the site that was being called before and failed now offers a https url that you should be able to just change the url and revert the change to the `AndroidManifest.xml` file

iqfareez:
Changes done as requested :wrench:

MaikuB:
Thanks, will test it out and then merge it in. I'll also change the PR title since the way it got fixed has changed and to better reflect that it's to do with the example app

PR #927, opened 2020-11-25
Add note about additional failure mode

srsudar:
On a Pixel 4XL I was successfully scheduling notifications but they were
never showing. Turns out this was because I was missing some of the
release build setup. The notification was firing but not showing,
because the system errored out when it couldn't find the resource.

This commit adds a note to the README about this failure mode.


flutter_local_notifications/README.md
```diff
@@ -169,7 +169,7 @@ Note that when a full-screen intent notification actually occurs (as opposed to
...
+⚠️ Ensure that you have configured the resources that should be kept so that resources like your notification icons aren't discarded by the R8 compiler by following the instructions [here](https://developer.android.com/studio/build/shrink-code#keep-resources). Without doing this, you might not see the icon you've specified in your app's notifications, or your notifications might never show, instead silently failing when they look for a removed resource. The configuration used by the example app can be found [here](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/example/android/app/src/main/res/raw/keep.xml) where it is specifying that all drawable resources should be kept, as well as the file used to play a custom notification sound (sound file is located [here](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/example/android/app/src/main/res/raw/slow_spring_board.mp3)).
```
MaikuB, line 172: One thing that I've been guilty of doing before is around having long sentences. Would you be able to break this up? Perhaps a separate sentence to state the specific scenario (scheduling notifications) and/or one that starts with something like "Worst case..."

MaikuB (changes requested):
Thanks for the PR and good pick up. Left a comment to see if a minor tweak could be done to break up the sentence as it feels a bit on the long side

flutter_local_notifications/README.md
srsudar: Sure, done.

Sorry for the double commit on the typo. This was a tricky file to edit without line breaks.

MaikuB: No worries

Issue #506, opened 2020-02-29
NPE in setupNotificationChannel on Galaxy A70s


dluksza:
I only know about this issue from crash reports in Google Play Console. I was not able to reproduce it on my own device nor on emulator.

I've got two or three reports from users that said that _notifications doesn't work_, unfortunately don't know what device they are using.

This error seems to be present since 0.8.4 up to 1.1.2. I'm going to release new app version with 1.2.0 soon. I do assume that this error is still present and probably it is related to some battery optimizations Samsung did on their side. But anyway wanted to report this, maybe there would be some kind of _nice_ workaround for it ;)

Here is the stacktrace I've got for Google Play Console:
```
java.lang.RuntimeException: 
  at android.app.ActivityThread.handleReceiver (ActivityThread.java:3624)
  at android.app.ActivityThread.access$1300 (ActivityThread.java:240)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1808)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:214)
  at android.app.ActivityThread.main (ActivityThread.java:7094)
  at java.lang.reflect.Method.invoke (Method.java)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:494)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:975)
Caused by: java.lang.NullPointerException: 
  at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.setupNotificationChannel (FlutterLocalNotificationsPlugin.java:46)
  at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.createNotification (FlutterLocalNotificationsPlugin.java)
  at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.showNotification (FlutterLocalNotificationsPlugin.java)
  at com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver.onReceive (ScheduledNotificationReceiver.java:77)
  at android.app.ActivityThread.handleReceiver (ActivityThread.java:3615)
  at android.app.ActivityThread.access$1300 (ActivityThread.java:240)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1808)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:214)
[...]

PR #944, opened 2020-12-09
[flutter_local_notifications] Fix potential android npe

dluksza:
From time to time I've got an NPE from app user coming from native Android code of `flutter_local_notifications`. I wasn't able to reproduce those errors locally. But patching the code and releasing new app with forked version seems to solve those random NPE's. I haven't seen them for about a week now.


MaikuB:
From what I remember from previous discussions, your issues were occurring as some configuration was missing (https://github.com/MaikuB/flutter_local_notifications/issues/506). Adding the configuration wasn't going retroactively fix issues for existing users when the app was misconfigured as you're still going to have data that wasn't serialised properly. The reason why your fix will work is applies fallback values to cater for these cases. This will make it more difficult to actually find out when a legitimate issue had occurred. As such, I won't be looking to take a PR on this at this point in time. Should you happen to find out how to reproduce this for an app that had been configured properly with details on how to reproduce this then I'd be happy to look at this again.

corcoran:
FWIW, we're also looking at a similar issue that we cannot reproduce locally that this PR seems to fix.

PR #963, opened 2020-12-31
[flutter_local_notifications] Add ability to set "subText" property in Android notifications

sidlatau:


MaikuB:
Thanks for the PR. Will try to take a look at this later in the week

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -251,4 +252,20 @@ class AndroidNotificationDetails {
...
+  /// This provides some additional information that is displayed in the
+  /// notification. No guarantees are given where exactly it is displayed.
+  ///
+  /// This information should only be provided if it provides an essential
+  /// benefit to the understanding of the notification. The more text you
+  /// provide the less readable it becomes. For example, an email client should
+  /// only provide the account name here if more than one email account has
+  /// been added.
+  ///
+  /// As of Build.VERSION_CODES.N this information is displayed in the
```
MaikuB, line 265: `Build.VERSION_CODES.N` is a constant within the Android SDK, could replace this with the Android version number as that's the convention that's been used within the plugin. This was done as Flutter developers may not have any familiarity with Android development to know the version codes or API levels. Outside of software development Android versions are now being referred to by numbers as well, which in this case is 7.0

MaikuB (changes requested):
Besides the requested doc comment change, everything else looks good to me. If you'd like to bump the plugin version and add a changelog entry, then I could merge it in and do a release after that though I could also do that as well so not a biggie :)

sidlatau:
Thanks for the review! I updated PR according comments.

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -251,4 +252,20 @@ class AndroidNotificationDetails {
   /// For a list of a values, refer to the documented constants prefixed with "FLAG_" (without the quotes) at https://developer.android.com/reference/android/app/Notification.html#constants_1.
   /// For example, use a value of 4 to allow the audio to repeat as documented at https://developer.android.com/reference/android/app/Notification.html#FLAG_INSISTEN
   final Int32List additionalFlags;
+
+  /// This provides some additional information that is displayed in the
```
MaikuB, line 256: Sorry missed this earlier but this plugin is following the effective dart documentation guidelines and currently the way this is done violates [this rule]( https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph) and [this rule](https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs). Whilst the latter is a preference, it has been used for this plugin (e.g. "Specifies...")

Could you look at this changing the documentation to meet the guidelines and what's been done in the plugin so far? I know this is currently lifted from the official Android documentation.

MaikuB (changes requested):
Thanks for that. Sorry I missed a part on the API docs in my previous. See the comment I left in the PR itself for more details

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
sidlatau: No problem, consistency is important.  I updated the PR, please check it again.

MaikuB (approved):

Issue #971, opened 2021-01-07
[Feature request] Improve API for setting intervals


erf:
I previously used `showDailyAtTime` to set a daily reminder - that was a pretty nice API. 

I saw it was deprecated, and now to add intervals was pretty rough..

- First you have to get a time zone, by initializing a database and calling either a platform call or getting a third party package. 
- Then you have to be sure the TZDateTime, given the timezone location, has a date in the future ( or it will crash ), even though i'm only interested in the hour and seconds of the day.
- Then you have to pick the `DateTimeComponents` which only have two items.

This was a lot of work to get up and running. I think this could be improved upon.

MaikuB:
You can still do daily with the new approach. Yes it's more work but the old approach had issues with daylight savings and when users changed locations that they end up in a different timezone. Even though that might not be concern for your app, as this plugin needs to work correctly in the general case this is why the new approach has been put in. 

If you'd like to stick with the old approach then you can stick with old versions of the plugin, maintain a fork or look at submitting a PR as I'm not going to keep the old approach given the issues it had

erf:
I got the new version working. I think it makes sense to build on that, but perhaps make the API a bit nicer.

I would suggest to automate / hide the time zone stuff for the user, as i think most users would prefer to use the local time, and rather make the location optional.

I also think avoiding crashing if the time is older, if you use `DateTimeComponents.time` as you only want remind at a specific time a day, so if the last year was used it should not matter.

Also `DateTimeComponents` could perhaps be extended, perhaps motivated by Swifts `DateComponents`.

MaikuB:
Whilst trying to follow the date components on iOS is a nice idea, it'd take a lot of work to do because Android doesn't have any support for that and has required custom code to implement what is there right now. As such, if you'd like to see changes in then this will require contribution from the community such as yourself as this is all done in my spare time. I'm happy to provide input but I don't have the time to be looking at implementing this

erf:
Ok, maybe drop the `DateComponents` suggestion for now, but my first two points, that is: "hide the time zone" and "allow older dates", could be fixed without too much work i think.

MaikuB:
Even still you're going to need to look at PR for that. The concept of hiding timezones can lead to issues. For example, what's expected when the user goes to a different timezone

erf:
I might look into making a PR for this, but i rather to discuss and clear up these details with you first. 

You could perhaps make an API that says `setIntervalForLocalTimezone` or something like that. 

[...]

PR #974, opened 2021-01-07
[flutter_local_notifications] Allow past dates when using DateComponents and other fixes

erf:
This fixes the problem with not being allowed to use previous dates when doing repeating daily notifications as discussed in #971. I also added a new example for this.

I also renamed `zonedSchedule` per platform to easier distinguish between these. These methods are however very similar and i think it makes sense to refactor these into a single `zonedSchedulePlatform` in a future PR, by making a common `NotificationsDetails` class.

I also added `.vscode` to gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 pubspec.lock
+.vscode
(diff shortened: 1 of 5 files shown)


erf:
I see now i should not have named the method names, because you have these various classes. I will make a new PR.

PR #975, opened 2021-01-08
[flutter_local_notifications] Check if parcelable notification is present on received intent

kaptnkoala:
Using the FlutterLocalNotification plugin in a production app we get some crashes reported to Crashlytics, because there is no "notificationDetails" or "notification" item attached to the intent. We were not able yet to identify the cause, where these incomplete notifications gets scheduled. But we think it would be good to have this sanity check in the notification receiver, dropping any malformed notifications that are received.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/ScheduledNotificationReceiver.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/ScheduledNotificationReceiver.java
@@ -29,14 +29,15 @@ public class ScheduledNotificationReceiver extends BroadcastReceiver {
         if (StringUtils.isNullOrEmpty(notificationDetailsJson)) {
             // This logic is needed for apps that used the plugin prior to 0.3.4
             Notification notification = intent.getParcelableExtra("notification");
-            notification.when = System.currentTimeMillis();
-            int notificationId = intent.getIntExtra("notification_id",
-                    0);
-            NotificationManagerCompat notificationManager = NotificationManagerCompat.from(context);
-            notificationManager.notify(notificationId, notification);
-            boolean repeat = intent.getBooleanExtra("repeat", false);
-            if (!repeat) {
-                FlutterLocalNotificationsPlugin.removeNotificationFromCache(context, notificationId);
+            if (notification != null) {
+                notification.when = System.currentTimeMillis();
+                int notificationId = intent.getIntExtra("notification_id", 0);
+                NotificationManagerCompat notificationManager = NotificationManagerCompat.from(context);
+                notificationManager.notify(notificationId, notification);
+                boolean repeat = intent.getBooleanExtra("repeat", false);
+                if (!repeat) {
+                    FlutterLocalNotificationsPlugin.removeNotificationFromCache(context, notificationId);
+                }
             }
         } else {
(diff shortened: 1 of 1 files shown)


MaikuB:
I noticed the change affects an old area of the code that shouldn't have been applicable to your application unless your application had used really early versions of the plugin. Do you have details on what versions your application had been using? The other thing that this issues indicates is that the application didn't complete the release build configuration. Particularly around having a Proguard rules file. This is another area you can check as there's gap there then your issue is due to misconfiguration. More recent versions of the plugin has minimised the need for this configuration and you can check by looking the version of the plugin is being used, pick the appropriate tag and look at the Proguard rules file of the example app.

Regarding the PR itself, the only difference here is a null check so it won't have an exception. It doesn't drop the notification as the plugin will reschedule notifications when a device is rebooted

kolotum:
We have rebuilt our existing app with Flutter. Our first flutter app version contained flutter_local_notifications version 3.0.1+6.
We have already added the mentioned Proguard rules in a later app update. But the crashes did still exist until we added the sanity check. We have never published any flutter_local_notifications version lower than 3.0.

MaikuB:
Thanks @kolotum for the info. So you mean you originally didn't add the rules? The configuration is needed that the notification data can be serialised and saved to shared preferences. If you originally missed this then you would have malformed data being saved as the rules at that point in time. Adding the rules later is not something that will fix issues retrospectively and this is expected the plugin is not going to be able to recover that data

kaptnkoala:
Hey @MaikuB, thanks for your extended response on this. Sorry for getting back so late on this issue!!

We also hoped that this problem would vanish when we explicitly added the missing GSON rules to the ProGuard configuration. But the amount of crashes we saw on Crashlytics stayed the same as before and did not decline over time, as we would have expected as all malformed notifications would have been rescheduled with valid ones. Also, when we increased the rollout percentage, that amount of crashes raised proportional, so also new users with matching ProGuard rules were facing the crash.

We checked the configuration.txt file that is created by the obfuscated Android build and realized, that another library had already defined most of the ProGuard rules needed for GSON. So probably we already had sufficient rules for GSON before we added the explicit config that we took from your example app.

What effect would it have if the ProGuard configuration was still broken? Would every notification have no `notificationDetails` item and crash the app on receive? 

Should we add some temporary analytics events to our fork, to get a grasp of the amount of valid and malformed notifications that we receive?

MaikuB:
>  But the amount of crashes we saw on Crashlytics stayed the same as before and did not decline over time, as we would have expected as all malformed notifications would have been rescheduled with valid ones.

@kaptnkoala There is data saved to shared preferences that is read when the device reboots to reschedule the alarm so you'd likely still see issues because that data is still there. That would still be malformed given adding rules later isn't a retrospective fix existing scheduled notifications. This is why I said your PR doesn't drop notifications as you thought it would. It's just silently ignoring the issue.

I would hope off on adding analytics for now as based on what you said, I think there's still an issue around the rules. If you could share which version of the plugin you used and the entire rules file that was in use when this issue occurred then that would be a good starting point to see what's going on

MaikuB:
Closing this as from what I'm reading so far, apps were shipped without having the Proguard rules. Adding the changes in this PR would mean there's no way to get remote crash reporting on when app has been misconfigured, thereby removing one way of knowing that there's an issue to be fixed. Those that want their app to ignore these malformed notifications will need to keep their own fork

PR #978, opened 2021-01-10
[flutter_local_notifications] Allows past dates when using matchDateComponents

erf:
This fixes the problem with not being allowed to use previous dates when doing repeating daily notifications as discussed in #971. I also added a new example for this.


flutter_local_notifications/example/lib/main.dart
```diff
@@ -316,6 +316,14 @@ class _HomePageState extends State<HomePage> {
                         await _scheduleDailyTenAMNotification();
                       },
                     ),
+                    PaddedRaisedButton(
+                      buttonText:
+                          'Schedule daily in 5 seconds last year notification '
```
MaikuB, line 321: Can you check the text and method name to say it'll schedule a daily 10:00 am notification similar to the other examples? The reason for this is that it's more difficult to figure out the time with the example that is there now. Part of this is due how it references 5 seconds in the future but uses a date last year. Expressing this as an actual time makes it more clearer that the time is constant as that is people usually try to schedule something that is recurring. As an example, in normal conversation one is more likely to say they go to the gym everyday at 7:00 AM instead as opposed to saying they go to the gym everyday in 5 seconds

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
```diff
@@ -298,7 +298,7 @@ class FlutterLocalNotificationsPlugin {
               id, title, body, scheduledDate, notificationDetails?.android,
               payload: payload,
               androidAllowWhileIdle: androidAllowWhileIdle,
-              matchDateComponents: matchDateTimeComponents);
+              matchDateTimeComponents: matchDateTimeComponents);
```
MaikuB, line 301: Thanks for picking up on this. I'll need to bump the plugin more as this is actually a breaking change, which in itself is fine. I'll do this in another PR as there's another change I'd like to do before this would get released

flutter_local_notifications/lib/src/helpers.dart
```diff
@@ -1,3 +1,4 @@
+import 'package:flutter_local_notifications/src/types.dart';
```
MaikuB, line 1: Use relative import

flutter_local_notifications/example/lib/main.dart
```diff
@@ -1131,6 +1158,19 @@ class _HomePageState extends State<HomePage> {
     return scheduledDate;
   }
 
+  tz.TZDateTime _nextInstanceOfInFiveSecondsLastYear() {
+    final now = tz.TZDateTime.now(tz.local);
```
MaikuB, line 1162: Type needs to be specified. I don't intend to remove this rule. Part of the reason why is Dart allows for dynamics. The 80 line length rule is more of a Flutter/Dart issue. For what it's worth, the Flutter repos requires code to specify types as well.

MaikuB (changes requested):
Thanks for the PR, see comments for changes that I've requested. Some of these are to do with issues that should've been picked up by the analyser.

flutter_local_notifications/example/lib/main.dart
erf: I understand, but the reason i choose "in 5 seconds", is so that we are actually able to test it, and don't have to wait for the next 10:00 am to actually see the notification. Should not people actually be able to test / see the notification in the examples?

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
erf: So i should include change? 

I also noticed i could rename the variable name `validateDateIsInTheFuture.matchDateTimeComponents`, should i do this?

flutter_local_notifications/example/lib/main.dart
MaikuB: They can but it wasn't meant to be immediate for the same reasons I explained earlier regarding "communication" and how recurring notifications/events are usually scheduled. Given, it's recurring, users still actually need to wait to validate that it repeats. In <x amount of time> requires a bit of calculation. Developers can still change the code to suit the time if needed

erf: Ok, but i should still note that it's last year right?

MaikuB: Yep so the text and method name would refer to something along the lines of "Schedule daily 10:00:00AM notification using last year's date"

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
MaikuB: Which change are you referring to including? Normally, it's better to separate out these changes to another PR since one adds functionality whilst another introduces a breaking change to isolate the changes in case a rollback is needed but this seems straightforward enough that I'm ok with it.

Yep forgot to say `validateDateIsInTheFuture.matchDateTimeComponents` could be renamed as well but forgot this as I had left my original comments in the old PR that was closed

erf: Ok, i think i've fixed all the comments now.

erf:
I think all the comments should be fixed and pushed now. Let me know if there is anything else.

flutter_local_notifications/example/lib/main.dart
```diff
@@ -316,6 +316,14 @@ class _HomePageState extends State<HomePage> {
                         await _scheduleDailyTenAMNotification();
                       },
                     ),
+                    PaddedRaisedButton(
+                      buttonText:
+                          'Schedule daily 10:00:00 am notification in your '
+                          'local time zone given last year',
```
MaikuB, line 322: there's a grammar issue with ending this sentence with `given last year`.  Unless you have another example that doesn't have a grammar issue, `...using last year's date` that  I gave as an example before should work

erf: fixed

MaikuB:
Thanks for making the changes. The other thing I had forgotten to ask that's actually more important is if you have tried testing these changes on all platforms yet, including using the old iOS notification APIs (iOS versions older than 10 use this)? The reason I'm asking the APIs were quite different that I believe the behaviour was if it's date in the past, it'll show a notification immediately and then progress to having a recurring notification. I think this was the other reason why I tried to make it a date in the future but had just remembered this now so apologies about that. At the moment I won't get be able to test this until a few days later or so. In the absence of having an older iOS simulator or device, one way of checking this is to remove the code on the iOS side that runs when the version check finds that the device is on iOS 10 or newer such that the code that remains uses the `UILocalNotification` APIs

erf:
I've only tested it on a newer iPhone SE now. I can try to test using the old method and also test on an Android device.

erf:
I've tested this now on the iOS simulator and on an Android device after forcing the use of `UILocalNotification` like you suggested. Is that enough or should i test on macOS also?

MaikuB:
If you could check macOS as well then that would be great. Though in case you're not aware of it. You'll need to go make sure you're on the right channel etc to get desktop working. Last time I tried macOS there was an issue with having to click twice that I think was a known issue with Flutter on the latest version of macOS

erf:
I tried to run macOS now, and it just show up as a gray box and i got the following error messages:

```
aunching lib/main.dart on macOS in debug mode...
Running pod install...                                           1,087ms
Building macOS application...
2021-01-14 11:59:54.424 flutter_local_notifications_example[14704:2220823] Failed to find path for "flutter_assets"
2021-01-14 11:59:54.424 flutter_local_notifications_example[14704:2220823] Failed to find path for "flutter_assets"
embedder.cc (860): 'FlutterEngineInitialize' returned 'kInvalidArguments'. The assets path in the Flutter project
arguments was missing.
2021-01-14 11:59:54.424 flutter_local_notifications_example[14704:2220823] Failed to initialize Flutter engine: error 2
embedder.cc (1427): 'FlutterEngineSendPointerEvent' returned 'kInvalidArguments'. Engine handle was invalid.
embedder.cc (1427): 'FlutterEngineSendPointerEvent' returned 'kInvalidArguments'. Engine handle was invalid.
embedder.cc (1427): 'FlutterEngineSendPointerEvent' returned 'kInvalidArguments'. Engine handle was invalid.
...
```

MaikuB:
Hmm not sure what's going on there. Leave the testing with me then for that one then. I would've tested it anyway but helping another set of hands etc helps :)

erf:
Maybe it's because i'm on a m1 mac..

MaikuB:
I ran into a similar issue on the macOS side and ended up recreating the example on the macOS side. Works fine now and was able to test your changes so will merge this in. Thanks :)

PR #980, opened 2021-01-11
[flutter_local_notifications] Replace commit() by apply() in Android implementation to avoid blocking UI thread

davidlepilote:
Modify commit() to apply() to avoid framerate falls when scheduling many notifications. commit() is called on the main thread and is blocking to the contrary of apply(), and it should lead to no bug due to asynchronism whatsoever.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -273,7 +273,22 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
         SharedPreferences sharedPreferences = context.getSharedPreferences(SCHEDULED_NOTIFICATIONS, Context.MODE_PRIVATE);
         SharedPreferences.Editor editor = sharedPreferences.edit();
         editor.putString(SCHEDULED_NOTIFICATIONS, json);
-        editor.apply();
+        tryCommittingInBackground(editor, 3);
+    }
+
+    private static void tryCommittingInBackground(final SharedPreferences.Editor editor, final int tries) {
+        if(tries == 0) {
+            return;
+        }
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                final boolean isCommitted = editor.commit();
+                if(!isCommitted) {
+                    tryCommittingInBackground(editor, tries - 1);
+                }
+            }
+        }).start();
     }
 
     static void removeNotificationFromCache(Context context, Integer notificationId) {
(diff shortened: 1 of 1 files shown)


MaikuB:
There can be an issue with using `apply()` as it doesn't wait until the changes are saved to storage. This is mentioned in the docs [here](https://developer.android.com/reference/android/content/SharedPreferences.Editor#apply()).

One important of that to highlight is the following

```
Unlike commit(), which writes its preferences out to persistent storage synchronously, apply() commits its changes to the in-memory SharedPreferences immediately but starts an asynchronous commit to disk and you won't be notified of any failures.
```

This means if `apply()` was used and the notification didn't get saved to storage, the user would've thought their notification had been scheduled fine. If the user's device is rebooted, then the notification wouldn't have been saved in storage to reschedule it as alarms are lost when a device is turned off.

As such, I'm not looking to take merge the PR in at this stage as it requires more changes to avoid that scenario where the `commit()` is needed to ensure the data is saved to storage. Perhaps the `shared_preferences` plugin has a way of tackling this that could be used here

davidlepilote:
The point is, the code does not check any failure on the `commit()` method. It means that if any failure occurs, anyway it won't be handled by the code right now. It is thus as safe as using `apply()`.

The documentation is crystal clear about this fact : 
```
As SharedPreferences instances are singletons within a process, it's safe to replace any instance of commit() with apply() if you were already ignoring the return value.
```
Also, calling `commit()` on a lot of scheduled notifications (let's say, 100) _does really_ affect the framerate of the app.

Finally, the only advantage of using `commit()` is to take advantage from the fact that you can handle errors by getting its return value, which is _not_ the case in the plugin. So there is no apparent reason not to use `apply()`here.

MaikuB:
Noted, in this case it looks like there's a gap on my part with regards to error handling so would be good to address that. It looks like the `shared_preferences` plugin addresses this with the use of `AsyncTask` as seen [here](https://github.com/flutter/plugins/blob/a0e793734eac4eeb566ce090c64ed928e72214b2/packages/shared_preferences/shared_preferences/android/src/main/java/io/flutter/plugins/sharedpreferences/MethodCallHandlerImpl.java#L119) that commits in the background and still makes use of the return value. Would this approach also solve the issue you're experience and if so, would it be possible to have the PR updated to make use of this? I would guess it would have similar performance as `apply()` whilst still allowing for the return value to be used.

If you'd have time to help address the gap with the error handling as well then that would be great but can understand if you only have to only look at making a change on how data is saved

davidlepilote:
Hi, no problem and you don't need to apologize at all. Your library is awesome.
I committed a new version calling `commit()` inside a `Thread` (`AsyncTask` is now deprecated in Android) and retrying three times if failing. I think it would address both our problems.

MaikuB:
Thanks for that. Forgot it's deprecated as I'm not a native Android dev but saw that mentioned during discussions. I will test this out when I get the chance to and will merge it after that. Thanks for taking the time to improve the plugin

PR #983, opened 2021-01-12
[flutter_local_notifications] Add tag and threadIdentifier platform-specific properties

mchudy:
This PR adds two platform-specific properties to `NotificationDetails`
- `tag` for Android which can be used for replacing notifications
- `threadIdentifier` for iOS and macOS which can be used for grouping notifications

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -37,7 +37,7 @@ A cross platform plugin for displaying local notifications.
    - [Periodically show a notification with a specified interval](#periodically-show-a-notification-with-a-specified-interval)
    - [Retrieveing pending notification requests](#retrieveing-pending-notification-requests)
    - [[Android only] Retrieving active notifications](#android-only-retrieving-active-notifications)
-   - [[Android only] Grouping notifications](#android-only-grouping-notifications)
+   - [Grouping notifications](#grouping-notifications)
    - [Cancelling/deleting a notification](#cancellingdeleting-a-notification)
    - [Cancelling/deleting all notifications](#cancellingdeleting-all-notifications)
    - [Getting details on if the app was launched via a notification created by this plugin](#getting-details-on-if-the-app-was-launched-via-a-notification-created-by-this-plugin)
@@ -448,7 +448,7 @@ await flutterLocalNotificationsPlugin.periodicallyShow(0, 'repeating title',
     androidAllowWhileIdle: true);
 ```
 
-### Retrieveing pending notification requests
+### Retrieving pending notification requests
 
 ```dart
 final List<PendingNotificationRequest> pendingNotificationRequests =
@@ -465,10 +465,20 @@ final List<ActiveNotification> activeNotifications =
         ?.getActiveNotifications();
 ```
 
-### [Android only] Grouping notifications
+### Grouping notifications
+
+#### iOS
+
+For iOS, you can specify `threadIdentifier` in `IOSNotificationDetails`. Notifications with the same `threadIdentifier` will get grouped together automatically.
+
+```dart
+const IOSNotificationDetails iOSPlatformChannelSpecifics =
+    IOSNotificationDetails(threadIdentifier: 'thread_id');
+```
+
+#### Android
 
 This is a "translation" of the sample available at https://developer.android.com/training/notify-user/group.html
-For iOS, you can specify `threadIdentifier` in `iOSNotificationDetails`. You could also display the summary notification (not shown in the example) as otherwise the following code would show three notifications.
 
 ```dart
 const String groupKey = 'com.android.example.WORK_EMAIL';
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -35,7 +35,7 @@ A cross platform plugin for displaying local notifications.
    - [Displaying a notification](#displaying-a-notification)
    - [Scheduling a notification](#scheduling-a-notification)
    - [Periodically show a notification with a specified interval](#periodically-show-a-notification-with-a-specified-interval)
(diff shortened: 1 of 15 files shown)


Albert221:
I'm looking forward to merging this PR, `tag` is a really handy feature, and in some use-cases, there's no way to implement features without it.

MaikuB:
Thanks for the PR, would you be able to update the example app to demonstrate the usages of these new properties?

Another thing I'm thinking about here is if notifications can be added with a tag, then having a (platform-specific) method  in the Android implementation of the plugin might be appropriate. Unless I've missed a reason this wouldn't be necessary

mchudy:
I added examples for the new properties. As for the platform-specific method, I'm not sure if that's necessary. Do you mean having an additional method like:
```dart
Future<void> showWithTag(
    int id,
    String title,
    String body,
    String tag,
    NotificationDetails notificationDetails, {
    String payload,
  })
```
to reflect the two `NotificationManager` methods?
```java
void notify(int id, Notification notification)
void notify(String tag, int id, Notification notification)
```

In my view, having the tag in the platform-specific details makes more sense and is more developer-friendly. The `show` method in this plugin is not a direct proxy for `NotificationManager` anyway. 

On the other hand, we should probably have some method allowing us to call the native `NotificationManager.cancel(String tag, int id)`. Not sure what would be the best API choice for that (this could also be implemented in a separate PR).

MaikuB:
My comment on a platform-specific method was with regards to exposing a method for cancelling a notification given an id and tag. However, given what's available then should able to have the `cancel()` method of the `FlutterLocalNotificationsPlugin` and `AndroidFlutterLocalNotificationsPlugin` updated to have additional, optional, named parameter called `tag`. Will take a bit longer to look at this PR as I'm having trouble running the example app on iOS when it was working fine earlier so need to look into this. I may need to recreate the iOS side of the example app.

flutter_local_notifications/README.md
```diff
@@ -468,7 +468,7 @@ final List<ActiveNotification> activeNotifications =
 ### [Android only] Grouping notifications
```
MaikuB, line 468: The code shown below was specific to Android and the original comment around iOS was how to handle the lack of grouping support. The changes done here implies that the code also demonstrates grouping through the `threadIdentifier` such that only thing that is missing in the code below is how to show a summary notification on iOS. However, the code hasn't been updated to indicate this. Would it have been possible to update the existing area of the code in the example and this section so that the result on all platforms is as close as they can be? If not, then additional changes would be needed here, possibly even to revert the readme back to what it was before

mchudy: You're right, this has become very confusing, especially as iOS has a native notification summary feature (via categories and `categorySummaryFormat`, `summaryArgument`, `summaryArgumentCount` properties) which is not yet supported by the library. I'll split this section into iOS and Android subsections, as the mechanism is completely different for those two platforms.

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
```diff
@@ -203,8 +203,18 @@ class FlutterLocalNotificationsPlugin {
   ///
   /// This applies to notifications that have been scheduled and those that
   /// have already been presented.
-  Future<void> cancel(int id) async {
-    await FlutterLocalNotificationsPlatform.instance?.cancel(id);
+  ///
+  /// The [tag] parameter specifies the Android tag. If it is provided,
```
Albert221, line 207: Simply use backticks for the method arguments

flutter_local_notifications/README.md
```diff
@@ -35,9 +35,9 @@ A cross platform plugin for displaying local notifications.
    - [Displaying a notification](#displaying-a-notification)
    - [Scheduling a notification](#scheduling-a-notification)
    - [Periodically show a notification with a specified interval](#periodically-show-a-notification-with-a-specified-interval)
-   - [Retrieveing pending notification requests](#retrieveing-pending-notification-requests)
+   - [Retrieving pending notification requests](#retrieving-pending-notification-requests)
```
MaikuB, line 38: Thanks :)

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -244,6 +244,16 @@ class AndroidFlutterLocalNotificationsPlugin
     });
   }
 
+  @override
+  Future<void> cancel(int id, {String tag}) async {
```
MaikuB, line 248: API docs for this class haven't been updated to mention the tag (note: this class is also public). Currently it inherits the original API docs. I'd suggest for now to copy and paste the original docs and then make an addition to describe the tag. I can look at setting up macros for reusable docs to minimise copy and paste in a separate PR, as use of macros is something I've only found out about recently

MaikuB (changes requested):
Been able to run this now and looks good :)

Got one question and a requested change as the Android implementation for cancelling with id and tag hasn't had the API docs updated. In case you're not familiar with generating the API docs locally so that you can check the results, see https://flutter.dev/docs/development/packages-and-plugins/developing-packages#adding-documentation

flutter_local_notifications/example/lib/main.dart
```diff
@@ -1017,6 +1039,24 @@ class _HomePageState extends State<HomePage> {
         3, 'Attention', 'Two messages', platformChannelSpecifics);
   }
 
+  Future<void> _showNotificationsWithTag() async {
```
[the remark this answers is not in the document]
mchudy: Oh, you're right. I wanted to demonstrate the behavior of updating the notification, but the same can be achieved with just ids, so that's unnecessary.

MaikuB (approved):

PR #997, opened 2021-01-18
Extend notification channels management

VAShapovalova:
Extend management of notification channels by adding a method to get existing notification channels for android

Add method getNotificationChannels
Update example


flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -5,12 +5,14 @@ import 'package:flutter/services.dart';
 import 'package:flutter_local_notifications_platform_interface/flutter_local_notifications_platform_interface.dart';
 import 'package:timezone/timezone.dart';
 
+import '../flutter_local_notifications.dart';
```
MaikuB, line 8: Shouldn't need to import this as that's what contains references to other files. If this was needed then, it tends to be an indication that something was missed

flutter_local_notifications/lib/src/platform_specifics/android/notification_channel_output.dart
```diff
@@ -0,0 +1,17 @@
+import 'enums.dart';
+
+/// Settings for Android notification channels.
+class AndroidNotificationChannelOutputInfo {
```
MaikuB, line 4: There's already an `AndroidNotificationChannel` class for notification channels, which should be used instead to avoid confusion. The only reason I can see why another class was not this PR hasn't retrieved all of the information for the other properties. From what I can see [here](https://developer.android.com/reference/android/app/NotificationChannel), it should be possible to get the remaining info

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -1224,4 +1230,26 @@ private void getActiveNotifications(Result result) {
             result.error(GET_ACTIVE_NOTIFICATIONS_ERROR_CODE, e.getMessage(), e.getStackTrace());
         }
     }
+    private void getNotificationChannels(Result result) {
+        if (VERSION.SDK_INT < VERSION_CODES.O) {
+            result.error(GET_NOTIFICATION_CHANNELS_ERROR_CODE, GET_NOTIFICATION_CHANNELS_ERROR_MESSAGE, null);
```
MaikuB, line 1235: Could you change to be consistent with the other APIs available? Currently no exceptions are thrown for other methods around notification channels to be consistent with [NotificationManagerCompat](https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat) as throwing an exception could prevent an app from executing the logic that follows it to show a notification when it shouldn't be a show stopper.

In this case, returning an empty list would make it consistent with the native APIs (see https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#getNotificationChannels())

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -1224,4 +1230,26 @@ private void getActiveNotifications(Result result) {
...
+            return;
+        }
+        try {
+            NotificationManager notificationManager = (NotificationManager) applicationContext.getSystemService(Context.NOTIFICATION_SERVICE);
+            List<NotificationChannel>  channels = notificationManager.getNotificationChannels();
+            List<Map<String, Object>> channelsPayload = new ArrayList<>();
+
+            for (NotificationChannel channel : channels) {
+                HashMap<String, Object> channelPayload = new HashMap<>();
```
MaikuB, line 1244: Made a comment on this in the `notification_channel_output.dart` file but there's other info that could be retrieved to populate the remaining properties of the `AndroidNotificationChannel` Dart class that already exists and would remove the need to have the new class that was created in this PR. Would you be able to make changes in this PR to do this?

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -114,6 +115,8 @@
     private static final String INVALID_LED_DETAILS_ERROR_CODE = "INVALID_LED_DETAILS";
     private static final String GET_ACTIVE_NOTIFICATIONS_ERROR_CODE = "GET_ACTIVE_NOTIFICATIONS_ERROR_CODE";
     private static final String GET_ACTIVE_NOTIFICATIONS_ERROR_MESSAGE = "Android version must be 6.0 or newer to use getActiveNotifications";
+    private static final String GET_NOTIFICATION_CHANNELS_ERROR_CODE = "GET_NOTIFICATION_CHANNELS_ERROR_CODE";
+    private static final String GET_NOTIFICATION_CHANNELS_ERROR_MESSAGE = "Android version must be 8.0 or newer to use getActiveNotifications";
```
MaikuB, line 119: Error message is incorrect but see comment further below as I mention having the method return an empty list instead

MaikuB (changes requested):

MaikuB:
Thanks for making the changes. I'll take a look through and may push updates directly to the PR to try to get this merged and released as soon as possible :)

Issue #1006, opened 2021-01-22
I can't send two or more notifications because the last one is changed to another ( I wanna keep all in tray)


GilbertoGmobil:
**Describe the bug**
1º : If I send one notification, it delivers.. but if the user don't click on it and I send another, the first notification is replaced by the second

2º : I send a notification, and the user are in 'myscreen' it delivers the notification twice. I send 10 notifications, the tenth is received 10 times

**Expected behavior**
1º I want to keep all notification at the tray. If the user don't clicked at the first and receive another, I want to keep showing the two.

2º Every notification have to be delivered only one time independently of the screen opened 

**Sample code to reproduce the problem**
```dart
import 'dart:convert';

import 'package:app_md/main.dart';
import 'package:app_md/utils/app_routes.dart';
import 'package:app_md/widgets/calendario_widget.dart';
import 'package:flutter/material.dart';
import 'package:flutter_local_notifications/flutter_local_notifications.dart';
import 'package:open_file/open_file.dart';

class ReceivedNotification {
  final int id;
  final String title;
  final String body;
  final String payload;

  ReceivedNotification(
      {@required this.id,
      @required this.title,
      @required this.body,
      @required this.payload});
}

class NotificationPlugin with ChangeNotifier {
  FlutterLocalNotificationsPlugin flutterLocalNotificationsPlugin =
      FlutterLocalNotificationsPlugin();

  AndroidInitializationSettings androidInitializationSettings;
  IOSInitializationSettings iosInitializationSettings;
  InitializationSettings initializationSettings;

  NotificationPlugin() {
    initLocalNotification();
  }

  Future<dynamic> onDidReceiveLocalNotification(
      int id, String title, String body, String payload) async {
    return ReceivedNotification(
        id: id, title: title, body: body, payload: payload);
  }

  Future onSelectNotification(String payload) async {
    if(payload == "myscreen"){
      navigatorKey.currentState.pushNamed("/myscreen");
    }
  }

  Future<void> initLocalNotification() async {
    androidInitializationSettings =
        AndroidInitializationSettings('mipmap/notification_icon');//here
    iosInitializationSettings = IOSInitializationSettings(
        onDidReceiveLocalNotification: onDidReceiveLocalNotification);
    initializationSettings = InitializationSettings(
        android: androidInitializationSettings, iOS: iosInitializationSettings);
    await flutterLocalNotificationsPlugin.initialize(initializationSettings,
        onSelectNotification: onSelectNotification);
[...]

PR #1010, opened 2021-01-25
improve documentation

robert-virkus:
I was not sure about the meaning of the ID and the examples all use `0` as the ID. I found the bug report #1006 where the user also struggles about it.


MaikuB (changes requested):

MaikuB:
> the examples all use 0 as the ID

if you're referring to the example app, that was to simply things to allow the cancel example to work. Otherwise, demonstrating every functionality would require,  the user would need to specify the id for every notification and the id to cancel as well

MaikuB:
See PR comments for changes requested. Also note what was written in the PR template around contributing guidelines and PR title

robert-virkus:
Thanks for the feedback, given the issues that ask to clarify the meaning of the `id` parameter and my own struggle I do not think that the meaning is clear, hence the pull request. But the amount of documentation is of course your concern. Thanks for this great plugin!

PR #1074, opened 2021-03-06
[flutter_local_notifications_platform_interface] upgrade plugin_platform_interface to latest version 2.0.0

tatlacas:
Other common packages such as **firebase_messaging** now depend  on `plugin_platform_interface ^2.0.0`. This package will therefore be incompatible if not upgraded.

### DIFF
--- a/flutter_local_notifications/pubspec.yaml
+++ b/flutter_local_notifications/pubspec.yaml
@@ -2,13 +2,13 @@ name: flutter_local_notifications
 description: A cross platform plugin for displaying and scheduling local
   notifications for Flutter applications with the ability to customise for each
   platform.
-version: 4.0.1+2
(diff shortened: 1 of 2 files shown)


MaikuB:
Thanks for the PR. The reason why the other plugins are using `plugin_platform_interface` 2.0 is that they've been migrated to null safety, which requires using version 2.0+  of `plugin_platform_interface`. There's already a null safety release of this plugin (https://pub.dev/packages/flutter_local_notifications/versions/5.0.0-nullsafety.0) that is in prerelease as the migration was done recently. What's been done in this PR would also fail for those using older versions of the Flutter SDK. This can be confirmed by getting the last stable release (1.22.6) and applying the changes done in this PR. As such I'll be closing/not merging this

PR #1081, opened 2021-03-09
[flutter_local_notifications] Fix the getNotificationAppLaunchDetails() example in the README

rocboronat:
The code in the example didn't compile. I played around until it worked :·) And here it is!

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -551,7 +551,7 @@ await flutterLocalNotificationsPlugin.cancelAll();
 
 ```dart
 final NotificationAppLaunchDetails notificationAppLaunchDetails =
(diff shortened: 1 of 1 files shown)


MaikuB:
Thanks for the PR. The readme was meant to be read more in sequence so if you do the initialisation steps that were described as the first step, this would've created an instance of the plugin with the same variable name. I only cover this once and mention this as the first step to avoid duplicating the initialisation code in every single code snippet that would otherwise make the readme even more longer. I would guess that based on your changes and issue you've described that you copied only the portion of the code for getting the launch details.

rocboronat:
Understood. Then, I close the PR. Thanks for your time, @MaikuB ! 👏

PR #1090, opened 2021-03-13
flutter_local_notifications package zonedSchedule method multiple notifications

onatcipli:
Hey there this package helps me to create some local notifications for reminders but since the app that I'm using schedules notifications like every n day (n depending on the user )

so for example, if the user schedule notification every 1 day for one month, if I was using the `DateTimeComponents.dayOfWeekAndTime` I'm getting the notification 4 times on every scheduled this is because of its only checks the **week and time** if I was using the `DateTimeComponents.time`, I'm getting the notification 30 times because this time its only checks time so that I think we need a more clear match in here so I was thinking adding a `DateTimeComponents.dayOfMonthAndTime` with the help of that we should be fine under 1-year schedules.

I did not have enough ios knowledge maybe you can help me with the ios side after that we can test this whole thing if it looks fine for you

### DIFF
--- a/flutter_local_notifications/CHANGELOG.md
+++ b/flutter_local_notifications/CHANGELOG.md
@@ -1,3 +1,7 @@
+# [4.0.1+3]
+
+* [android/iOS] DateTimeComponents dayOfMonthAndTime added.
+
 # [4.0.1+2]
 
 * [iOS/macOS] fixed issue where not requesting any permissions (i.e. all the boolean flags were set to false) would still cause a permissions prompt to appear. Thanks to the PR from [Andrey Parvatkin](https://github.com/badver)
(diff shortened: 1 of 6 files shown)


onatcipli:
@MaikuB  Fyi, the ios part is not completely finished

MaikuB:
I'm having some trouble understanding what you've written but it sounds like you essentially want monthly notifications. This isn't supported as everyone's use case for this is different and would most likely need to be custom code to solve. The approach you seem to be taking has issues as well. For example,  say that a notification should occur on the 31st day of the month. You'll get a notification on the 31st of March but the next notification would be 31st of May. Perhaps this would be fine for your application but from what I remember seeing from experimenting with iOS before, iOS would have a notification on the 30th of April as well. This would be different behaviour to what you've got on Android. It's not clear, at least to me, what the rules are in how iOS decides to handle these boundary cases. This is another why the plugin doesn't support this.

So to summarise, you'll most likely need to come up with something specific to your app.

PR #1097, opened 2021-03-20
[flutter_local_notifications] Add copyWith Methods to Platform NotificationDetails

aschulz90:
This pull request allows for re-using of platform notification details among different notifcation use cases.

e.g.:
```

const AndroidNotificationDetails defaultAndroidNotiDetails = AndroidNotificationDetails(
  'Channel ID',
  ''Channel Name',
  'Channel Description',
  importance: Importance.max,
  priority: Priority.high,
  showWhen: false,
);

FlutterLocalNotificationsPlugin().show(
              1,
              'First Group Title',
              'Second Group Body',
              NotificationDetails(android: defaultAndroidNotiDetails.copyWith(
                    groupKey: 'First Group'
              )),
);

FlutterLocalNotificationsPlugin().show(
              2,
              'Second Group Title',
              'Second Group Body',
              NotificationDetails(android: defaultAndroidNotiDetails.copyWith(
                    groupKey: 'Second Group',
                    showWhen: true,
                    when: DateTime.now().millisecondsSinceEpoch
              )),
);

```


MaikuB:
Thanks for the PR. Whilst this may be useful to some developers, I'm not looking to have this method added as it's not a core API. These are more like extensions so I would suggest that you publish this as another package that others may decide to use via the use of extension methods

proninyaroslav:
@MaikuB 
Things like this are part of the Flutter API, for example in the `Theme` class. Why not?

PR #1112, opened 2021-04-05
[flutter_local_notifications] updated timezone dependency

MaikuB:


arslantalha:
Hi, as I am new I am not sure if I am at the right place but whenever I am trying to install flutter_local_notifications in the pubspec.yaml i get the error,  "flutter_local_notifications >=4.0.0-dev.2 <5.0.0-nullsafety.0 depends on timezone ^0.6.1," 
Hence I am wondering if I'm doing something wrong as you have mentioned it as updated in the merged post. 

kind regards

Talha

MaikuB:
What version of the plugin are you using? The logs would seem to indicate that you're using a prerelease. Check your app and pub to make sure you're on the latest version. The changelog will have info on what updates were made in each version. You can also subscribe to when releases published on the GitHub repository via the watch option

arslantalha:
Thanks for the Feedback Maiku. 

After you comment I checked my pubspec yaml. i was set to 
     flutter_local_notifications: ^5.0.0+1
after changing it to 
     flutter_local_notifications: ^5.0.0
it worked out somehow without anyway.

PR #1114, opened 2021-04-07
[flutter_local_notifications] fix for nil value in notification.request.content.userInfo

badver:
Sometimes there is nil value in notification.request.content.userInfo[...]
Just add checking for nil in if statement.

### DIFF
--- a/flutter_local_notifications/macos/Classes/FlutterLocalNotificationsPlugin.swift
+++ b/flutter_local_notifications/macos/Classes/FlutterLocalNotificationsPlugin.swift
@@ -85,16 +85,16 @@ public class FlutterLocalNotificationsPlugin: NSObject, FlutterPlugin, UNUserNot
     @available(OSX 10.14, *)
     public func userNotificationCenter(_ center: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
         var options:UNNotificationPresentationOptions = []
-        let presentAlert: Bool = (notification.request.content.userInfo[MethodCallArguments.presentAlert] as? Bool) ?? false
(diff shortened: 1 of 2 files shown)


MaikuB:
Do you have examples on how the scenario could happen? Trying to understand the root cause as I haven't seen any reports on this and I would've thought all notifications created from this plugin when run on macOS 10.14+ would have these values due to the code at https://github.com/MaikuB/flutter_local_notifications/blob/c0087bce1f8fedd04de2c8a3cd1d259315242155/flutter_local_notifications/macos/Classes/FlutterLocalNotificationsPlugin.swift#L386

badver:
It happens often when a desktop app is minimized on macOS 10.15.7 and a user has clicked on a notification from FCM.

MaikuB:
That sounds odd as I thought userinfo is populated and the code for the fcm plugin would would indicate that it is too https://github.com/FirebaseExtended/flutterfire/blob/44c410c4cb0f01d173c348e835a6a497416d174b/packages/firebase_messaging/firebase_messaging/ios/Classes/FLTFirebaseMessagingPlugin.m#L279

If it isn't then the bug should exist in the fcm plugin too. Based on this, I suspect the issue lies elsewhere. It would be great if you're able to provide a link to a repo with a minimal app that can two-sided this.

The other thing is the changes in this PR would end up having the plugin process notifications coming from other plugins. This should be avoided as it would prevent other plugins from doing so. Some of them, like fcm, rely on this to trigger callbacks that would be received in the Flutter/Dart side. The iOS implementation of the plugin does have some code to do this but I hadn't brought it over to the macOS implementation as last I saw, the Flutter engine needed changes on macOS to allow for multiple notification plugins to work together like it does on iOS.

badver:
Sorry for the delay. I cannot reproduce this problem anymore.
I think it's related to firebase_messaging, not your package, so let's close this as unreproducible.

PR #1131, opened 2021-04-18
[flutter_local_notifications] Add new DateTimeComponents (DayOfMonthAndTime, DateAndTime)

DenisShakinov:
I guess it's connected somehow with [https://github.com/MaikuB/flutter_local_notifications/issues/92](https://github.com/MaikuB/flutter_local_notifications/issues/92). Also it seems iOS 8-9 isn't supported since it does not have calendar triggers.


MaikuB:
Thanks for the PR. I'll need to keep this open for a while to test. From I remember when I looked at this a long time ago and why this wasn't supported was that iOS handled boundary conditions differently. I recall that if a notification was set to match for 31st day of the month (e.g. starts on January), there'd still be a notification next month even if there isn't 31 days in the next month.

MaikuB:
Thanks a lot for this :)

PR #1205, opened 2021-06-23
Pre-work to support Flutter web

cylyl:


MaikuB:
Haven't had a chance to look in detail yet but looks like the PR is causing existing tests to fail. The platform interface is meant to define the common interface that a _platform-specific implementation_  is supposed to implement. The PR has moved all the platform-specific types into the platform interface and now forces all platform-specific implementations to know about the types of other platforms. This plugin isn't like some of the existing Flutter plugins where the platform interface package has a generic method channel implementation as there's a lot more platform-specific behaviour so the changes with moving types would need to be reverted

MaikuB:
This PR has been open for a while now with no response from the original author. As such I'll be closing this due to lack of engagement and for all I know, this PR could've been opened by mistake

PR #1232, opened 2021-07-15
[flutter_local_notifications] Update build.gradle - swap "android-S" to 31

tim-mardesen:
Instead of specifying the compileSdkVersion by the preview name, it is best to use the actual API number, and android-S's version number is 31.

By difference, some machines cannot fetch version "android-S", as it is a preview SDK build, whereas 31 is a non-preview build, and can be fetched by all machines.

### DIFF
--- a/flutter_local_notifications/android/build.gradle
+++ b/flutter_local_notifications/android/build.gradle
@@ -22,7 +22,7 @@ rootProject.allprojects {
 apply plugin: 'com.android.library'
 
 android {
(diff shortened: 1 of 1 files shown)


tim-mardesen:
@MaikuB as this is an easy change, it might be better to close this PR and perform this change in your android_12_support branch. 

Is there an ETA when a new version of this package will hit pub.dev? @Phantom0001

MaikuB:
Thanks for the PR. Will get this out soon

Issue #1247, opened 2021-07-28
flutter_local_notifications 6.1.0 got error when running in flutter web.


phonghoang2k:
**Describe the bug**
When update plugin into version 6.1.0, this error is returned when running in flutter web (I have disabled the related flutter_local_notifications's code using !kIsWeb for only run in mobile device).
```
../../.pub-cache/hosted/pub.dartlang.org/ffi-1.1.2/lib/src/utf16.dart:5:8: Error: Not found: 'dart:ffi'
import 'dart:ffi';
       ^
../../.pub-cache/hosted/pub.dartlang.org/posix-2.1.2/lib/src/unistd/unistd.dart:209:15: Error: Type 'ffi.Int32' not found.
  ffi.Pointer<ffi.Int32> __pipedes,
              ^^^^^^^^^
../../.pub-cache/hosted/pub.dartlang.org/posix-2.1.2/lib/src/unistd/unistd.dart:209:3: Error: Type 'ffi.Pointer' not found.
  ffi.Pointer<ffi.Int32> __pipedes,
  ^^^^^^^^^^^
../../.pub-cache/hosted/pub.dartlang.org/posix-2.1.2/lib/src/unistd/unistd.dart:209:3: Error: Expected 0 type arguments.
  ffi.Pointer<ffi.Int32> __pipedes,
...
```
**To Reproduce**

1. Upgrade the plugin to latest version 6.1.0.
2. Runs app on Flutter web, you will see error return.

**Expected behavior**
the app runs successfully and not encountered error.

**Sample code to reproduce the problem**
The bug only happens at the build time and not related to any particular piece of code. 
Checking the log, maybe there's somethings wrong when using the posix-2.1.2 lib because when i downgraded the plugin to version 5.0.0+4, everything run perfectly with no error.

raviganwal:
Yes I just update to` flutter_local_notifications: ^7.0.0-dev.3 ` and I am facing same error. while my app was working fine with `^6.0.0`.

nori-flutter:
I got the same error. Everything was okay with version 6.0.0 for web build.

MaikuB:
Thanks for the info. Yes it looks like adding Linux support has caused an issue with this.

[...]

PR #1249, opened 2021-07-28
[flutter_local_notifications_linux] Fix build failing on Web

proninyaroslav:
Web doesn't support FFI, so we need to strip all transitive dependencies at compile time, that using FFI
Fixed https://github.com/MaikuB/flutter_local_notifications/issues/1247


proninyaroslav:
@MaikuB 
I made a compile-time stub for the Linux plugin, I also removed the runtime checking for the integer width, because this duplicates the functionality of the `dbus` package. Now building for the Web happens without any problems.

MaikuB:
Thank.  To carry on the convo from the issue thread, I probably wasn't clear enough that I also did a conditional import. The difference though is that the stub had empty implementations for all methods. I won't be able to test this PR until later but I would thought that would be needed as difference between this plugin and the connectivity one is that this plugin has platform-specific methods and the overrides take extra platform-specific parameters. Therefore, I would've thought an app is compiling for the web but makes calls directly to the Linux plugin (e.g. https://pub.dev/documentation/flutter_local_notifications_linux/latest/flutter_local_notifications_linux/LinuxFlutterLocalNotificationsPlugin/initialize.html or https://pub.dev/documentation/flutter_local_notifications_linux/latest/flutter_local_notifications_linux/LinuxFlutterLocalNotificationsPlugin/getSystemIdMap.html), it would fail to compile as these won't exist in the stub.

Another thing is I think it'd still be good to re-implement the posix calls as the `posix` package depends on ffi versions >= 0.1.3 but less < 1.0.0 to avoid potential problems around dependency conflicts with what's baked into the Flutter SDK. There are very few calls to implement that I'd say it's worth doing to remove `posix` as a dependency

proninyaroslav:
> it would fail to compile as these won't exist in the stub.
Yes, you are right, thanks for the comment. I forgot that the main plugin implementation calls these methods in code. I moved these methods into a separate interface that extends `FlutterLocalNotificationsPlatform`, and the real implementation and the stub implement it.
In my local repo, I did a revert that disables Linux support and checked the build of the example app for Linux and Web. I have not noticed any problems.

proninyaroslav:
> Another thing is I think it'd still be good to re-implement the posix calls

Yes, I'll push it today.

proninyaroslav:
@MaikuB 
POSIX calls: d4ec7453d3a6650ec358bfa1b62f7a002bea7b17
Method stubs:  e69b18d2a66be8b564ccc43be56d6e15cb18586a

flutter_local_notifications_linux/lib/flutter_local_notifications_linux.dart
```diff
@@ -1,7 +1,12 @@
 /// The Linux implementation of `flutter_local_notifications`.
 library flutter_local_notifications_linux;
 
-export 'src/flutter_local_notifications.dart';
+// flutter_local_notifications_linux depends on dbus and posix
+// which uses FFI internally; export a stub for platforms that don't
+// support FFI (e.g., web) to avoid having transitive dependencies
+// break web compilation.
+export 'src/flutter_local_notifications_stub.dart'
+    if (dart.library.ffi) 'src/flutter_local_notifications.dart';
```
MaikuB, line 9: I know this logic is based on what's in the community plus plugins but given the comments and intent, perhaps the logic could be better expressed. The code reads like it should load the real implementation for platform thats support FFI but the intent is to load a stub for the web so apps can compile when targeting web. This could be done by kind of inverting the logic

```
export 'src/flutter_local_notifications.dart'
     if (dart.library.js) 'src/flutter_local_notifications_stub.dart';
```

reference: https://github.com/flutter/cocoon/blob/eb8d951bf547256785229538efc47d3459906155/app_flutter/lib/service/downloader.dart#L5

MaikuB (commented):
Left some comments on the PR. Let me know what you think

flutter_local_notifications_linux/lib/flutter_local_notifications_linux.dart
MaikuB: Hmm actually ignore my comment. From a generic POV, you'd want to see if the APIs you use are actually supported

flutter_local_notifications_linux/lib/src/ffi/libc.dart
```diff
@@ -0,0 +1,25 @@
+import 'dart:ffi' as ffi;
+
+import 'dart:io' show Platform;
+
+/// FFI binding for C library
+class Libc {
+  /// Constructs an instance of [Libc].
+  factory Libc() => _libc;
+
+  Libc._() {
```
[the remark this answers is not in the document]
proninyaroslav: I reduced the boilerplate, but I didn't combine this with `LinuxPlatformInfo`, as this would make FFI calls less testable . Currently POSIX is part of the tests and formally we check the FFI bindings correctness.

flutter_local_notifications_linux/lib/flutter_local_notifications_linux.dart
proninyaroslav: It really doesn't make much of a difference in the case of the Web: it will work in both cases. But if the Web ever gets FFI support, or if there is another platform that doesn't support FFI, my export option will make more sense, because it's not just limited to the Web, but any platform that doesn't support FFI.
I can change this now if you wish.

PR #1262, opened 2021-08-05
[flutter_local_notifications_linux] Using TMPDIR env as a primary temporary path

proninyaroslav:
TMPDIR is a standard variable on UNIX/Linux systems, and is often used in containers such as Flatpak to redirect to a temporary folder inside a sandbox. This allows not to make hard bindings to the /tmp directory

### DIFF
--- a/flutter_local_notifications_linux/lib/src/platform_info.dart
+++ b/flutter_local_notifications_linux/lib/src/platform_info.dart
@@ -27,8 +27,15 @@ class LinuxPlatformInfo {
         final int pid = _posix.getpid();
         final int userId = _posix.getuid();
         final int sessionId = _posix.getsid(pid);
+        final Map<String, String> env = Platform.environment;
+        final String? tmpdir = env['TMPDIR'];
         runtimeDir = Directory(
-          path.join('/tmp', processName, '$userId', '$sessionId'),
+          path.join(
+            tmpdir == null || tmpdir.isEmpty ? '/tmp' : tmpdir,
+            processName,
+            '$userId',
+            '$sessionId',
+          ),
         );
       } else {
         runtimeDir = Directory(path.join(xdg.runtimeDir!.path, processName));


MaikuB:
Just had a thought on this and that I reckon this only be should be saved to the same location the `shared_preferences_linux` plugin saves to. This is based on where the application is located. It would align with what the ecosystem is doing to save application-related data. Here are the links for reference https://github.com/flutter/plugins/blob/60100908443fffee28df82735a9c4019cf549227/packages/shared_preferences/shared_preferences_linux/lib/shared_preferences_linux.dart#L37, which in turn calls https://github.com/flutter/plugins/blob/60100908443fffee28df82735a9c4019cf549227/packages/path_provider/path_provider_linux/lib/path_provider_linux.dart#L26

Let me know what you think

proninyaroslav:
The mapping of the plugin ID to the notification ID is currently a temporary file, because if you log out and log in again, the notification IDs will be counted back from 1. Therefore, I use two options for storing this file: `XDG_RUNTIME_DIR` or `/tmp`, since `XDG_RUNTIME_DIR` can be not available on some distro. Using a hard link to `/tmp` is also not a good idea, because, as I said earlier, the path to `/tmp` can be overridden by the user/system/app container.

proninyaroslav:
`XDG_DATA_HOME` directory that used by `shared_preferences` does not apply for temporary files.

MaikuB:
Ah yes, I missed that notifications would be cleared etc upon a new session so a temporary location is preferred. Thanks.

flutter_local_notifications_linux/lib/src/platform_info.dart
```diff
@@ -27,8 +27,15 @@ class LinuxPlatformInfo {
         final int pid = _posix.getpid();
         final int userId = _posix.getuid();
         final int sessionId = _posix.getsid(pid);
+        final Map<String, String> env = Platform.environment;
+        final String? tmpdir = env['TMPDIR'];
         runtimeDir = Directory(
```
MaikuB, line 32: @proninyaroslav sorry I missed this before but given this change has been made, this would require the API docs for `runtimePath` to be updated as well right?

proninyaroslav: Yes, thank you.

PR #1269, opened 2021-08-06
[flutter_local_notifications]: fix typos

Ascenio:
Fix small typos related to the term `initialization`.


MaikuB:
Thanks for the PR but I'm not looking to merge this in. These are not mistakes unless you want that a lot of countries are spelling words incorrectly. `Initialize` is the American English spelling and there are plenty of other countries (e.g. Australia, which is where I live) using British English. The former was used only for the API because a lot of APIs are using American English (e.g. a `Color` class, which is spelt as `Colour` in British English)

PR #1274, opened 2021-08-08
[flutter_local_notifications_linux] Replace 'developer.gnome.org' links to 'specifications.freedesktop.org'

proninyaroslav:
GNOME refresh the developer portal and removed some of the old links.

### DIFF
--- a/flutter_local_notifications_linux/lib/src/model/initialization_settings.dart
+++ b/flutter_local_notifications_linux/lib/src/model/initialization_settings.dart
@@ -21,6 +21,7 @@ class LinuxInitializationSettings {
   final LinuxNotificationIcon? defaultIcon;
 
   /// Specifies the default sound for notifications.
+  /// Typical value is `ThemeLinuxSound('message')`
   final LinuxNotificationSound? defaultSound;
 
   /// Causes the server to suppress playing any sounds, if it has that ability.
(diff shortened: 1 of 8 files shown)


flutter_local_notifications_linux/lib/src/model/initialization_settings.dart
```diff
@@ -21,6 +21,7 @@ class LinuxInitializationSettings {
   final LinuxNotificationIcon? defaultIcon;
 
   /// Specifies the default sound for notifications.
+  /// Typical value is `ThemeLinuxSound('message')`
```
proninyaroslav, line 24: I thought it would be better to mention this in the documentation, rather than make it as the default value. How do you think?

MaikuB: I don't use Linux enough to comment with great confidence but from what I could tell, this seems to make sense. Not sure if it's an issue with my Ubuntu VM but I did notice the example app isn't playing the default alert sound when showing a basic notification. The custom sound example wasn't working for me either. Are these working for you?

proninyaroslav: @MaikuB 
As I found out, no sound is played by default.  You need to set your own sound or use a theme sound, for example the standard `ThemeLinuxSound ('message')` notification sound.  For this reason, I decided to add a comment.

PR #1297, opened 2021-08-30
[flutter_local_notifications] Increase timezone version dependency

xvrh:
[`timezone`](https://pub.dev/packages/timezone) has a new version. Would be great to support it.


MaikuB:
Thanks a lot for this :)

PR #1268, opened 2021-08-06
[flutter_local_notifications] Android 12 support, Linux support and platform detection rework

MaikuB:
Closes #1306 #1171 #1324 

Note: these changes have been published in a pre-release as Android 12 is in beta. For now, this won't be merged in and published as a stable release until Android 12 itself is published as stable

PR #1299, opened 2021-09-02
[flutter_local_notifications] Android 12 Fix

abahnj:
This adds the required exported flag required for broadcast receivers on android 12


MaikuB:
An existing PR and prerelease exists for this https://github.com/MaikuB/flutter_local_notifications/pull/1268

MaikuB:
Closing as an existing PR and prerelease exists. Note that value for exported flag should've been false instead of true as intent is that only this plugin should be able to fire the receivers

PR #1227, opened 2021-07-11
[flutter_local_notifications]: allow to set largeIcon from base64 encoded string and byte array

lexxxel:
Basically, this allows to set the large icon from flutter directly, e.g. when the image is received from a firebase cloud message.

PR #1316, opened 2021-09-21
[flutter_local_notifications]: allow byte arrays as source for icons

lexxxel:
Implements the same change that has been done in #1227 but for icons (e.g. used for `PersonDetails`)

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -492,7 +492,8 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
         return context.getResources().getIdentifier(name, DRAWABLE, context.getPackageName());
     }
 
-    private static byte[] objectDataToByteArray(Object data) {
+    @SuppressWarnings("unchecked")
+    private static byte[] castObjectToByteArray(Object data) {
         byte[] byteArray;
         // if data is deserialized by gson, it is of the wrong type and we have to convert it
         if (data instanceof ArrayList) {
@@ -507,7 +508,6 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
         return byteArray;
     }
 
-    @SuppressWarnings("unchecked")
     private static Bitmap getBitmapFromSource(Context context, Object data, BitmapSource bitmapSource) {
         Bitmap bitmap = null;
         if (bitmapSource == BitmapSource.DrawableResource) {
@@ -515,7 +515,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
         } else if (bitmapSource == BitmapSource.FilePath) {
             bitmap = BitmapFactory.decodeFile((String) data);
         } else if (bitmapSource == BitmapSource.ByteArray) {
-            byte[] byteArray = objectDataToByteArray(data);
+            byte[] byteArray = castObjectToByteArray(data);
             bitmap = BitmapFactory.decodeByteArray(byteArray, 0, byteArray.length);
         }
 
@@ -547,7 +547,7 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
                 }
                 break;
             case ByteArray:
-                byte[] byteArray = objectDataToByteArray(data);
+                byte[] byteArray = castObjectToByteArray(data);
                 icon = IconCompat.createWithData(byteArray, 0, byteArray.length);
             default:
                 break;
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -517,22 +517,22 @@ public class FlutterLocalNotificationsPlugin implements MethodCallHandler, Plugi
         return bitmap;
     }
 
-    private static IconCompat getIconFromSource(Context context, String iconPath, IconSource iconSource) {
+    private static IconCompat getIconFromSource(Context context, Object data, IconSource iconSource) {
         IconCompat icon = null;
         switch (iconSource) {
             case DrawableResource:
-                icon = IconCompat.createWithResource(context, getDrawableResourceId(context, iconPath));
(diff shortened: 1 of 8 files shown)


lexxxel:
@MaikuB looks like you are busy right know. Do you think you could be able to review this PR in the next 1 or 2 weeks?

MaikuB:
Yeah got a lot of things happening at work. No promises but I think I could review during that timeframe. If the reason you're asking is so you can make use of it then note you could have your app point to your fork via a Git dependency. From a quick glance I took, there's no example making use of the changes in the PR. Would you be able to update the example app to include one?

lexxxel:
Of course I will add an example. I had an issue with pointing to a fork, because one has to point to a subdirectory, too. Otherwise an error occurred saying the pubspec is missing in the root of the git repo. Do you know how I can change the 'root' dir in the pubspec?

MaikuB:
You can specify the `path`. See https://dart.dev/tools/pub/dependencies#git-packages

MaikuB:
Forgot to add that I'll review once the example app has been updated as it makes it easier to verify the functionality

lexxxel:
> Forgot to add that I'll review once the example app has been updated as it makes it easier to verify the functionality

No, that was clear to me 😄. I added some code to the example to show the usage.

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -521,6 +521,19 @@ private static IconCompat getIconFromSource(Context context, String iconPath, Ic
                     throw new RuntimeException(e);
                 }
                 break;
+            case ByteArray:
+                byte[] byteArray;
```
MaikuB, line 525: Would you be able to refactor the logic around assigning the value of `byteArray`? It's a repeat of what was in `getBitmapFromSource`.

flutter_local_notifications/example/lib/main.dart
```diff
@@ -1074,6 +1074,13 @@ class _HomePageState extends State<HomePage> {
       bot: true,
       icon: BitmapFilePathAndroidIcon(largeIconPath),
     );
+    final Person chef = Person(
+        name: 'Master Chef',
+        key: '3',
+        uri: 'tel:111222333444',
+        icon: ByteArrayAndroidIcon.fromBase64String(
+            await _base64encodedImage('http://placekitten.com/48/48')));
```
MaikuB, line 1082: Can this be made `https` instead of `http` as I recall that using `http` could lead to issues

flutter_local_notifications/lib/src/platform_specifics/android/icon.dart
```diff
@@ -1,12 +1,20 @@
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'enums.dart';
+
 /// Represents an icon on Android.
-abstract class AndroidIcon {
+abstract class AndroidIcon<T> {
```
MaikuB, line 7: Thanks, these changes make sense and makes it consistent with how bitmaps are handled. They are however, breaking changes and with Android 12 release is around the corner. Do you think you could try to make this PR target the 9.0.0 branch instead so I can try to bundle it with the 9.0.0 release?

MaikuB (changes requested):
Thanks for taking the time to do this. I've left some comments in the PR, the one I wanted to call out here in case it gets missed is to could you try to make this PR could target 9.0.0 branch? It currently has an open [PR](https://github.com/MaikuB/flutter_local_notifications/pull/1268) for Android 12 support etc. As there are breaking changes in the PR you submitted, it would've required a major version bump and would like to try to bundle this with that release. This would avoid having to a 9.0.0 and then a 10.0.0 release immediately after and I was looking to release 9.0.0 once Android 12 has rolled out. This is supposedly right around the corner

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
lexxxel: done

flutter_local_notifications/example/lib/main.dart
lexxxel: done

flutter_local_notifications/lib/src/platform_specifics/android/icon.dart
lexxxel: rebased and changed target

lexxxel: Is this what you wanted?

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -492,6 +492,21 @@ private static int getDrawableResourceId(Context context, String name) {
         return context.getResources().getIdentifier(name, DRAWABLE, context.getPackageName());
     }
 
+    private static byte[] objectDataToByteArray(Object data) {
```
MaikuB, line 495: Can you apply the `@SuppressWarnings("unchecked")` annotation here? Not having the annotation here would result in warnings for all consumers of this library around unsafe operations. This should also allow for removing this annotation from the `getBitmapFromSource` method.

Regarding the name of the method, could this be something like `castObjectToByteArray`? Reason I'm proposing this is to improve readability since methods are actions and the name is lacking a verb to describe the action

flutter_local_notifications/lib/src/platform_specifics/android/icon.dart
MaikuB: Yep

MaikuB: Thanks :)

MaikuB (changes requested):
Thanks for the quick turnaround. Left a comment on the Java side of the code to address. Rest looks good to me :)

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
lexxxel: No problem, I applied all of your suggestions - Java and it's convention are not my strong suit.

MaikuB (approved):

lexxxel:
🙇  Thank you very much

PR #1351, opened 2021-10-22
[flutter_local_notifications] Schedule multiple notifications with different data

mojtabaghiasi:
flutter channel process cost in android can be high, so we can schedule multiple notifications with one channel method call.
we can add a list of notification data including id, title, body,  scheduled date, notification details, and more. with this method, w register all notifications just with one channel method call.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -87,7 +87,7 @@ import io.flutter.plugin.common.PluginRegistry;
  * FlutterLocalNotificationsPlugin
  */
 @Keep
-public class FlutterLocalNotificationsPlugin implements MethodCallHandler, PluginRegistry.NewIntentListener, FlutterPlugin, ActivityAware {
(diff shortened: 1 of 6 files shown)


mojtabaghiasi:
@MaikuB Could you check this pull request?

MaikuB:
Will do but note that this was something requested before and I had decided not to have in the plugin to avoid bloating the core APIs. What kind of performance improvements are you experiencing now relative to scheduling them by iterating to make a schedule each notification? From what I saw, the Flutter team had been working on improving performance of using platform channels too

mojtabaghiasi:
@MaikuB When we want to schedule multi notifications (e.g 50 scheduled notifications), we should schedule by iterating to schedule each of them, when we use this method of scheduling multiple notifications, we can see UI lag and UI freeze on the main thread, I saw this problem many times. you can test this situation with some weak performance devices and observe this problem.

MaikuB:
I don't have access to such devices to check. What you're describing seems to be related to a general problem that can occur with Flutter applications and isn't specific to the plugin. There's an example shown [here](https://flutter.dev/docs/cookbook/networking/background-parsing) around how this can occur with parsing JSON and how it can be dealt with using isolates. I've heard from others in the community that they have been able to use isolates with this plugin for the scenario where multiple notifications need to be scheduled too. Given it's a general issue then a general solution would be better and be more scalable. Have you tried using this approach?

abdimussa87:
I've tried running this in an isolate, but for some reason the UI still freezes

msxenon:
the same issue here

PR #1363, opened 2021-10-30
[flutter_local_notifications] Make manifest compatible with SDK 31+

luca-colazzo:
To make an app installable to Android 12+ you need to specify that property


MaikuB:
The property is already there that this PR introduces a duplicate value and causes builds to fail. Therefore I'll be closing this PR

Issue #1355, opened 2021-10-25
Feature Request: Allow setting `defaultActionName` on Linux to null


karmanyaahm:
And consequently return an empty array here if it is null:

https://github.com/MaikuB/flutter_local_notifications/blob/03680366b0c3c232ce74a002fdadaac74168462d/flutter_local_notifications_linux/lib/src/notifications_manager.dart#L206-L214

This would be very helpful for notifications that shouldn't be clickable, since clicking them doesn't do anything. Having notifications that don't trigger an action being clickable just confuses the user.

I can submit a pull request for this if it would be accepted?

MaikuB:
Closing this since the PR was closed and the results differ on different distributions

PR #1367, opened 2021-10-30
[flutter_local_notifications_linux] Allow disabling Linux default action

karmanyaahm:
The following commits are part of this PR:
- [flutter_local_notifications_linux] Allow disabling the default action (#1355)
   - make `LinuxInitializationSettings.defaultActionName` nullable to indicate removing the default action
   - add bool `LinuxNotificationDetails.enableDefaultAction` (default true) to indicate whether one specific notification should have the default action 
   - update LinuxNotificationManager._buildActions to account for the above
- [flutter_local_notifications_linux] Add tests for actions
  - change `buildNotifyMethodValues` to use passed in actions OR default actions if null
  - this doesn't affect any other tests since they don't use actions
  - add tests for actions
- [flutter_local_notifications] add non-clickable linux-specific example
  - this uses only the `LinuxNotificationDetails.enableDefaultAction` way of disabling actions
      rather than the Initializer way
- 5fd2236c is a change necessary for testing. IDK if it should be kept or not?

### DIFF
--- a/flutter_local_notifications/example/lib/main.dart
+++ b/flutter_local_notifications/example/lib/main.dart
@@ -787,6 +787,12 @@ class _HomePageState extends State<HomePage> {
                           await _showLinuxNotificationDifferentLocation();
                         },
                       ),
+                      PaddedElevatedButton(
+                        buttonText: 'Show non-clickable notification',
+                        onPressed: () async {
+                          await _showLinuxNotificationNonClickable();
+                        },
+                      )
                     ],
                   ],
                 ),
@@ -2232,6 +2238,20 @@ Future<void> _showLinuxNotificationDifferentLocation() async {
   await flutterLocalNotificationsPlugin.show(
     0,
     'notification on different screen location',
+    null,
+    platformChannelSpecifics,
+  );
+}
+
+Future<void> _showLinuxNotificationNonClickable() async {
+  const LinuxNotificationDetails linuxPlatformChannelSpecifics =
+      LinuxNotificationDetails(enableDefaultAction: false);
+  const NotificationDetails platformChannelSpecifics = NotificationDetails(
+    linux: linuxPlatformChannelSpecifics,
+  );
+  await flutterLocalNotificationsPlugin.show(
+    0,
+    'notification that is not clickable',
     null,
     platformChannelSpecifics,
   );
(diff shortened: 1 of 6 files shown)


MaikuB:
Thanks for the PR. I'll do so some testing and will merge if it all looks good

MaikuB:
Took a look at this. Could you confirm my understanding of what the PR is trying to achieve? The PR and related issue states this is about allowing for notifications to not be clickable. Is this actually true and/or accurate to what it's about? From what I've observed the notification is still clickable. It looks more like what you want is the ability for a notification to not cause something to happen when its clicked. Is this correct?

If so, then I would this PR actually cause more confusion and wouldn't be something that I'd be looking to merge in. The reason for this is generally speaking the user has actually clicked on the notification. From what there, it falls into your application to decide what should happen based on the notification that was clicked. This sequence of events would technically map to a callback and then application logic to analyse the notification's details to see if certain logic would run. This is how it works on other platforms too. The changes in this PR doesn't align with the sequence of events, would introduce inconsistency with other platforms and allowing a nullable action could add more confusion when a user action actually took place.

Hope that made sense. To achieve what I understand you're trying to do, you could make use for the notification payload to add extra data in that is read in the callback to decide what should happen.

The change in https://github.com/MaikuB/flutter_local_notifications/commit/5fd2236cef2f5a0853f48f1a812ef9b84820035b would be useful to have though

karmanyaahm:
> Is this actually true and/or accurate to what it's about? From what I've observed the notification is still clickable.

On KDE Plasma 5.23.0 the notification is not clickable. See the following images for mouse pointer differences:

image
image

> If so, then I would this PR actually cause more confusion and wouldn't be something that I'd be looking to merge in.

If clickable behavior is inconsistent across desktop environments, then I totally agree that it would increase confusion by letting users click on notifications that are just supposed to be informational, and not trigger any action. 

Though, I'm unsure why any distribution/desktop environment would do that? Which desktop environment are you using for this test?

MaikuB:
Ubuntu 20.04 via a VM that I run using Parallels

karmanyaahm:
Hmmm... I observe the same behavior in Ubuntu 21.10. I'll see if there's any solution to this during the weekend. If not, i'll close this PR.

karmanyaahm:
It seems like the Gnome design is intentionally like that to be in line with their design philosophy. I'm sure other desktop environments also have different sets of behaviors like this, so it's probably not worth it to merge this PR.

@MaikuB Thanks for your time. Hopefully this new knowledge will help me make better design decisions for my app in general.

MaikuB:
No worries, thanks for the info on this too.

PR #1376, opened 2021-11-04
[flutter_local_notifications] Add groupKey field to ActiveNotification

drstranges:


flutter_local_notifications/example/lib/main.dart
```diff
@@ -1960,6 +1960,7 @@ class _HomePageState extends State<HomePage> {
                 children: <Widget>[
                   Text(
                     'id: ${activeNotification.id}\n'
+                    'group: ${activeNotification.groupKey ?? 'no group'}\n'
```
MaikuB, line 1963: nit: for consistency, could the null coalescing be removed? The example will already display "null" when a notification doesn't have a title or body

MaikuB:
Thanks so much for the PR. I left a comment there, would you be able to take a look?

flutter_local_notifications/example/lib/main.dart
drstranges: ok, i see.

PR #1385, opened 2021-11-12
[flutter_local_notifications] Adds google-java-format as GH actions

ened:
This adds automatic code formatting use google-java-format (the default for Flutter projects) to the Android side of the project. The GH action will **actually change the code** and create a commit to analyse.

This should help with merging code of multiple contributors.

@MaikuB WDYT

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/checkout@v2 # v2 minimum required
       - name: Test
         run: |
-          brew install clang-format
+          which clang-format || brew install clang-format
           find . -name '*.m' -exec clang-format -i {} \;
           git diff --exit-code
 
@@ -28,6 +28,6 @@ jobs:
       - uses: actions/checkout@v2 # v2 minimum required
       - name: Test
         run: |
-          brew install swiftlint
+          which swiftlint || brew install swiftlint
           swiftlint --fix
           git diff --exit-code
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -20,6 +20,8 @@ jobs:
         run: |
           which clang-format || brew install clang-format
           find . -name '*.m' -exec clang-format -i {} \;
+          find . -path '*/ios/**/*.h' -exec clang-format -i {} \;
+          find . -path '*/macos/**/*.h' -exec clang-format -i {} \;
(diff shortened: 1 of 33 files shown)


MaikuB:
I remember you bringing this up applying formatting before and I'm happy to go with this :)

Would you by chance have some details on how this could be run locally first? Also, is do you know of something similar for Objective-C?

ened:
@MaikuB the project is mostly using Swift, so that would use SwiftLint. I'll get it added to this branch.

MaikuB:
Thanks, Objective-C is still in use for the iOS implementation given the issues Flutter had with Swift back then. Swift is used for macOS.

Regarding formatting of Java, I forgot I could follow the trail of what the action uses and found https://github.com/google/google-java-format :)

ened:
@MaikuB ok sorry I did not properly scan - there is a lot of ObjC code. I have added swift & objc formatting now and some GH actions. This project is using Cirrus CI, can you check if Cirrus AND GH actions could be used?

MaikuB:
It looks like it should be fine as the checks still run and I've seen the official plugins repo used both as well

ened:
> It looks like it should be fine as the checks still run and I've seen the official plugins repo used both as well

In that case we should be ready to merge.

.github/workflows/format.yml
```diff
@@ -0,0 +1,33 @@
...
+          args: "--skip-sorting-imports --replace"
+
+  objc_format:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2 # v2 minimum required
+      - name: Test
+        run: |
+          which clang-format || brew install clang-format
+          find . -name '*.m' -exec clang-format -i {} \;
```
MaikuB, line 22: I'm not an expert with Linux commands but if i'm not mistaken this particular line should ideally change to `find . -name '*.m' -or -name '*.h' clang-format -i {} \;` to allow for it to analyse header (`.h`) files too

MaikuB:
I left one comment. Happy take advice on it and didn't want to push the suggested change directly on your PR and merge right away in case I'm wrong etc. It is getting late in Sydney though so probably won't get to continue on this until the following morning.

.github/workflows/format.yml
ened: Done. :)

PR #1391, opened 2021-11-15
Added support for active notifications on iOS

morvagergely:
I noticed that there were no support for active notifications on iOS, and I thought it could be useful. If the iOS version is older than 10, the `activeNotifications` will return an empty list. This however, is a breaking change.


MaikuB:
Thanks for the PR. Wanted to check, is the reason why you're saying it's a breaking change is because of the `getActiveNotifications` method being renamed? If so, I would actually say to revert back to the old name and it's the `pendingNotifications()` method that in hindsight should be changed to `getPendingNotifications()` though that is something that could be done in a separate PR. The reason for this is to improve readability as `await getPendingNotifications()` (IMO) better indicates that the code is awaiting an operation to be done. Without the `get` prefix, it the name looks more like something that should be associated with a property. The plugin also has more cases of having the `get` prefix as a convention too e.g. `getNotificationChannels()`

morvagergely:
I changed the function name back to `getActiveNotifications`. I added the nullable `payload` member to the `ActiveNotification` class. It might still be a breaking change if anyone ever initialized their own `ActiveNotification` with the class contructor, since the parameters are not optional/named.

MaikuB:
Thanks for making the change. Will try find some time to test this out before it's merged in. FYI I may change the target branch for this as there are some potentially breaking iOS/macOS changes I'm considering too. If I do, then I might create branch to combine these into a pre-release

MaikuB:
Just realised you raised a very good point about the `ActiveNotification`'s constructor. Adding the groupKey to it could be a breaking change for anyone that created instances of it. I'm going to remove that in one of the 9.x releases so it'll go in a 10.x release

flutter_local_notifications/macos/Classes/FlutterLocalNotificationsPlugin.swift
```diff
@@ -241,6 +243,25 @@ public class FlutterLocalNotificationsPlugin: NSObject, FlutterPlugin, UNUserNot
         }
     }
 
+    func getActiveNotifications(_ result: @escaping FlutterResult) {
+        if #available(OSX 10.14, *) {
+            UNUserNotificationCenter.current().getDeliveredNotificationRequests { (requests) in
```
MaikuB, line 248: Can you fix the compilation error? The name of the actual method is different. See https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649520-getdeliverednotifications. The code to extract the details would also need to change as a result though should be similar to what you had in the iOS implementation but translated to Swift

MaikuB (changes requested):
Thanks I've reviewed the PR and give it a spin. Since I removed `groupKey` you'll need to update your fork to resolve the conflicts, which should be a matter of keeping your changes instead. I've left some comments and found that the macOS implementation doesn't compile as it refers to a method that doesn't exist. Could you also change the target branch to the 10.0.0 branch that I've pushed up?

If you need some assistance then let me know

flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
```diff
@@ -234,6 +268,16 @@ - (void)pendingNotificationRequests:(FlutterResult _Nonnull)result {
   }
 }
 
+- (void)getActiveNotifications:(FlutterResult _Nonnull)result {
+  if (@available(iOS 10.0, *)) {
+    [self activeUserNotificationRequests:result];
+  } else {
+    NSMutableArray<NSMutableDictionary<NSString *, NSObject *> *>
+      *activeNotifications = [[NSMutableArray alloc] initWithCapacity:0];
+    result(activeNotifications);
```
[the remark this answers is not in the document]
morvagergely: I have found the `FlutterError` in this file. Could you help me what parameters to use in this case for `code`, `message`, and `details`?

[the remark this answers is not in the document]
MaikuB: Sure

- Currently on Android, code is `GET_ACTIVE_NOTIFICATIONS_ERROR_CODE` https://github.com/MaikuB/flutter_local_notifications/blob/a5a84f095a6e7a0d193a8acfc747427a9ff867d4/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L124. In hindsight, this could be done better to be generic error code on both Android, iOS and macOS (e.g. `OS_VERSION_NOT_SUPPORTED`. Options here are to update it, including on Android or you can iOS and macOS consistent and I can update it later
- `message` would be similar to what's on Android, which is currently `Android version must be 6.0 or newer to use getActiveNotifications`. Replace `Android` with `iOS`, `macOS` and then the version number with the appropriate OS version
- `details` can be omitted as it should be an optional parameter

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -99,11 +99,11 @@
       "deleteNotificationChannelGroup";
   private static final String CREATE_NOTIFICATION_CHANNEL_METHOD = "createNotificationChannel";
   private static final String DELETE_NOTIFICATION_CHANNEL_METHOD = "deleteNotificationChannel";
-  private static final String GET_ACTIVE_NOTIFICATIONS_METHOD = "getActiveNotifications";
   private static final String GET_NOTIFICATION_CHANNELS_METHOD = "getNotificationChannels";
   private static final String START_FOREGROUND_SERVICE = "startForegroundService";
   private static final String STOP_FOREGROUND_SERVICE = "stopForegroundService";
   private static final String PENDING_NOTIFICATION_REQUESTS_METHOD = "pendingNotificationRequests";
+  private static final String ACTIVE_NOTIFICATIONS_METHOD = "getActiveNotifications";
```
MaikuB, line 106: Forgot to mention that since the method from Dart side has been renamed back to `getActiveNotifications`, `ACTIVE_NOTIFICATIONS_METHOD` can be renamed to `GET_ACTIVE_NOTIFICATIONS_METHOD` for consistency. Similar comment applies to the iOS and macOS code

flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
```diff
@@ -234,6 +271,17 @@ - (void)pendingNotificationRequests:(FlutterResult _Nonnull)result {
...
 }
 
+- (void)getActiveNotifications:(FlutterResult _Nonnull)result {
+  if (@available(iOS 10.0, *)) {
+    [self activeUserNotificationRequests:result];
+  } else {
+    result([FlutterError
+      errorWithCode:GET_ACTIVE_NOTIFICATIONS_ERROR_CODE
+            message:GET_ACTIVE_NOTIFICATIONS_ERROR_MESSAGE
+            details:[NSNull null]]);
```
MaikuB, line 281: whilst passing [NSNull null] may work, my understanding is that it's more appropriate to pass `nil` as NSNull is more used for when `nil` is prohibited (ref: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/NumbersandValues/Articles/Null.html) and there's existing examples in the Flutter ecosystem where you find `nil` used e.g. https://github.com/flutter/plugins/blob/b548ae18646d1725a8e1f004c26bf42e4d362aa8/packages/battery/battery/ios/Classes/FLTBatteryPlugin.m#L62

MaikuB (commented):
Thanks for making the changes. Left some comments. I'll try to retest the changes later in the week

isaacfi:
Is this change included in the last version 9.3.2?

I tried to call the implementation for iOS but I have a compilation error:

The method 'getActiveNotifications' isn't defined for the type 'IOSFlutterLocalNotificationsPlugin'.
Try correcting the name to the name of an existing method, or defining a method named 'getActiveNotifications'

```
final platformImplementation =
        _notificationsPlugin!.resolvePlatformSpecificImplementation();
    if (platformImplementation is AndroidFlutterLocalNotificationsPlugin) {
      return await platformImplementation.getActiveNotifications();
    } else if (platformImplementation is IOSFlutterLocalNotificationsPlugin) {
      return await platformImplementation.getActiveNotifications();
    } else {
      return [];
    }

return await platformImplementation.getActiveNotifications(); <-- HERE
```

isaacfi:
Thanks @MaikuB,

I'll try the 10.0.0 prerelease, and when do you estimate become 10.0.0 into a stable release?

MaikuB:
No estimate for it as it'll move to stable when it's ready and also relies on the community contributing.

As this is open source you can fork, add the relevant bits you need and have your app point to the fork

Issue #1390, opened 2021-11-15
Android notification details is required?


LeonardoRosaa:
First of all, thanks for the package 😄 

**Describe the bug**

When not put notification details for `Android` Platform I received this error message:

```shell
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): Failed to handle method call
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): 	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.setupNotificationChannel(FlutterLocalNotificationsPlugin.java:981)
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): 	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.createNotification(FlutterLocalNotificationsPlugin.java:198)
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): 	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.showNotification(FlutterLocalNotificationsPlugin.java:1042)
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): 	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.show(FlutterLocalNotificationsPlugin.java:1376)
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): 	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.onMethodCall(FlutterLocalNotificationsPlugin.java:1262)
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:233)
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): 	at io.flutter.embedding.engine.dart.DartMessenger.handleMessageFromDart(DartMessenger.java:84)
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): 	at io.flutter.embedding.engine.FlutterJNI.handlePlatformMessage(FlutterJNI.java:865)
E/MethodChannel#dexterous.com/flutter/local_notifications(18666): 	at android.os.MessageQueue.nativePollOnce(Native Method)
[...]

PR #1393, opened 2021-11-16
[ flutter_local_notifications]: improved without android specific details

LeonardoRosaa:
The according issue #1390, is necessary android specific details to Android Platform. But, it's can confuse, because field is optional.

A simple fix is been add [assert](https://dart.dev/guides/language/language-tour#assert) to validate field for Android Platform.

### DIFF
--- a/flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
+++ b/flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
@@ -191,6 +191,10 @@ class FlutterLocalNotificationsPlugin {
       return;
     }
     if (defaultTargetPlatform == TargetPlatform.android) {
+      assert(
+        notificationDetails != null,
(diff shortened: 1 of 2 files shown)


flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
```diff
@@ -191,6 +191,10 @@ class FlutterLocalNotificationsPlugin {
       return;
     }
     if (defaultTargetPlatform == TargetPlatform.android) {
+      assert(
```
MaikuB, line 194: Actually in hindsight, I believe the more appropriate fix is for `AndroidNotificationDetails` to be non-nullable in the method signatures of methods for display/scheduling notifications on Android. I'll close this PR as there's a 10.0.0 pre-release i'm working on where this (breaking) change can be done

PR #1400, opened 2021-11-20
[flutter_local_notifications]remove addition of groupKey to ActiveNotification so that it goes in a major release

MaikuB:
FYI @drstranges that i'll be removing this from the 9.x stable release as I should published the changes in your PR (#1376) as part of a major release


drstranges:
Any changes? In release 10?

MaikuB:
It's there but was documented in the platform interface's change log. I'd need to go back to update the change log for the plugin

PR #1414, opened 2021-12-07
[flutter_local_notifications_linux] DBus 0.6 upgrade

12people:
A simple upgrade to DBus ^0.6.0, which now features named parameters.

### DIFF
--- a/flutter_local_notifications_linux/lib/src/dbus_wrapper.dart
+++ b/flutter_local_notifications_linux/lib/src/dbus_wrapper.dart
@@ -47,5 +47,6 @@ class DBusWrapper {
 
   /// Creates a stream of signal with the given [name].
   DBusRemoteObjectSignalStream subscribeSignal(String name) =>
-      DBusRemoteObjectSignalStream(object: _object, interface: _destination, name: name);
+      DBusRemoteObjectSignalStream(
+          object: _object, interface: _destination, name: name);
 }
--- a/flutter_local_notifications_linux/lib/src/dbus_wrapper.dart
+++ b/flutter_local_notifications_linux/lib/src/dbus_wrapper.dart
@@ -15,8 +15,8 @@ class DBusWrapper {
     _destination = destination;
     _object = DBusRemoteObject(
       DBusClient.session(),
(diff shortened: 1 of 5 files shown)


flutter_local_notifications_linux/lib/src/dbus_wrapper.dart
```diff
@@ -47,5 +47,5 @@ class DBusWrapper {
 
   /// Creates a stream of signal with the given [name].
   DBusRemoteObjectSignalStream subscribeSignal(String name) =>
-      DBusRemoteObjectSignalStream(_object, _destination, name);
+      DBusRemoteObjectSignalStream(object: _object, interface: _destination, name: name);
```
MaikuB, line 50: could you add a trailing comma to address the issue where it breaks 80 character line limit rule?

MaikuB:
Thanks for the PR. Left one comment for a minor code style issue. Besides that, would you be up for bumping the plugin to 0.4.0 and including a changelog entry about how the dbus dependency was bumped? Alternatively, I could try to add those in as part of your PR too. Let me know which approach you'd prefer.

12people:
Done.

Not sure what you meant by the trailing comma, but I used VSCodium's autoformatting to solve the 80 characters per line issue. Hope this solution is ok.

12people:
I bumped the flutter_local_notifications_linux version to 0.4.0 too, which is why I imagine tests are failing now, as the 0.4.0 version is not on pubdev yet.
Could you cherrypick this PR and revert that particular change?

MaikuB:
Not sure what you mean with cherry picking the PR but apologies as I may not have been clear enough on this. What could be done is to revert the change on `flutter_local_notifications` pubspec.yaml file but leave the change on the Linux plugin. From there, I could (squash) merge it, publish the Linux plugin and I can looking at bumping the Linux plugin dependency separately. That is you could leave that to me :)

Note: I'll need to see what tooling is available for assisting with situations like this. I believe it'll be something like
https://github.com/invertase/melos

12people:
Ok, reverted :)

PR #1422, opened 2021-12-20
[flutter_local_notifications] Implement 'areNotificationsEnabled' for Android

komape:
Implemented the ability to check if notifications are enabled for Android. Further details about the Android part of it, see the [official docs](https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#areNotificationsEnabled()).


MaikuB:
@komape thanks for the PR. Not sure what you're referring to but looks like it's all green. Could you elaborate further in case I'm missing something. Perhaps Cirrus/GitHub was temporarily reporting the wrong status?

komape:
@MaikuB Now everything's green. 

'Build Android example app' and 'Build Linux example app' showed the status that they were canceled. Maybe they were just scheduled for later and showed the wrong status.

flutter_local_notifications/README.md
```diff
@@ -166,6 +167,7 @@ Notification icons should be added as a drawable resource. The example project/c
 
 When specifying the large icon bitmap or big picture bitmap (associated with the big picture style), bitmaps can be either a drawable resource or file on the device. This is specified via a single property (e.g. the `largeIcon` property associated with the `AndroidNotificationDetails` class) where a value that is an instance of the `DrawableResourceAndroidBitmap` means the bitmap should be loaded from an drawable resource. If this is an instance of the `FilePathAndroidBitmap`, this indicates it should be loaded from a file referred to by a given file path.
 
+If the plugin is initialized in the `main()` method make sure that `WidgetsFlutterBinding.ensureInitialized()` is executed before. Otherwise it can happen, that the plugin throws Null Pointer Exceptions. Also see issue [#1364](https://github.com/MaikuB/flutter_local_notifications/issues/1364) for more information.
```
MaikuB, line 170: This doesn't have to do with plugin being initialised. This is to do with trying to use platform channels in general so it could help with calling any method that uses platform channels.

komape: Oh, this is from another issue where I created a PR weeks ago. Seems like it was accidentally merged into this one. Should I keep it in or remove it?

MaikuB: If you could remove it then that would be preferred. Regarding submission of PRs, a tip I learnt that can help in situations like this is to create a branch in your fork instead of having it all on the main/master. That allows the latter to be kept up to date more easily and can separate features/fixes per branch

komape: done.

MaikuB:
FYI @ened that I also merged this in

PR #1433, opened 2021-12-29
Adds support for melos 

ened:
During the work on #880 I noticed the project could be managed slightly easier rgd. dependencies, tests and so on.

The `melos` package could help with this - now you can run things like this:

```
melos bootstrap
melos exec flutter pub get
melos run test
melos run analyze
```

The `test` commands is probably the most useful as it can run for all the packages.

It is not integrated in the CI process, yet, but could be.

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -1,0 +1,21 @@
+name: flutter_local_notifications
+repository: https://github.com/MaikuB/flutter_local_notifications
+packages:
+  - flutter_local_notifications/*
+  - flutter_local_notifications_linux/*
+  - flutter_local_notifications_platform_interface/*
+
+ide:
+  intellij: false
+
+scripts:
+  analyze:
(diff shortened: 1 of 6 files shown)


MaikuB:
Thanks for the PR. Was going to look into melos so much appreciated for this :)

Will look further into this tomorrow

melos.yaml
```diff
@@ -0,0 +1,21 @@
...
+ide:
+  intellij: false
+
+scripts:
+  analyze:
+    run: melos exec -c 1 -- "dart analyze . --fatal-infos"
+    description: Run dart analyzer in a specific package.
+  test:
+    description: Run tests in a specific package.
+    run: melos exec --concurrency=1 -- "flutter test"
```
MaikuB, line 17: nit: could use `-c 1` like the `analyze` script does for consistency. Having said that, I'll merge this in as this is something I could address in a separate PR to rework CI to make use of melos :)

PR #1435, opened 2021-12-30
Rework CI to use melos

MaikuB:
Continues on #1433 where support for melos was added

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,6 @@ task:
   pub_cache:
     folder: ~/.pub-cache
   upgrade_script:
-    - flutter channel stable
     - flutter upgrade
   install_melos_script:
     - dart pub global activate melos
@@ -89,7 +88,6 @@ task:
   pub_cache:
     folder: ~/.pub-cache
   upgrade_script:
-    - flutter channel stable
     - flutter upgrade
   install_melos_script:
     - dart pub global activate melos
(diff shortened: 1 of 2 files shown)


MaikuB:
FYI @ened that this is merged in now so you may want to update #880. Haven't integrated the `analyze` and `format` scripts as that is probably best to be part of GitHub actions that I'd need to do more reading in. Thanks again for starting the work on this

ened:
@MaikuB ok will do. Will you add iOS integration tests as well?

MaikuB:
I would but I had trouble figuring out how to run those on Cirrus before so if there's a reference I could use for that then let me know. I'll do some more searching in the mean time

ened:
@MaikuB start simulator and run the test I think. Try this file for reference https://github.com/fluttercommunity/flutter_workmanager/blob/main/.github/workflows/test.yml

MaikuB:
Yep from memory, my issue was finding the appropriate name and iOS version of the simulator to start. I'll take another look and see if I switch to using GitHub actions for all of the checks. Haven't stayed up to date with it but like it supports running on macOS VMs so should work

ened:
@MaikuB I think the simulator spin-up code will work for both GH Actions and Cirrus, no need to switch all of the CI unless you planned to anyway..

MaikuB:
Yep I know though switching isn't just to get this working, it's more to allow all the checks, formatting etc occur as part of one workflow. 

I didn't use GitHub actions years ago as from memory it seemed limited in what it could do back then

MaikuB:
Hmm looks like getting things like Flutter setup may involve scripts so I'll park the idea of switching for now...

PR #1438, opened 2021-12-30
Update CI to run integration tests on iOS

MaikuB:


MaikuB:
@ened integration tests should now run on iOS. You may notice from the commits that I originally tried to get it working on macOS and Linux too but to no avail. I don't think anything can be done about macOS as there's the permissions prompt to deal with. It's not an issue on iOS by making use of https://github.com/wix/AppleSimulatorUtils. Not sure what's going on with Linux though, it seems like the app refuses to run. Might be an issue with priviledges

PR #1442, opened 2022-01-04
[flutter_local_notifications_linux] Add actions support in Linux

proninyaroslav:
All action features that are available in the Desktop Notification Specification have been implemented. Actions such as textbox are not implemented due to the lack of a standardized implementation. The `onSelectNotificationAction` callback currently doesn't work, because `flutter_local_notifications` depends on the old version of `linux_local_notifications_linux` and I can't map `onSelectNotificationAction` to Linux implementation. After the release of Linux plugin, you need to add it here:
https://github.com/MaikuB/flutter_local_notifications/blob/3a9eae23cd723175908d7809a9d58374c4243e16/flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart#L141-L146
For the test, you can change `pubspec.yaml` of  `flutter_local_notifications` this way:
```YAML
flutter_local_notifications_linux:
    path: '../flutter_local_notifications_linux'
```

Tested on Fedora 35 (GNOME 41):

Снимок экрана от 2022-01-04 20-18-29

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -49,7 +49,7 @@ A cross platform plugin for displaying local notifications.
 * **Android 4.1+**. Uses the [NotificationCompat APIs](https://developer.android.com/reference/androidx/core/app/NotificationCompat) so it can be run older Android devices
 * **iOS 8.0+**. On iOS versions older than 10, the plugin will use the UILocalNotification APIs. The [UserNotification APIs](https://developer.apple.com/documentation/usernotifications) (aka the User Notifications Framework) is used on iOS 10 or newer.
 * **macOS 10.11+**. On macOS versions older than 10.14, the plugin will use the [NSUserNotification APIs](https://developer.apple.com/documentation/foundation/nsusernotification). The [UserNotification APIs](https://developer.apple.com/documentation/usernotifications) (aka the User Notifications Framework) is used on macOS 10.14 or newer.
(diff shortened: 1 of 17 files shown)


proninyaroslav:
By the way, I forgot to mention: unlike Android & iOS, Linux runs the `onSelectNotificationAction` callback in the current main isolate without creating a separate one. This is due, among other things, to the initial limitations (please see `Linux limitations` in the README), as well as to the fact that the Flutter Engine implementation for Linux doesn't have the ability to start the engine with a custom Dart entry point, I opened an issue about this a long time ago https://github.com/flutter/flutter/issues/86599. Windows seems to have a similar problem https://github.com/flutter/flutter/issues/93537

MaikuB:
Awesome, thanks so much for this! I'll take a look further. Note that the plugin has recently added [melos](https://github.com/invertase/melos) to allow working with monorepos such as this one more easily. It'll search for references to local packages and will do the overrides behind the scenes. This would've allowed update the `initialize` method and can use melos to run the example app. Steps taken to do so would be to install melos first, run `melos bootstrap`  then run `melos exec "flutter run"`. The latter command will search all of the packages, eventually find the example app and then execute `flutter run` from there :)

Regarding the support for background isolates on Linux, I'm not surprised as macOS currently has a similar limitation that @ened had pointed out. I raised an issue that led to discussions about process-wide isolates a long time ago as well (https://github.com/flutter/flutter/issues/23904)

PR #1459, opened 2022-01-24
add coloring background support in android

benechiu:

### DIFF
--- a/flutter_local_notifications/example/lib/main.dart
+++ b/flutter_local_notifications/example/lib/main.dart
@@ -722,18 +722,18 @@ class _HomePageState extends State<HomePage> {
                         },
                       ),
                       PaddedElevatedButton(
+                        buttonText:
+                            'Start foreground service with blue background notification',
+                        onPressed: () async {
+                          await _startForegroundServiceWithBlueBackgroundNotification();
+                        },
+                      ),
+                      PaddedElevatedButton(
                         buttonText: 'Stop foreground service',
                         onPressed: () async {
                           await _stopForegroundService();
                         },
-                      ),
-                      PaddedElevatedButton(
-                        buttonText:
-                            'Show colored background notification with forgground service',
-                        onPressed: () async {
-                          await _showColoredBackgroundNotification();
-                        },
-                      ),
+                      ),                      
                     ],
                     if (!kIsWeb &&
                         (Platform.isIOS || Platform.isMacOS)) ...<Widget>[
@@ -989,8 +989,6 @@ class _HomePageState extends State<HomePage> {
       importance: Importance.max,
       priority: Priority.high,
       ticker: 'ticker',
-      color: Colors.blue,
-      colorized: true,
       actions: <AndroidNotificationAction>[
         AndroidNotificationAction(
           'text_id_1',
@@ -2125,14 +2123,7 @@ class _HomePageState extends State<HomePage> {
             payload: 'item x');
   }
 
-  Future<void> _stopForegroundService() async {
-    await flutterLocalNotificationsPlugin
-        .resolvePlatformSpecificImplementation<
-            AndroidFlutterLocalNotificationsPlugin>()
-        ?.stopForegroundService();
-  }
-
-  Future<void> _showColoredBackgroundNotification() async {
+  Future<void> _startForegroundServiceWithBlueBackgroundNotification() async {
     const AndroidNotificationDetails androidPlatformChannelSpecifics =
         AndroidNotificationDetails(
       'your channel id',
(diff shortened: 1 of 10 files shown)


benechiu:
Screenshot_20220124-181714325

flutter_local_notifications/example/lib/main.dart
```diff
@@ -191,6 +191,11 @@ Future<void> main() async {
   await flutterLocalNotificationsPlugin.initialize(
     initializationSettings,
     onSelectNotification: (String? payload) async {
+      await flutterLocalNotificationsPlugin
```
MaikuB, line 194: Could you revert this change? Whilst I get this was originally done, it has a few issues
- it's not related to this PR
- results in an unnecessary call every time a notification is tapped on
- leads to inconsistent behaviour with when a notification is tapped on whilst app is running and when it's tapped but app wasn't running

flutter_local_notifications/example/lib/main.dart
```diff
@@ -1289,6 +1301,40 @@ class _HomePageState extends State<HomePage> {
     return filePath;
   }
 
+  Future<void> _showBigPictureNotificationColoringBackground() async {
+    final String largeIconPath = await _downloadAndSaveFile(
```
MaikuB, line 1305: Does demonstrating the `colorized` functionality need to be tied to showing pictures? If not, could this example code snippet be isolated to bare minimum to demonstrate it?

flutter_local_notifications/pubspec.yaml
```diff
@@ -9,8 +9,10 @@ dependencies:
   clock: ^1.1.0
   flutter:
     sdk: flutter
-  flutter_local_notifications_linux: ^0.4.1
-  flutter_local_notifications_platform_interface: ^6.0.0
+  flutter_local_notifications_linux: #^0.4.1
```
MaikuB, line 12: Could you please revert all the pubspec.yaml changes? It's not necessary I suspect the reason you did these changes was so you could test more easily. If so, note this plugin recently changed to use [melos](https://melos.invertase.dev). If you bootstrap melos then it will do changes behind the scenes to use the local dependencies

MaikuB (changes requested):
Thanks for the PR, see the comments I left there. Note that tests are also failing as you'll need to update the Dart tests to account for the addition of the `colorized` property

flutter_local_notifications/example/lib/main.dart
```diff
@@ -982,6 +989,8 @@ class _HomePageState extends State<HomePage> {
       importance: Importance.max,
       priority: Priority.high,
       ticker: 'ticker',
+      color: Colors.blue,
```
MaikuB, line 992: was there a reason for making this change? It looks like this change has had no effect on the result so i'm guessing it was from your experiments and should be removed

flutter_local_notifications/example/lib/main.dart
```diff
@@ -1015,6 +1024,13 @@ class _HomePageState extends State<HomePage> {
     await flutterLocalNotificationsPlugin.show(id++, 'Text Input Notification',
         'Expand to see input action', platformChannelSpecifics,
         payload: 'item x');
+
+    // await flutterLocalNotificationsPlugin
```
MaikuB, line 1028: as it's commented out (presumably from your testing), could this be removed?

flutter_local_notifications/example/lib/main.dart
```diff
@@ -727,6 +727,13 @@ class _HomePageState extends State<HomePage> {
                           await _stopForegroundService();
                         },
                       ),
+                      PaddedElevatedButton(
+                        buttonText:
+                            'Show colored background notification with forgground service',
```
MaikuB, line 732: For consistency, could you shift this button to be right after the existing "Start foreground service" one? The button text could then become "Start foreground service with notification that has a blue background". This also makes it more intuitive for users to locate the button to stop the foreground service as it would be below the button that starts a foreground service.

`_showColoredBackgroundNotification()` could then be renamed to something like `_startForegroundServiceWithBlueBackgroundNotification()`

MaikuB (changes requested):
Thanks for making the changes. I left some additional comments and questions

MaikuB (approved):

MishaPadalka:
@MaikuB @benechiu Hello! Is this functionality working now?

Levi-Lesches:
It should! Try setting `AndroidNotificationDetails.color` and `.colorized = true`. However, note the following from the [Android docs](https://developer.android.com/reference/android/app/Notification.Builder#setColorized(boolean)): 

> Set whether this notification should be colorized. When set, the color set with [setColor(int)](https://developer.android.com/reference/android/app/Notification.Builder#setColor(int)) will be used as the background color of this notification
> 
> This should only be used for high priority ongoing tasks like navigation, an ongoing call, or other similarly high-priority events for the user
>
> For most styles, the coloring will only be applied if the notification is for a foreground service notification. However, for [MediaStyle](https://developer.android.com/reference/android/app/Notification.MediaStyle) and [DecoratedMediaCustomViewStyle](https://developer.android.com/reference/android/app/Notification.DecoratedMediaCustomViewStyle) notifications that have a media session attached there is no such requirement.

If you're having issues getting this to work, please open a new issue with your code and what you've tried.

Issue #746, opened 2020-08-07
Desktop support (Linux, Windows)


proninyaroslav:
I see that you are trying to implement macOS support and it would be appropriate to add Windows and Linux as well. Windows has a native Win32 API, while Linux uses `libnotify` library, which covers cases for different desktop environments.

MaikuB:
As the plugin implements a federated plugin structure, the ability to add support for other platforms is open to the community to do so. It sounds like you're interested in leveraging that work, if so then I would suggest you contribute back by adding support for the other platforms.

proninyaroslav:
Yes, I'll try to help you as soon as I get free time.

MaikuB:
Note that I'm not familiar with either Win32 or Linux. As far as I know, they require using C++, which I'm not particularly familiar with so I'm not well-positioned to implement this myself, though I could probably scrap bits of knowledge here and there to review code if needed

proninyaroslav:
In general, any language that supports FFI can be used. But C++ is probably the easiest option.

proninyaroslav:
Although it might be worth using Dart + FFI.

artemsivcev:
Hi! Our team want to help with this issue. Is there any new information about this feature?

MaikuB:
If you want to help then the best way is to submit a PR. There's a draft one for Linux support https://github.com/MaikuB/flutter_local_notifications/pull/888 that you help the author with, help test out etc

MaximRyabovol:
We have some decision for windows notifications. It uses a [process_run](https://pub.dev/packages/process_run) plugin and powershell commands. It works perfectly in debug, but in realise version of the app there are some problems. If you interesting you can find our variant [here](https://github.com/flutter/flutter/issues/70926).

HrX03:
For linux i found a lib using gnome/gtk notification model https://github.com/canonical/desktop_notifications.dart, might want to take a peek. Saw a pr that already tried to use this model but had some issues iirc, maybe this one could work better.

jinfagang:
Why close? This **still an issue!!**

MaikuB:
@jinfagang this wasn't closed and has a needs help label even so as mentioned before in #1183, support for this would need a PRs/contributions from the community

Sunbreak:
A PoC of notification plugin on Android/iOS/Windows/Linux/macOS: https://github.com/woodemi/quick_notify

proninyaroslav:
[...]

PR #1473, opened 2022-02-09
[flutter_local_notifications] [WIP] Windows implementation of notifications

kennethnym:
This draft PR contains Windows implementation of this plugin. Still a WIP.

- [ ] `initialize` **WIP**
- [ ] `show` **WIP**
- [ ] `periodicallyShow` (not sure if possible due to limitation of win32)
- [x] `cancel` 
- [x] `cancelAll` 
- [ ] `pendingNotificationRequests`
- [x] `onSelectNotification` **WIP**

cc @azchohfi
Related issue: #746


azchohfi:
Hey @kennethnym, I've cloned your branch and built the source, and a simple plain notification worked totally fine for me.
The problem with your particular code is the template=ToastGeneric. If you switch that to, for example, ToastText01, then it works fine.
This value needs to be one of the valid [TileTemplateType](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Notifications.TileTemplateType?view=winrt-22000)

kennethnym:
I tried using the default templates but it would just show "New notification" instead of the actual body. Q

kennethnym:
@azchohfi To illustrate, this is the code that I have at the moment:
```c++
const auto doc = winrt::Windows::UI::Notifications::ToastNotificationManager::GetTemplateContent(winrt::Windows::UI::Notifications::ToastTemplateType::ToastText01);
const auto nodes = doc.GetElementsByTagName(L"text");
nodes.Item(0).AppendChild(doc.CreateTextNode(winrt::to_hstring(title)));

winrt::Windows::UI::Notifications::ToastNotification notif{ doc };
const auto notifier = winrt::Windows::UI::Notifications::ToastNotificationManager::CreateToastNotifier(L"com.dexterous.example");

notifier.Show(notif);
```
and this is the toast shown:
image

`title` is `'plain title'`, but it is not shown in the notification.

azchohfi:
Oh, I see. The parameters are being sent from dart to C++, but your GetString method is not querying them properly.
You should wrap the key inside an EncodableValue:
```c++
	const auto pair = m->find(flutter::EncodableValue(key));
```

Also, when you call the GetString method, you should store it not using a reference(&):
```c++
		const auto body = Utils::GetString("body", args).value();
```

You can use Visual Studio 2022 or 2019 to attach to the running exe and debug the C++ code. There is also a solution at "flutter_local_notifications\flutter_local_notifications\example\build\windows\example.sln" that will have the reference to all the files so you can add breakpoints, etc.

I've also found another issue. You need to add the id's to the texts, like this:
```c++
		XmlDocument doc;
		doc.LoadXml(L"\
			<toast>\
				<visual>\
					<binding template=\"ToastText01\">\
						<text id=\"1\">abc</text>\
						<text id=\"2\">def</text>\
					</binding>\
				</visual>\
			</toast>");
```

You can also debug this using cout:
```c++
std::cout << winrt::to_string(doc.GetXml()) << std::endl;
```

I've tested this changes here and now the notifications shows properly:
image

Let me know if this helps!

kennethnym:
Basic title body notification is working, but nothing beyond that.

kennethnym:
@lzzy12 Just an update: I have implemented `cancel` and `cancelAll` in the latest commit.

MaikuB:
FYI this reminds me that `zonedSchedule` currently part of the platform interface but should be so that will happen soon. Note that there's a 10.0.0 branch right now for the 10.0.0 prerelease

azchohfi:
@kennethnym any updates on this PR?

kennethnym:
Hi sorry, busy with coursework at the moment, I will continue the work this weekend.

kennethnym:
Just a bit of progress update: I am currently in the process of implementing onSelectNotification callback.

kennethnym:
At the moment, `onSelectNotification` is only called once, after a toast is clicked and the callback invoked, Windows would not notify the callback of interactions with subsequent toasts. I am investigating what is going wrong.

kennethnym:
Sorry for the silence, I am currently busy with courseworks. At the moment I am still stuck with the issue above. I would really appreciate it If anyone can help me with it.

azchohfi:
Hello @kennethnym. I've tried your branch and I managed to make it work more than once. Cherry-pick this commit from my fork:
https://github.com/azchohfi/flutter_local_notifications/commit/d363e5c7e314a2a484996d42100d89cfaeb7b59e

kennethnym:
Ahh i see the issue now, its `REGCLS_MULTIPLEUSE` vs `REGCLS_SINGLEUSE`. I can't believe I have missed such an obvious thing. Thank you so much.

sidevesh:
@kennethnym I tried your WIP branch and had some issues with notifications and msix packaging,
@azchohfi found the culprit,
https://github.com/YehudaKremer/msix/issues/118#issuecomment-1081134367 something to keep in mind.
Looks like we might have to use the correct `CreateToastNotifier` method based on if the app is packaged or not.

kennethnym:
Thanks for bring that to attention, I'll investigate how to handle msix apps as well.

kennethnym:
@sidevesh Can you test if the latest commit fixes your issue?

sidevesh:
I get the following error when building an app with the latest commit:
```
C:\Users\test\Projects\Side-projects\tomatoro\tomatoro-app\windows\flutter\ephemeral\.plugin_symlinks\flutter_local_notifications\windows\flutter_local_notifications_plugin.cpp(224): error C2220: the 
following warning is treated as an error [C:\Users\test\Projects\Side-projects\tomatoro\tomatoro-app\build\windows\plugins\flutter_local_notifications\flutter_local_notifications_plugin.vcxproj]      
C:\Users\test\Projects\Side-projects\tomatoro\tomatoro-app\windows\flutter\ephemeral\.plugin_symlinks\flutter_local_notifications\windows\flutter_local_notifications_plugin.cpp(224): warning C4715: '`anonymous namespace'::FlutterLocalNotificationsPlugin::Initialize': not all control paths return a value [C:\Users\test\Projects\Side-projects\tomatoro\tomatoro-app\build\windows\plugins\flutter_localBuild process failed.
 
#0      WindowsBuild.build (package:msix/src/windowsBuild.dart:33:7)
<asynchronous suspension>
#1      Msix._createMsix (package:msix/msix.dart:73:7)
<asynchronous suspension>
#2      Msix.create (package:msix/msix.dart:33:5)
<asynchronous suspension>
#3      main (file:///C:/Users/test/AppData/Local/Pub/Cache/hosted/pub.dartlang.org/msix-3.3.1/bin/create.dart:4:3)
<asynchronous suspension>
pub finished with exit code 255
```

kennethnym:
Thanks for reporting, it's an easy fix.

kennethnym:
@sidevesh the issue should be fixed.

sidevesh:
Hey @kennethnym , after the latest commit, I am able to build the app but notification has stopped showing up now.
I had actually made a simple patch earlier replacing the `CreateToastNotifier` with the correct version and had gotten it to work with the app name showing up correctly, so we should be close to a solution.
I encountered another issue which is that on clicking the notification it opens another copy of the app rather than bring the app to focus or dismiss.
In macOS and Linux, clicking the notification without any on `SelectNotification` just dismisses the notification and brings the app to focus and I am trying to replicate that on Windows.
Is it possible to implement that behavior ?

kennethnym:
@sidevesh From the issue you linked earlier, @azchohfi mentioned:
> For deciding whether to use the overloaded method...
HasIdentity true = CreateToastNotifier()
HasIdentity false = CreateToastNotifier(myAumid)"

I have in fact integrated this into the code in previous commits already:
```c++
if (hasIdentity.value())
    toastNotifier = winrt::Windows::UI::Notifications::ToastNotificationManager::CreateToastNotifier();
else
    toastNotifier = winrt::Windows::UI::Notifications::ToastNotificationManager::CreateToastNotifier(winrt::to_hstring(aumid));
```
Is the app name still not showing up correctly?

As for the other issues, I still need some time to figure out what's going on that is causing them to occur.

sidevesh:
hey @kennethnym , tried it again and same  behavior, notifications have stopped showing up,
so can't really see if the app name in notification is fixed or not.

sidevesh:
Like I mentioned previously I had simply replaced the `CreateToastNotifier` with the correct version and managed to fix the app name,
although that would mess up non msix apps so your fix with the check built in is the proper solution here.

kennethnym:
I am struggling to understand why it would stop working altogether. Exams are coming up for me; I'll start working on it again after exams.

sidevesh:
Also, any pointers about the issue: `on clicking the notification it opens another copy of the app rather than bring the app to focus or dismiss` ?

kennethnym:
I have no idea what is causing that issue either. It wasn't happening before, and I never changed the code related to that.

sidevesh:
> I have no idea what is causing that issue either. It wasn't happening before, and I never changed the code related to that.

Okay, I will try testing this in a non msix build to see if msix packaging is causing the difference in behavior, might help us debug this issue.

keanallen:
Almost there buddy!

DomingoMG:
I'm looking forward to using this on Windows, isn't it long?

keanallen:
please :(

kennethnym:
Sorry for the silence. I've been very busy lately, so I was not able to squeeze time to work on this. I will give this another look over the weekend.

keanallen:
Hello, guys

flutter_local_notifications/windows/flutter_local_notifications_plugin.cpp
```diff
@@ -0,0 +1,308 @@
...
+			<toast>\
+				<visual>\
+					<binding template=\"ToastGeneric\">\
+					</binding>\
+				</visual>\
+			</toast>");
+
+		const auto bindingNode = doc.SelectSingleNode(L"//binding[1]");
+
+		if (title.has_value()) {
```
lightrabbit, line 250: According to this link: https://docs.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/adaptive-interactive-toasts?tabs=xml
Maybe we can pass the XML text directly to the plugin and then combine the title and content etc into XML text on the Dart code side?
In this way, more Toast styles can be easily supported in the future.

flutter_local_notifications/windows/registration.cpp
```diff
@@ -0,0 +1,268 @@
...
+/// 68d0c89d-760f-4f79-a067-ae8d4220ccc1
+/// </summary>
+static constexpr winrt::guid CALLBACK_GUID{
+	0x68d0c89d, 0x760f, 0x4f79, {0xa0, 0x67, 0xae, 0x8d, 0x42, 0x20, 0xcc, 0xc1}
+};
+
+/// <summary>
+/// String representation of the callback GUID.
+/// </summary>
+const std::string CALLBACK_GUID_STR = "{68d0c89d-760f-4f79-a067-ae8d4220ccc1}";
```
lightrabbit, line 32: According the resources doc in  UpdateRegistry. It looks like every application should have a unique GUID, so the GUID here should be configurable. Maybe change the GUID to be passed from the initialization parameter instead?

flutter_local_notifications/windows/flutter_local_notifications_plugin.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+	std::optional<bool> FlutterLocalNotificationsPlugin::HasIdentity() {
+		if (!IsWindows8OrGreater()) {
+			// OS is windows 7 or lower
+			return false;
+		}
+
+		UINT32 length;
+		auto err = GetCurrentPackageFullName(&length, nullptr);
+		if (err != ERROR_INSUFFICIENT_BUFFER) {
```
lightrabbit, line 186: When I run the example here, it returns the error code 87(ERROR_INVALID_PARAMETER) which causes the initialization to fail. But when I still return false(not std:nullopt) here, the plugin still works fine.

lightrabbit (commented):
I found some parts in the code that could be improved, could you please take a look?

flutter_local_notifications/windows/flutter_local_notifications_plugin.cpp
kennethnym: That is something that I have thought about actually - a Windows specific method that allows users to pass in XML of the notification toast directly.

flutter_local_notifications/windows/registration.cpp
kennethnym: Gotcha, will fix.

kennethnym:
Quick update - I recently switched to working on a MacBook fully as I will be away from my main computer for an extended period. Parallel seems to be a good option for running Windows on a MacBook. As soon as I get that up and running I should be able to start working on this (longrunning) PR, addressing @lightrabbit 's helpful comments as well.

flutter_local_notifications/windows/flutter_local_notifications_plugin.cpp
lightrabbit: I found the cause of the problem. The variable `length` needs to be initialized.
Replace `UINT32 length;` to `UINT32 length = 0;` could resolve this issue.

keanallen: Wow! Can't wait! Thank you guys for working on this.

kennethnym:
I think this is a known issue at the moment. Unfortunately I don't have the time at the moment to look into this, but when I do I will see what's going I'll see what I can do to fix it!

xBLACKICEx:
hi @kennethnym is there any news for support on windows ？

Levi-Lesches:
This can be closed since #2366 was merged

Issue #1465, opened 2022-02-01
How to access the event of the notification execution


touagueni:
I am using Flutter Local Notifications Plugin to schedule notifications on demand in my app.

1.The notifications work fine and arrive as scheduled.
2.The onSelectNotification executes fine when the particular notification is selected. ... await flutterLocalNotificationsPlugin.initialize( initializationSettings, onSelectNotification: selectNotification ); ...

The question: how to access the event of the notification at the execution time?, in other words when a local scheduled notification gets delivered.

Thanks

MaikuB:
There's no such event for it within the plugin. If it's technically possible to implement then this is something you'll need to look it and could submit a PR back to the repo

touagueni:
Hi MaikuB,
Many Thanks for your reply.

Sorry I have no idea on how to implement it, at the same time we need it badly for our app. We have actually used server based notification first and we had access the event in question (i.e. the reception of a notification) then because our app will be used in parts of the world where the internet bandwidth was not particularly good we have decided to go for local notifications for the reminders. Local notifications does the job perfectly for what we are trying to do, except that we don't know how to show the number of notifications on the notifications badge inside the app.
Any help would appreciated.

Thanks and Kind Regards

Tewfik

MaikuB:
As you pivoted your original solution to use native APIs, then I would said you really be trying to understand said APIs and prepared to develop a solution using them as open-source libraries may not cover all scenarios. I'm not able to help further myself as this is something I maintain in my own spare time and I have my own priorities too. So as mentioned before, you'll likely need to look into further yourself or seek others to help. It may be that this requires the use of background isolates that were used for notification actions and this was something that contributed by the community. As this is open-source, you're free to fork the repo as a starting point and make the appropriate modifications

okbye123:
This will help https://www.youtube.com/watch?v=aQTvI8L-mZs

PR #1482, opened 2022-02-12
[flutter_local_notifications] Implement\Extend 'didReceiveLocalNotification'

vizhan-lanars:
Callback for handling when a notification is triggered while the app is
in the foreground. `didReceiveLocalNotification` callback will fire on new notification on Android & iOS (version 10+ as well).
Related issue (#1465).


MaikuB:
Thanks for PR. A couple of questions I should've earlier as this will play part on if this would be merged

- what use case are you looking to cover?
- is there a reason why this is only specific to Android given Flutter would target multiple platforms?

vizhan-lanars:
1. I'm developing an app in which there is an icon that informs users about unread/new notifications. Currently while being on the screen with this widget there is no way to update it in real-time when notification received.
2. I aimed to have it available on mobile app: iOS and Android. `didReceiveLocalNotification` callback already exists in iOS. Unfortunately, I am not familiar with other platform capabilities.

MaikuB:
> 1. I'm developing an app in which there is an icon that informs users about unread/new notifications. Currently while being on the screen with this widget there is no way to update it in real-time when notification received.

What do you then when the app isn't running? How do you deal with when a notification ends up being dismissed? Whilst I don't know the full details behind your scenarios, this sounds like something that is better handled via a few options (a) server side (potentially with push notifications), (b) badges on the app icon or (c) using the `getActiveNotifications()` method

> 2. I aimed to have it available on mobile app: iOS and Android. didReceiveLocalNotification callback already exists in iOS. Unfortunately, I am not familiar with other platform capabilities.

Given the scenarios you're looking to cover and how you think the callback exists on iOS leads me to think you may have a misunderstanding on how it works on iOS. Have you actually tried using that callback and checked the docs on what it's applicable to?

vizhan-lanars:
I've implemented the same for iOS. 
Please, take a look.

vizhan-lanars:
@MaikuB 
Could you take a look at this PR?
Is it a valuable feature for you? Because for me it is.

MaikuB:
Will take a closer look when I can do so as this requires more thought and testing the recent smaller PRs that have come through recently

MaikuB:
Looks like this works as expected and if it were to be merged in then would need to see if it can be expanded to other platforms. However, if you look at why this callback existed and how it worked before then this PR introduces a potentially significant breaking change. The callback existed because on iOS versions older than 10, when the app is in the foreground a notification won't be shown. This isn't a limitation with the plugin but because of how older versions of iOS behaves. With this callback, apps could decide what to do when this scenario happens. A common solution as demonstrated by the example app is to present a dialog. Since this PR now makes the callback happen on all versions of iOS and Android, the community would now need to make sure that the logic they had before only applies to iOS versions older than 10. If you want to reproduce this issue, you can run the example app on your fork and compare that with the previous behaviour. Using your fork, you'll find that whenever a notification is triggered and the app is in the foreground that a Cupertino dialog will always appear. Have you considered how the community can deal with this and if so, what would the options be?

vizhan-lanars:
**Scenario 1**
If this change would be a part of the plugin major version update it would be acceptable to have breaking changes. In this case, the community could handle showing notifications for older iOS versions (foreground application) on client-side (e.g. make a platform & OS version check within `dart:io` Platform).
**Scenario 2**
Consider adding a new common callback on a par with `onSelectNotification` on plugin initialization. It would be better for further development (other platform implementation).

In any case, implementation for Android & iOS is a good start. It'll cover the main needs at the beginning.

MaikuB:
To clarify, when you say "Scenario 1" and "Scenario 2", did you mean to express as different solutions to the question I posed? If so, though I have my own thoughts, I'm curious to know which approach you believe is better. It sounds like it would be the 2nd one (?)

vizhan-lanars:
Yes, you're right.
**Scenario 2** would be better from my point of view. Because with current implementation it's needed to pass the same callback for every platform setting.

MaikuB:
Apologies for late response. I agree with your view too. With that in mind, do you think you could update the PR to match this? As for the name of the new callback, I would suggest it to be called something like `onDidShowNotification`. The reason for the naming is to align with some proposed changes discussed in https://github.com/MaikuB/flutter_local_notifications/pull/1489 around dealing when a user taps on a notification or notification action

vizhan-lanars:
Hi, sure. I'll come back to this PR later.
Here's a lot of stuff to do, because of russian war
**Pray**

MaikuB:
I take it based on your response that you are based in Ukraine but yeah I hope this war stops and that you stay safe. I've been checking regularly what the updates are. Hope this ends soon :(

schoendiener:
Are there any news about this feature? I would need this in my app to play custom sounds in background when the notification is shown since I want to play soundfiles recorded during runtime and, as far as I understand, these are not possible to set as notification sound.

PR #1487, opened 2022-02-15
[flutter_local_notifications] fix colorized backwards compatibility issue

MaikuB:


MaikuB:
FYI @benechiu that this fix needed to be done to follow up on the PR that was submitted as scheduled notifications in older versions of the plugin will have a null `colorized` property after the JSON representing the notification is deserialised after being read from shared preferences

PR #1489, opened 2022-02-15
[flutter_local_notifications] 10.0.0-dev.8 Support for Android Action App Launch

AdamBridges:
This PR allows you to launch your app from an Android notification action.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -249,9 +249,14 @@ public class FlutterLocalNotificationsPlugin
           }
         }
 
+
+
         @SuppressLint("UnspecifiedImmutableFlag")
-        final PendingIntent actionPendingIntent =
+        PendingIntent actionPendingIntent =
             PendingIntent.getBroadcast(context, requestCode++, actionIntent, actionFlags);
+        if (action.launchApp == true) {
+          actionPendingIntent = PendingIntent.getActivity(context, requestCode++, intent, actionFlags);
+        }
 
         final Spannable actionTitleSpannable = new SpannableString(action.title);
         if (action.titleColor != null) {
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -249,13 +249,15 @@ public class FlutterLocalNotificationsPlugin
           }
         }
 
-
-
         @SuppressLint("UnspecifiedImmutableFlag")
         PendingIntent actionPendingIntent =
             PendingIntent.getBroadcast(context, requestCode++, actionIntent, actionFlags);
         if (action.launchApp == true) {
-          actionPendingIntent = PendingIntent.getActivity(context, requestCode++, intent, actionFlags);
+          String actionReferences =
+                  "{payload: "+notificationDetails.payload+", action_id: "+action.id+"}";
+          intent.putExtra(PAYLOAD, actionReferences);
+          actionPendingIntent = PendingIntent
+                  .getActivity(context, requestCode++, intent, actionFlags);
         }
 
         final Spannable actionTitleSpannable = new SpannableString(action.title);
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationAction.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationAction.java
@@ -82,6 +82,7 @@ public class NotificationAction {
   private static final String SHOWS_USER_INTERFACE = "showsUserInterface";
   private static final String ALLOW_GENERATED_REPLIES = "allowGeneratedReplies";
   private static final String CANCEL_NOTIFICATION = "cancelNotification";
+  private static final String LAUNCH_APP = "launchApp";
 
   public final String id;
   public final String title;
@@ -92,11 +93,13 @@ public class NotificationAction {
   @Nullable public final Boolean showsUserInterface;
   @Nullable public final Boolean allowGeneratedReplies;
   @Nullable public final IconSource iconSource;
+  @Nullable public final Boolean launchApp;
   public final List<NotificationActionInput> actionInputs = new ArrayList<>();
 
   public NotificationAction(Map<String, Object> arguments) {
     id = (String) arguments.get(ID);
     cancelNotification = (Boolean) arguments.get(CANCEL_NOTIFICATION);
+    launchApp = (Boolean) arguments.get(LAUNCH_APP);
     title = (String) arguments.get(TITLE);
 
     Integer a = (Integer) arguments.get(TITLE_COLOR_ALPHA);
--- a/flutter_local_notifications/example/lib/main.dart
+++ b/flutter_local_notifications/example/lib/main.dart
@@ -960,6 +960,7 @@ class _HomePageState extends State<HomePage> {
           'Action 2',
           titleColor: Color.fromARGB(255, 255, 0, 0),
           icon: DrawableResourceAndroidBitmap('secondary_icon'),
+          launchApp: true,
         ),
         AndroidNotificationAction(
           navigationActionId,
--- a/flutter_local_notifications/example/lib/navigation_event_listener.dart
+++ b/flutter_local_notifications/example/lib/navigation_event_listener.dart
@@ -28,7 +28,11 @@ class _NavigationEventListenerState extends State<NavigationEventListener> {
     super.initState();
 
     if (!IsolateNameServer.registerPortWithName(port.sendPort, portName)) {
-      throw Exception('Could not register SendPort');
+      // throw Exception('Could not register SendPort');
+
+      // Exception was conflicting when notification action was selected
+      // from the background.
+      debugPrint('WARNING: SendPort is already registered');
     }
 
     // ignore: avoid_annotating_with_dynamic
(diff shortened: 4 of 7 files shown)


MaikuB:
Looks like the PR takes the quick approach that you mentioned in https://github.com/MaikuB/flutter_local_notifications/issues/17#issuecomment-1038572190 that the payload is a combination of the notification id and action id. I believe the way actions have been implemented needs to pivot to what I described in https://github.com/MaikuB/flutter_local_notifications/issues/17#issuecomment-1039994533. Are you able to help with that? Even just doing part of it for one platform would help. From what you mentioned before, you also had to use another package to deal with handling notification action launching the app on iOS. I assume that is still the case and changing the APIs to behave like I described would remove the need to do so.

AdamBridges:
I will investigate implementing what you have suggested for Android but iOS will be the greater challenge for me, so if iOS is something you don't mind handling, then that would be greatly appreciated.

MaikuB:
Sure, this will require changes to platform interface as well. Would a PR to your fork work?

AdamBridges:
> Sure, this will require changes to platform interface as well. Would a PR to your fork work?
I am not sure. 

Upon investigating, though, it has occurred to me that the `SelectNotificationCallback` class is only called when the app has not been terminated, relying on the `NotificationAppLaunchDetails` class to handle any necessary logic a user might have, which seems redundant to me. That said, I think the ideal scenario is to lean on the `NotificationAppLaunchDetails` class and pass additional parameters in such as `NotificationActionDetails` and booleans that state whether the app was launched from a background or terminated state. 

With the above implemented, I think you can pretty much combine the `SelectNotificationCallback` class and `SelectNotificationActionCallback` class into one class to handle all background and terminated operations, yes?

MaikuB:
> Upon investigating, though, it has occurred to me that the SelectNotificationCallback class is only called when the app has not been terminated, relying on the NotificationAppLaunchDetails class to handle any necessary logic a user might have, which seems redundant to me.

Given you are comparing a type for a callback (`SelectNotificationCallback`) with a class that is passed to a callback (`NotificationAppLaunchDetails`), I wonder if you have some confusion. Neither of them rely on each other. `onSelectNotification` exists because that is the only way for apps to listen for when a user has tapped on a notification so that they can take users to another page in the app. This requires it to work when the app is the in foreground and can only work in that way. This requires "listening" to the callback to work. It's not possible to use this in the scenario where the notification has launched the app. That scenario would require calling the [getNotificationAppLaunchDetails](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/FlutterLocalNotificationsPlugin/getNotificationAppLaunchDetails.html) method to see if a notification launched the app. It's not possible to use the aforementioned callback as by the time you listen, the event has already happened and won't work with the flow on how Flutter apps startup. This is why `getNotificationAppLaunchDetails` exists so it can invoked on startup in the `main` function so apps can decide what the initial route is.  Both deal with different scenarios and there is a need for both. The FCM plugin works in a similar way except that it has a stream instead of a callback to handle when a notification has been tapped on. It also has a method to get the details of the notification that launched the app.

In terms of dealing with if an action launched an app, it could potentially be either a new method as I mentioned in https://github.com/MaikuB/flutter_local_notifications/issues/17#issuecomment-1039994533 or use `getNotificationAppLaunchDetails` and extend `NotificationAppLaunchDetails` to include details of the action.

> With the above implemented, I think you can pretty much combine the SelectNotificationCallback class and SelectNotificationActionCallback class into one class to handle all background and terminated operations, yes?

That isn't ideal as being able to handle terminated operations requires top-level function or a static method to be associated with the callback. If it is combined as you suggest then users of the plugin can no longer use a method scoped to the class where initialisation is done. If a user is trying to handle when a notification action should trigger navigation to another page, it requires them to put in more hoops so that the logic gets to a point where a `BuildContext` exists to use the navigator to push a page. This is compared to the current state where it's possible to have a `StatefulWidget`, call the `initalize` method within `initState` of the corresponding `State` class, listen to `onSelectNotification` and being able to have access to the `BuildContext` via the state's [context](https://api.flutter.dev/flutter/widgets/State/context.html) property. This is why I was proposing have different callbacks for dealing with when a notification action was tapped on in a terminated state and another callback for foreground notification actions where the action needs to have one property turned on to indicate it's meant for scenarios where the app is running or launch the app if it's not.

Hope that made it clearer. Might be worth taking another look at what I wrote in https://github.com/MaikuB/flutter_local_notifications/issues/17#issuecomment-1039994533

AdamBridges:
> Hope that made it clearer. Might be worth taking another look at what I wrote in [#17 (comment)](https://github.com/MaikuB/flutter_local_notifications/issues/17#issuecomment-1039994533)

I am not confused about what you had suggested in the comment; I was just suggesting an alternative as I thought that it might be more efficient but evidently not as I seem to be confused about the classes used. 

Anyway, in regards to [#17 (comment)](https://github.com/MaikuB/flutter_local_notifications/issues/17#issuecomment-1039994533), it just seems unnecessary to have another static function that basically will behave exactly like `onSelectNotification` with the only difference of including `NotificationActionDetails` when called. Is there some sort of benefit to excluding the notification payload? If not, perhaps just including a nullable `NotificationActionDetails` parameter in every `onSelectNotification` and `getNotificationAppLaunchDetails` call would be enough, and probably just rename `onSelectNotification` to `onSelectNotificationOrAction`. Thoughts?

MaikuB:
I wasn't proposing to have another static function. What I was proposing is that `onSelectNotificationAction` gets changed so could be used with methods scoped to the same class so it has access to a `BuildContext` so a notification action could trigger navigation to a different page in the app, much like how `onSelectNotification` works. A different callback would be used to deal with actions that just run logic in the background due to the requirements of it needing a top-level or static function. IMO what the plugin should enable with regards to notification actions is

1. the ability for the action trigger updates to the UI, which includes navigating to a page and launching the app if it's not running
2. the ability for an action to execute some logic without needing the app to be running. this is currently there thanks to the PR from @ened but this capability doesn't work for all platforms. It only works on Android and iOS so on those platforms, the callback is for this scenario but on Linux and macOS, it is used for the scenario 1. This is where having a separate callback is cleaner as this is the callback that requires a top-level or static function. Having separate requirements for the method that can be associated with the different platforms can cause confusion. FCM takes similar approach with the `onBackgroundMessage` stream/callback too. It would avoid conditionals in the code to do something like if it's Linux or macOS, emit to a stream that can be listened and process within the scope of where a `BuildContext` could be used for navigation and do something different for Android and iOS. 

>  Is there some sort of benefit to excluding the notification payload? If not, perhaps just including a nullable NotificationActionDetails parameter in every onSelectNotification and getNotificationAppLaunchDetails call would be enough, and probably just rename onSelectNotification to onSelectNotificationOrAction.

Might be a misunderstanding here but I wasn't proposing to have the notification payload excluded, `NotificationActionDetails` already contains the notification payload. I was referring to separate callbacks for tapping on a notification and tapping on a notification action. This was just my initial thoughts though so combining it into one callback could work though I would say there's still a need to deal with background notification actions  via a separate callback (i.e. scenario 2 mentioned above). Ignoring the background actions, mote that on iOS/macOS, it is done via one callback (see https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649501-usernotificationcenter?language=objc) but requires developers to put a bunch of if statements to look at the action identifier as even tapping on a notification associates that with a action identifier (see https://developer.apple.com/documentation/usernotifications/unnotificationdefaultactionidentifier?language=objc).  This leads me to think if one callback is used for processing notifications and notification actions then an enum to define the possible action types (1) tapping on a notification, (2) tapping on a custom notification action and (3) dismissing a notification. The plugin doesn't handle dismissal of notifications right now so that enum value for that could be added at a later date.

For what it's worth, I also took a look at plugins on other ecosystems for inspiration. One of them is a React one and it's from the same company that works on the Flutter Firebase plugins. They have a separate callbacks for foreground and background events and have an enum for different action types too. Here's a link to their docs https://notifee.app/react-native/docs/android/interaction if you'd like to take a look

AdamBridges:
> TL;DR to summarise the options I see for the APIs
> 
> a. There'd be separate callbacks for notification actions and notifications themselves:
> onSelectForegroundNotification onSelectForegroundNotificationAction, onSelectBackgroundNotificationAction.
> b. onDidReceiveForegroundNotificationResponse and onDidReceiveBackgroundNotificationResponse allow for dealing with notification action or notification being tapped on i.e. the same idea you had for onSelectNotificationOrAction (note: trying to avoid conjunction words where possible)

I think option b makes the most sense. I assume the function parameter for `SelectNotificationCallback` will replace String as it should only require `NotificationActionDetails` since it includes a notification's payload anyway? 

> in either case, getNotificationAppLaunchDetails() could be used to determine if a notification or notification action launched the app. As upon further thought, have a separate calls (e.g. getNotificationAppLaunchDetails() and a call to something like getNotificationActionAppLaunchDetails()) seems particularly awkward

Indeed. Simply replacing the `String` payload parameter with a `NotificationActionDetails` parameter should be all that's necessary.

Since the `NotificationActionDetails` class is going to take a larger role in providing data to notifications, perhaps it should be renamed to simply `NotificationReceivedDetails` or something like that, yes?

MaikuB:
> I think option b makes the most sense. I assume the function parameter for SelectNotificationCallback will replace String as it should only require NotificationActionDetails since it includes a notification's payload anyway?

Yep

> Since the NotificationActionDetails class is going to take a larger role in providing data to notifications, perhaps it should be renamed to simply NotificationReceivedDetails or something like that, yes?

Yep, ideally this should align with the method if possible so if it the method names used are like what I suggested earlier, this could be `NotificationResponse`. Then could rename `getNotificationAppLaunchDetails` to get `getLaunchNotificationResponse` for consistency

MaikuB:
Since we now seem to have an agreement on an approach, does this help unblock progress on this? What help would you need?

AdamBridges:
> Since we now seem to have an agreement on an approach, does this help unblock progress on this? What help would you need?

Good question. I have to figure out how method calls work, mainly. I was going to have more time to help you as I was waiting on some other dependencies for a project I am managing but those dependencies were just met this past week, so now I'm fairly busy. I'll try to make these changes before the end of March on the Android side of things but iOS is going to be lost on me, so I'll need you for that.

MaikuB:
Thanks for the reply. What I could do then is create a separate branch for this work and share the details when it's up

MaikuB:
Created this branch for the rework https://github.com/MaikuB/flutter_local_notifications/tree/10.0.0_actions_rework. Will close this PR as what's in here won't apply with the proposed changes

PR #1496, opened 2022-02-19
[flutter_local_notifications] iOS Time Sensitive notifications

### LINKED ISSUE #1495: Support for UNNotificationContent.interruptionLevel on iOS
Hello,

I was trying to see if I could send "time sensitive" and "critical" notifications on iOS using this plugin, but I did not find a way.

I understand that this should be controlled by `UNNotificationContent.interruptionLevel` as described here: https://developer.apple.com/documentation/usernotifications/unnotificationinterruptionlevel

And here:
https://developer.apple.com/design/human-interface-guidelines/ios/system-capabilities/notifications#helping-people-manage-notifications

Can you please confirm that this feature is indeed not implemented? If so, could I help with implementing it?

Great job on the plugin by the way, thank you! 

Cheers,
Marton

maprohu:
Hi!

Fixes #1495 

Note that I also added support for "critical" notifications simply because they appeared in the same enumeration in the API as "time sensitive" ones:

https://developer.apple.com/documentation/usernotifications/unnotificationinterruptionlevel?language=objc
https://developer.apple.com/documentation/usernotifications/unnotificationinterruptionlevel/unnotificationinterruptionlevelcritical?language=objc

However, I was not able to test how these work. I understand that it needs special approval from Apple, which I do not have:
https://developer.apple.com/contact/request/notifications-critical-alerts-entitlement/

Also note that this is my first time ever writing (seeing?) Objective-C code, i was mainly copy-pasting and adapting already existing code from the same file. I hope I did it right.

Probably also my first PR on Github. :)

Cheers,
Marton

### DIFF
--- a/flutter_local_notifications/lib/src/platform_specifics/ios/notification_details.dart
+++ b/flutter_local_notifications/lib/src/platform_specifics/ios/notification_details.dart
@@ -3,16 +3,17 @@ import 'notification_attachment.dart';
 /// Configures notification details specific to iOS.
 class IOSNotificationDetails {
   /// Constructs an instance of [IOSNotificationDetails].
-  const IOSNotificationDetails(
-      {this.presentAlert,
-      this.presentBadge,
-      this.presentSound,
-      this.sound,
-      this.badgeNumber,
-      this.attachments,
-      this.subtitle,
-      this.threadIdentifier,
-      this.interruptionLevel});
+  const IOSNotificationDetails({
+    this.presentAlert,
+    this.presentBadge,
+    this.presentSound,
+    this.sound,
+    this.badgeNumber,
+    this.attachments,
+    this.subtitle,
+    this.threadIdentifier,
+    this.interruptionLevel,
+  });
 
   /// Display an alert when the notification is triggered while app is
   /// in the foreground.
@@ -82,7 +83,6 @@ class IOSNotificationDetails {
 /// Constants that indicate the importance and delivery
 /// timing of a notification.
 enum InterruptionLevel {
-
   /// The system adds the notification to the notification
   /// list without lighting up the screen or playing a sound.
   passive,
--- a/flutter_local_notifications/lib/src/platform_specifics/ios/notification_details.dart
+++ b/flutter_local_notifications/lib/src/platform_specifics/ios/notification_details.dart
@@ -3,16 +3,16 @@ import 'notification_attachment.dart';
 /// Configures notification details specific to iOS.
 class IOSNotificationDetails {
   /// Constructs an instance of [IOSNotificationDetails].
-  const IOSNotificationDetails({
-    this.presentAlert,
-    this.presentBadge,
-    this.presentSound,
-    this.sound,
-    this.badgeNumber,
-    this.attachments,
-    this.subtitle,
-    this.threadIdentifier,
-  });
+  const IOSNotificationDetails(
+      {this.presentAlert,
+      this.presentBadge,
+      this.presentSound,
+      this.sound,
+      this.badgeNumber,
+      this.attachments,
+      this.subtitle,
+      this.threadIdentifier,
+      this.interruptionLevel});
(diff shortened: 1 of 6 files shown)


flutter_local_notifications/lib/src/platform_specifics/ios/notification_details.dart
```diff
@@ -71,4 +72,38 @@ class IOSNotificationDetails {
   ///
   /// This property is only applicable to iOS 10 or newer.
   final String? threadIdentifier;
+
+  /// The interruption level that indicates the priority and
+  /// delivery timing of a notification.
+  ///
+  /// This property is only applicable to iOS 10 or newer.
```
MaikuB, line 79: The comment says applicable for iOS 10 or newer but shouldn't that be iOS 15 or newer?

flutter_local_notifications/lib/src/platform_specifics/ios/notification_details.dart
```diff
@@ -71,4 +72,38 @@ class IOSNotificationDetails {
...
+  final InterruptionLevel? interruptionLevel;
```
MaikuB, line 80: As the plugin supports null-safety, what if this is were to be made non-nullable? My understanding is that `active` is the default interruption level to begin with. If that's the case, making it non-nullable and having `active` as default would be better

MaikuB:
Thanks for the PR, I left some comments. Another thing I noticed is this PR is specific to iOS. What about macOS? The notification APIs from Apple are same on both platforms so it would best to do this for macOS as well. There's a 10.0 pre-release version of the plugin at the moment (corresponding branch is https://github.com/MaikuB/flutter_local_notifications/tree/10.0.0) where I've done refactoring of the Dart classes so they are shared with iOS and macOS. It would be best if the PR could be done to both platforms by using that branch as the base for this work.

maprohu:
Okay, I am looking into the new branch (10.0.0) to see what it would take to add the changes to both iOS and macOS. 
First thing that I noticed is that the iOS plugin is implemented in Objective-C whereas the one for macOS is in Swift. 
As I am not very experienced with these, I just wanted to ask if there is a particular reason for the use of different languages, or it just happened to be like that?

maprohu:
I just tried to run the `flutter_local_notifications/example` app on macos (from branch 10.0.0). I did not see any notifications, whatever button I clicked. The same example app works fine on iOS Simulator, notifications do appear. Is there any particular setting that I may be missing to make notification appear in the macOS desktop app?

flutter_local_notifications/lib/src/platform_specifics/ios/notification_details.dart
maprohu: Correct. I wil fix that.

maprohu:
I added a new PR, replacing this one: #1501

MaikuB:
> As I am not very experienced with these, I just wanted to ask if there is a particular reason for the use of different languages, or it just happened to be like that?

That was because Objective C was the default language for Flutter plugins that target iOS back then and using Swift was known to cause issues back then. By the time macOS work had started, things became more stable and Swift became the default

> Is there any particular setting that I may be missing to make notification appear in the macOS desktop app?

Running it should've triggered a notification to accept permissions. It won't be a permissions dialog like on iOS. Perhaps you missed it but you should be able to go into system preferences to change the notification settings for each app and the example app should be one the apps shown there

flutter_local_notifications/lib/src/platform_specifics/ios/notification_details.dart
MaikuB: Sounds reasonable. I suppose it also mimics the fact that even if one were to write their app entirely in Objective C or Swift, that you can omit the interruption level too.

MaikuB:
Will close this since you've opened a new PR to avoid confusion and have answered all the questions :)

PR #1501, opened 2022-02-22
[flutter_local_notifications] adding ios/macos time sensitive interruption level

### LINKED ISSUE #1495: Support for UNNotificationContent.interruptionLevel on iOS

maprohu:
Fixes #1495 

Contains changes both for iOS and macOS, based on branch `10.0.0`

Please note my comment in the previous PR, I could not test the notifications on macOS yet. I did not really look into why no notifications show up on my Mac when running the `example` app. Do you happen to have a suggestion where to look?


maprohu:
I finally managed to test the notifications when running the `example` app as a macOS desktop application. As you suggested, I needed to enable the notifications in System Preferences. 

Note that in order for the time sensitive notifications to work, I had to add the "Time Sensitive Notifications" capability to the macOS project. And that required some change in the signing preferences of the xcode project, and I even ended up registering a new bundle ID for the example app in my Apple developer account. I did not commit these changes, as they are tied to my Apple account. Without making these changes in the xcode project the "time sensitive" setting on notifications are ignored and they behave like normal notifications.

flutter_local_notifications/lib/src/platform_specifics/darwin/notification_details.dart
```diff
@@ -84,4 +86,10 @@ class DarwinNotificationDetails {
   /// This must refer to a [DarwinNotificationCategory] identifier configured
   /// via [InitializationSettings].
   final String? categoryIdentifier;
+
+  /// The interruption level that indicates the priority and
+  /// delivery timing of a notification.
+  ///
+  /// This property is only applicable to iOS 15.0 or newer.
```
MaikuB, line 93: As this class is for both iOS and macOS, could you add the macOS version too?

MaikuB (changes requested):
Thanks for taking another stab at this. Noticed a couple of minor issues. Rest looks good to me :)

flutter_local_notifications/lib/src/platform_specifics/darwin/initialization_settings.dart
```diff
@@ -9,6 +9,7 @@ class DarwinInitializationSettings {
     this.requestAlertPermission = true,
     this.requestSoundPermission = true,
     this.requestBadgePermission = true,
+    this.requestCriticalPermission = true,
```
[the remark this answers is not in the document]
maprohu: Good point. I will change the default accordingly.

flutter_local_notifications/lib/src/platform_specifics/darwin/notification_details.dart
maprohu: Sure, adding comment.

MaikuB:
@maprohu in case it got missed, note that there are conflicts that would need to be resolved as there were changes done. You'll need to update your branch with the latest changes from 10.0.0 to try to resolve them

maprohu:
Hello Michael,
I just fixed the issues that you mentioned above.
Note that I see that there seem to be some conflicts in the branch. Let me know if that something that I should try to fix or you can handle it. 
Cheers,
Marton

MaikuB:
As this is your first PR, I would suggest you can try to see how to fix them as this will probably not be the only time that you'll be submitting PRs in the future :)

From what I could see, they should be relative straightforward to resolve. Main change to note is that recently I made the boolean parameters of the `requestPermissions` method non-nullable and with default values of `false`. That should help you determine what the resolution would be. If you get stuck though then give me a shout and I can help resolve them

maprohu:
@MaikuB 
I think I made a mistake and by accident reverted your changes regarding the `requestPermission` "false by default" parameter. Give me a second...

maprohu:
@MaikuB 
Ok, I think I made it. I do need to step up my merge game... :)

Let me know how it looks!

MaikuB (approved):

MaikuB:
Looks good to me. I noticed a comment around the ordering of the enum values that I may had missed, good call though made me think that another option is to assign specific values. Something that could be updated later though

MaikuB:
Hmm on further thought, I'll update how the enum is defined before releasing this. For reference, the approach will be similar to how the `Importance` class is defined. Thinking behind this is to make it less prone to errors if the ordering changes even though it's not likely and can help make it more explicit what the value is.

maprohu:
Ok, great, thanks! :)
About the enums, I was also in doubt. Unfortunately (or by design?) Dart enums are rather simple. It would be nice to have instance fields defined for each enum item. 
I agree, the way `Importance` is defined seems to be the next best thing.

tripany:
> I think breaks building for iOS < 15.
> 
> Property 'interruptionLevel' not found on object of type 'UNMutableNotificationContent *'

Scratch that, upgrading from xCode 12 to xCode 13 solved it.

PR #1559, opened 2022-04-11
[flutter_local_notifications] Throw ArgumentError when trying to initialize LocalNotification without  per Platform settings

talamaska:
change: throw ArgumentError when trying to initialize LocalNotification without  per Platform settings
add: tests
add: inline docs

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -307,6 +307,7 @@ await flutterLocalNotificationsPlugin.initialize(initializationSettings,
 ```
 
 Initialisation can be done in the `main` function of your application or can be done within the first page shown in your app. Developers can refer to the example app that has code for the initialising within the `main` function. The code above has been simplified for explaining the concepts. Here we have specified the default icon to use for notifications on Android (refer to the *Android setup* section) and designated the function (`selectNotification`) that should fire when a notification has been tapped on via the `onSelectNotification` callback. Specifying this callback is entirely optional but here it will trigger navigation to another page and display the payload associated with the notification.
+
 Note that all settings are nullable, because we don't want to force developers so specify settings for platforms they don't target. You will get a runtime ArgumentError Exception if you forgot to pass the settings for the platform you target.
 
 ```dart
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -307,7 +307,7 @@ await flutterLocalNotificationsPlugin.initialize(initializationSettings,
 ```
 
 Initialisation can be done in the `main` function of your application or can be done within the first page shown in your app. Developers can refer to the example app that has code for the initialising within the `main` function. The code above has been simplified for explaining the concepts. Here we have specified the default icon to use for notifications on Android (refer to the *Android setup* section) and designated the function (`selectNotification`) that should fire when a notification has been tapped on via the `onSelectNotification` callback. Specifying this callback is entirely optional but here it will trigger navigation to another page and display the payload associated with the notification.
-Note that all settings are nullable, because we don't want to force developers so specify settings fro platforms they don't target. You will get a runtime ArgumentError Exception if you forgot to pass the settings for the platform you target.
+Note that all settings are nullable, because we don't want to force developers so specify settings for platforms they don't target. You will get a runtime ArgumentError Exception if you forgot to pass the settings for the platform you target.
 
 ```dart
 void selectNotification(String payload) async {
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -307,6 +307,7 @@ await flutterLocalNotificationsPlugin.initialize(initializationSettings,
 ```
 
 Initialisation can be done in the `main` function of your application or can be done within the first page shown in your app. Developers can refer to the example app that has code for the initialising within the `main` function. The code above has been simplified for explaining the concepts. Here we have specified the default icon to use for notifications on Android (refer to the *Android setup* section) and designated the function (`selectNotification`) that should fire when a notification has been tapped on via the `onSelectNotification` callback. Specifying this callback is entirely optional but here it will trigger navigation to another page and display the payload associated with the notification.
+Note that all settings are nullable, because we don't want to force developers so specify settings fro platforms they don't target. You will get a runtime ArgumentError Exception if you forgot to pass the settings for the platform you target.
 
 ```dart
 void selectNotification(String payload) async {
--- a/flutter_local_notifications/example/integration_test/flutter_local_notifications_test.dart
+++ b/flutter_local_notifications/example/integration_test/flutter_local_notifications_test.dart
@@ -60,7 +60,7 @@ void main() {
         }
         if (Platform.isMacOS) {
           expect(e.message,
-              'MacOS settings must be set when targeting macOS platform.');
+              'macOS settings must be set when targeting macOS platform.');
         }
       }
     });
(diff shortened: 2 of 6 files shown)


talamaska:
The solution is not using asserts, because they are not flexible enough and also they force us to not use const constructors, which is undesirable. Throwing and ArgumentError is the right thing to do. The check will happen before the null check operator used on null value exception throws. Added more Readme and inline docs for the devs to make clearer.

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
```diff
@@ -118,6 +119,22 @@ class FlutterLocalNotificationsPlugin {
     if (kIsWeb) {
       return true;
     }
+    if (Platform.isAndroid && initializationSettings.android == null) {
```
MaikuB, line 122: The plugin already does platform checks down below so wouldn't it better to shift the null checks to happen within those if statements? This avoids checking the platform twice. Note that tests are also currently failing too as calls like `Platform.isAndroid` depends on the environment used to run the code. This is why the preferred way to do platform checks is via using `defaultTargetPlatform` as this can be overridden during tests. In case it helps, this is done by other plugins too (e.g. https://github.com/firebase/flutterfire/search?q=defaultTargetPlatform). By shifting the null checks to occur within the existing if statements, this should avoid the duplication I mentioned and fix the tests

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
```diff
@@ -118,6 +119,22 @@ class FlutterLocalNotificationsPlugin {
...
+      throw ArgumentError(
+          'iOS settings must be set when targeting iOS platform');
+    }
+    if (Platform.isLinux && initializationSettings.linux == null) {
+      throw ArgumentError(
+          'Linux settings must be set when targeting Linux platform');
+    }
+    if (Platform.isMacOS && initializationSettings.macOS == null) {
+      throw ArgumentError(
+          'MacOS settings must be set when targeting MacOS platform');
```
MaikuB, line 136: casing issue here the platform itself is referred to as "macOS" with a lowercase "m"

flutter_local_notifications/example/integration_test/flutter_local_notifications_test.dart
```diff
@@ -34,6 +34,35 @@ void main() {
...
+        if (Platform.isIOS) {
+          expect(e.message,
+              'iOS settings must be set when targeting iOS platform');
+        }
+        if (Platform.isLinux) {
+          expect(e.message,
+              'Linux settings must be set when targeting Linux platform');
+        }
+        if (Platform.isMacOS) {
+          expect(e.message,
```
MaikuB, line 61: casing issue, the platform itself is referred to as "macOS" with a lowercase "m"

flutter_local_notifications/README.md
```diff
@@ -307,6 +307,7 @@ await flutterLocalNotificationsPlugin.initialize(initializationSettings,
 ```
 
 Initialisation can be done in the `main` function of your application or can be done within the first page shown in your app. Developers can refer to the example app that has code for the initialising within the `main` function. The code above has been simplified for explaining the concepts. Here we have specified the default icon to use for notifications on Android (refer to the *Android setup* section) and designated the function (`selectNotification`) that should fire when a notification has been tapped on via the `onSelectNotification` callback. Specifying this callback is entirely optional but here it will trigger navigation to another page and display the payload associated with the notification.
+Note that all settings are nullable, because we don't want to force developers so specify settings for platforms they don't target. You will get a runtime ArgumentError Exception if you forgot to pass the settings for the platform you target.
```
MaikuB, line 310: Given how you wrote this note on its own line, did you intend for this be its own paragraph? If so then you're missing a line break to make this happen. Without the line break this sentence actually continues to be part of the same paragraph so can make it more confusing to match what's written "in code" with how it should appear

MaikuB (changes requested):
Thanks for the PR. I left some comments on the changes but also wanted to ask if you could clarify the following point

> The solution is not using asserts, because they are not flexible enough and also they force us to not use const constructors, which is undesirable. 

Whilst I don't have an issue with using `ArgumentError`, the point of const constructors doesn't apply to this PR. I assume you're referring to have code in the body of the constructor, if so wouldn't that happen for `ArgumentError`s too?  In actual fact, assertions are the one that **do** allow for const constructors. There a lot widgets within Flutter with const constructors and assertions. Perhaps you're not familiar with how this can be done but there's also an example within the repository https://github.com/MaikuB/flutter_local_notifications/blob/3b5f61d231cb70a6e57998cdedb52bd4569fb535/flutter_local_notifications/lib/src/types.dart#L42

talamaska:
> Thanks for the PR. I left some comments on the changes but also wanted to ask if you could clarify the following point
> 
> > The solution is not using asserts, because they are not flexible enough and also they force us to not use const constructors, which is undesirable.
> 
> Whilst I don't have an issue with using `ArgumentError`, the point of const constructors doesn't apply to this PR. I assume you're referring to have code in the body of the constructor, if so wouldn't that happen for `ArgumentError`s too? In actual fact, assertions are the one that **do** allow for const constructors. There a lot widgets within Flutter with const constructors and assertions. Perhaps you're not familiar with how this can be done but there's also an example within the repository
> 
> https://github.com/MaikuB/flutter_local_notifications/blob/3b5f61d231cb70a6e57998cdedb52bd4569fb535/flutter_local_notifications/lib/src/types.dart#L42

yes, but they do not refer to Platform.

MaikuB:
> yes, but they do not refer to Platform.

Not sure what you were trying to answer with this one but if it's to do with assertions then given my comments on the PR where the platform checks you added aren't needed as they already exist, suppose assertions were added then they would be done within the various if blocks for each platform to check to see if the corresponding settings isn't null.

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
talamaska: The first MacOs is the beginning of a sentence. As far I know English grammatic, you must start a sentence with a Capital letter.

flutter_local_notifications/example/integration_test/flutter_local_notifications_test.dart
talamaska: Not sure what you want to say here https://api.flutter.dev/flutter/package-platform_platform/Platform/isMacOS.html

MaikuB: For convenience, I'll reply to both of your responses related this topic here. First off is you are linking to the wrong doc. What you were using came from dart:io, see https://api.flutter.dev/flutter/dart-io/Platform/isMacOS.html. The rule of having first letter in uppercase for the first word in a sentence doesn't apply here as Apple has gone with lowercase "m" as part of they're branding. if you click through the link you will see that it in turn has a link to Wikipedia with macOS as the first word with lowercase "m" too. If you do a search for "macOS" then you should find other examples, particularly articles written that have it as the first word and respect the casing Apple has chosen due to branding too.

In other words, this will need lowercase "m" for all instances where it is mentioned. Note that even in your original PR you had "iOS" even when it was the first word. This is correct though not consistent with what you thought/point around casing rules

talamaska: fixed

MaikuB:
Thanks for making the changes. Been really so haven't been able to revisit this and I just noticed a minor formatting thing for API docs so I made the change directly here so this can go out soon

MaikuB (approved):

PR #1562, opened 2022-04-12
[flutter_local_notifications] Fix table of contents in README

HendrikF:
This fixes the table of contents at https://pub.dev/packages/flutter_local_notifications .

Android and iOS Setup do not work as the emoji breaks the link fragment.

Unfortunately I just realized, that GitHub seems to handle the #️Number Sign emoji differently, so it currently only works on GitHub.
With this fix it would work on pub.dev, but not on GitHub. :neutral_face:

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -19,12 +19,12 @@ A cross platform plugin for displaying local notifications.
    - [Linux limitations](#linux-limitations)
 - **[📷 Screenshots](#-screenshots)**
 - **[👏 Acknowledgements](#-acknowledgements)**
-- **[⚙️ Android Setup](#️-android-setup)**
+- **[🔧 Android Setup](#-android-setup)**
    - [Custom notification icons and sounds](#custom-notification-icons-and-sounds)
    - [Scheduled notifications](#scheduling-a-notification)
    - [Fullscreen intent notifications](#full-screen-intent-notifications)
    - [Release build configuration](#release-build-configuration)
-- **[⚙️ iOS setup](#️-ios-setup)**
+- **[🔧 iOS setup](#-ios-setup)**
    - [General setup](#general-setup)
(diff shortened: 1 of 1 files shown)


MaikuB:
Thanks for the PR and picking up on this. Maybe it's a GitHub issue but I'm having trouble making out what the changes are. Are you able to elaborate on what exactly changed. Here's what I see on GitHub so to my untrained eye, it looks the same though GitHub but there must be a difference if GitHub picked up on it

image

How did you validate it worked on pub.dev too? Am I also right in saying that getting rid of emojis would completely get rid of the issue? I would assume so as emojis was a contibrution from @psyanite and I hadn't received reports of issues besides incorrect links prior to their addition. If this is causing issues then I would say it's better to remove the emojis so that the community can actually use the table of contents on both pub.dev and GitHub

HendrikF:
Oh now that's interesting. This is what I am seeing:
Bildschirmfoto von 2022-04-18 14-05-36

I also see emojis using developer tools on pub.dev:
Bildschirmfoto von 2022-04-18 14-06-59
If I change the href using devtools, the links in the toc start working.

I am using Firefox 99 on Ubuntu 20.04.4 LTS.

MaikuB:
Hmm so it was a browser issue. I was using Safari and also checked Chrome on macOS. Given the inconsistent behaviour, do you think you'd be able to look into either removing the emojis or seeing if there's a different set that would be suitable but would work on pub.dev and GitHub?

HendrikF:
I don't understand why GitHub is generating the # emoji within the link fragment. As it only happens for headings, stating with the gear emoji, maybe we should just change that gear emoji? That way the two links would work on both platforms.

HendrikF:
I also opened a bug report with GitHub.

MaikuB:
Thanks for this. Hopefully, pub.dev won't be the one with an issue with this :D

PR #1571, opened 2022-04-20
[flutter_local_notifications] DST, function typo, header and newline consistency in readme

EnduringBeta:
My understanding is that the best way of referring to DST is not "daylight savingS" but "daylight saving", and I put in "time" for completeness for a readme. https://en.wikipedia.org/wiki/Daylight_saving_time

Some headers were level 4 and 5 after a level 2 header, so I bumped them up. I removed extra newlines and added missing ones since most of the doc uses the style of 2 newlines after a header.

---


MaikuB:
Thanks for the PR. I know DST is the more official term but at least with others I've spoken to, seems like it's common to refer it to daylight savings too but I'm not sure if that is just an Australian thing. Regardless, will merge it in.

EnduringBeta:
"Savings" is okay, you're right, but using the full, official term will avoid any disagreement. No worries either way.

PR #1593, opened 2022-05-12
[flutter_local_notifications] Update gson package for vulnerabilities

briannaAndCo:
Fixes CVE-2022-25647.

This updates an android jar to a version without the vulnerability. Vulnerability is considered HIGH and more info is [here](https://nvd.nist.gov/vuln/detail/CVE-2022-25647#range-7910512)

### DIFF
--- a/flutter_local_notifications/android/build.gradle
+++ b/flutter_local_notifications/android/build.gradle
@@ -36,6 +36,6 @@ android {
 dependencies {
     implementation "androidx.core:core:1.3.0"
     implementation "androidx.media:media:1.1.0"
-    implementation "com.google.code.gson:gson:2.8.6"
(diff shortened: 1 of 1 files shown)


MaikuB:
Thanks a lot for the PR. I noticed there's 2.9.0 that was out in February so curious if you know there's a why 2.9.0 shouldn't be used instead. Not following the development of gson but as you knew about this issue i thought I'd ask in case you might know

briannaAndCo:
I picked 1.8.9 because it's the first version without the vulnerability. And while I know jumping minor versions _shouldn't_ cause issues, it sometimes can in my experience, so it seemed safer to stay with the same minor version if possible.

MaikuB:
Thanks that was my guess but wanted to check

PR #1600, opened 2022-05-18
[flutter_local_notifications] Fix progress notification in example app of 10.0.0

lucasribolli:
Progress notification are not updating the same notification in example app as showed in the video. Using one id should fix it.

https://user-images.githubusercontent.com/42274899/169070560-7b1a5dc1-8f60-4f11-b1b8-8b0de23bc792.mp4


MaikuB:
Good spot. Thanks!

PR #1602, opened 2022-05-19
 [flutter_local_notifications_linux] Add support for the file path icons

### LINKED ISSUE #1565: AssetsLinuxIcon doesn't quite work as named
Flutter does not have access to the asset_path in the dart runtime. The rootBundle uses a binaryMessage system to fetch assets because as to it's mobile origins, native code was required to load them. This was kept up and as best I can tell, on desktop it's implemented probably in the autogenerated or included C code.

if the application is installed in a non standard way, this AssetsLinuxIcon doesn't work. We hacked my_application.cc for linux to support more natural linux installations of `assets` to /usr/share/appname/assets or ~/.local/share/appname/assets. While in app any user of AssetImage works in our app, using your AssetsLinuxIcon doesn't because it only gets a local relative path and only works when the assets dir is right in the executing directory hence the relative path happens to resolve.

I'm not sure what the proper solution is since we are doing slightly non standard behaviour but at a minimum you might want to rename the class from AssetsLinuxIcon since it isn't access flutter assets properly but just acting as a normal straight relative path, so it was a confusing error for a while until I understood what was happening.

On my end I'll prolly have to write my own detection code in dart so it can find the assets in one of the 3 or 4 locations we would install them and construct and absolute path to give to "AssetsLinuxIcon".

thanks! and sorry :D

proninyaroslav:
Adding the `FilePathLinuxIcon` class will allow the developer to specify a custom path in the file system to the icon, if other options, such as `AssetsLinuxIcon`, aren't suitable (for example, the assets path is different from the standard one).

Close #1565

### DIFF
--- a/flutter_local_notifications/example/lib/main.dart
+++ b/flutter_local_notifications/example/lib/main.dart
@@ -177,6 +177,9 @@ class HomePage extends StatefulWidget {
 }
 
 class _HomePageState extends State<HomePage> {
+  final TextEditingController _linuxIconPathController =
+      TextEditingController();
+
   @override
   void initState() {
     super.initState();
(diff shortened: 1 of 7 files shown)


MaikuB:
Thanks for updating the PR. Will review and test as soon as I can

MaikuB:
Thanks. LGTM but as mentioned in https://github.com/MaikuB/flutter_local_notifications/issues/1565, I'd like to see if @dballard could also give this a spin and provide feedback

dballard:
It appears good to me but I am struggling to test it in our app, I can adjust pubspec.lock to include a path to flutter_local_notifications but not flutter_local_notifications_linux because it's a dep by version of flutter_local_notifications.

But I think I now have code to use it, so if I still have problems I can always try following up with a PR now that the hard work is done! thanks!

MaikuB:
@dballard thanks for the update. I believe your issue could be solved using https://dart.dev/tools/pub/dependencies#dependency-overrides

dballard:
thanks @MaikuB  that did the trick!
can confirm this works for our use case!
thank you so much @proninyaroslav this is awesome!

PR #1603, opened 2022-05-19
[flutter_local_notifications] Update example

proninyaroslav:
Related to https://github.com/MaikuB/flutter_local_notifications/pull/1602


MaikuB:
Could you add this to the #1602 instead? Makes it easier to verify and can see that checks are currently failing. The plugin has switched to using [melos](https://melos.invertase.dev) a while back so it's possible for a monorepo to reference local dependencies

proninyaroslav:
@MaikuB 
Merged to #1602

PR #1609, opened 2022-05-27
[flutter_local_notifications] add AndroidNotificationCategory to make Android categories more accessible

lucasribolli:
Add all the Android categories values with documentation in AndroidNotificationCategory class from https://developer.android.com/reference/androidx/core/app/NotificationCompat.html#constants_2. 

The goal is make this property more accessible for the users. It results in a little breaking change since the current `category` property is of `String` type. However, I believe it’s worth using it. 

Also, I tried to make it just like LinuxNotificationCategory of Linux package to keep a pattern.

### DIFF
--- a/flutter_local_notifications/example/lib/main.dart
+++ b/flutter_local_notifications/example/lib/main.dart
@@ -1628,7 +1628,7 @@ class _HomePageState extends State<HomePage> {
     final AndroidNotificationDetails androidNotificationDetails =
         AndroidNotificationDetails('message channel id', 'message channel name',
             channelDescription: 'message channel description',
-            category: AndroidNotificationCategory.message(),
+            category: AndroidNotificationCategory.message,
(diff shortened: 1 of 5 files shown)


MaikuB:
Thanks for making the changes. Given what you said, do you think you could add links to to the Android API docs for each category? For example, https://developer.android.com/reference/androidx/core/app/NotificationCompat#CATEGORY_CALL() is the one for the call category

lucasribolli:
I add it, of course. It will be similar to [AndroidServiceForegroundType](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/lib/src/platform_specifics/android/enums.dart#L75).
The links need a `()` at the end so I will replace it by `%28%29` in encoded link. It works the same and does not add one more `)` in documentation from consumers vision

Issue #1519, opened 2022-03-07
Change AudioAttributes to the notifications in android 


bornold:
Hello @MaikuB 👋 
First of all, thank you so much for the plugin! Huge fan!

Have a use case where we would like to change [AudioAttributes](https://developer.android.com/reference/android/media/AudioAttributes) for the notifications here: https://github.com/MaikuB/flutter_local_notifications/blob/2c0b2db29fef939f1355cd708f8c5fee73deed22/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L953

Specifically to `AudioAttributes.USAGE_ALARM`

We are maintaining a fork so there is no problem for us just change the attribute, but I was wondering if there would be any interest in adding this to [AndroidNotificationDetails](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart)?

If so, I would be happy to write a PR

MaikuB:
A PR would be welcome but would like know what the thought would be on how users of the plugin specify the attribute? If you're looking to expand this then I suspect it makes  sense to allow any of the allowed attribute values to be specified then for backwards compatibility, assume the default value is `AudioAttributes.USAGE_NOTIFICATION`

bornold:
Cool!
Yes, the user will specify the attribute in [AndroidNotificationDetails](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart)
as a [enum](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/lib/src/platform_specifics/android/enums.dart) that defaults to `AudioAttributes.USAGE_NOTIFICATION`.

Even though it does not really make any sense to specify some of the attributes available in [`AudioAttributes.Builder.setUsage` ](https://developer.android.com/reference/android/media/AudioAttributes.Builder#setUsage(int)) for a notification I believe it's best to include them all.


Guess I'll start working on a PR that in the coming weeks branching of [10.0.0](https://github.com/MaikuB/flutter_local_notifications/tree/10.0.0)?

MaikuB:
This sounds like something that wouldn't be a breaking change so if that's the case, it could use master as the target branch so that this goes out in a 9.x release

PR #1649, opened 2022-07-15
[flutter_local_notifications]  Add audioAttributesUsage to NotificationDetails

bornold:
Allows users to change  AudioAttributes (https://developer.android.com/reference/android/media/AudioAttributes) to the notification sound

Closes: https://github.com/MaikuB/flutter_local_notifications/issues/1519

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -972,7 +972,7 @@ public class FlutterLocalNotificationsPlugin
       notificationChannel.setGroup(notificationChannelDetails.groupId);
       if (notificationChannelDetails.playSound) {
         AudioAttributes audioAttributes =
-            new AudioAttributes.Builder().setUsage(AudioAttributes.USAGE_NOTIFICATION).build();
+            new AudioAttributes.Builder().setUsage(notificationChannelDetails.audioAttributesUsage).build();
         Uri uri =
             retrieveSoundResourceUri(
                 context, notificationChannelDetails.sound, notificationChannelDetails.soundSource);
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -971,8 +971,11 @@ public class FlutterLocalNotificationsPlugin
       notificationChannel.setDescription(notificationChannelDetails.description);
       notificationChannel.setGroup(notificationChannelDetails.groupId);
       if (notificationChannelDetails.playSound) {
+        Integer audioAttributesUsage = notificationChannelDetails.audioAttributesUsage != null
+                ? notificationChannelDetails.audioAttributesUsage
+                : AudioAttributes.USAGE_NOTIFICATION;
         AudioAttributes audioAttributes =
-            new AudioAttributes.Builder().setUsage(notificationChannelDetails.audioAttributesUsage).build();
+                new AudioAttributes.Builder().setUsage(audioAttributesUsage).build();
         Uri uri =
             retrieveSoundResourceUri(
                 context, notificationChannelDetails.sound, notificationChannelDetails.soundSource);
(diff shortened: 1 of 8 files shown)


MaikuB (changes requested):
Thanks for the PR. It looks like the code is missing some logic to handle how there could be scheduled notification that would have values for the new fields added on the Java/Android side.  This means changes need to be added to fallback to say the audio usage is for notifications. Without this, what would happen is a scheduled notification is deserialised from shared preferences and a NPE would occur

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/models/NotificationDetails.java
```diff
@@ -268,6 +271,10 @@ private static void readPlatformSpecifics(
       notificationDetails.tag = (String) platformChannelSpecifics.get(TAG);
       notificationDetails.colorized = (Boolean) platformChannelSpecifics.get(COLORIZED);
       notificationDetails.number = (Integer) platformChannelSpecifics.get(NUMBER);
+      notificationDetails.audioAttributesUsage = (Integer) platformChannelSpecifics.get(AUDIO_ATTRIBUTES_USAGE);
+      if (notificationDetails.audioAttributesUsage == null) {
+        notificationDetails.audioAttributesUsage = AudioAttributes.USAGE_NOTIFICATION;
+      }
```
bornold, line 277: Not sure if this should null checks or not, can revert if it is unnecessary

MaikuB: The null checks you've added aren't necessary as the Dart/Flutter API won't allow this to happen. The issue I mentioned in my previous message can still occur so this can be reverted. I'll point out exactly where the issue I mentioned can be resolved

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -972,7 +972,7 @@ private static void setupNotificationChannel(
       notificationChannel.setGroup(notificationChannelDetails.groupId);
       if (notificationChannelDetails.playSound) {
         AudioAttributes audioAttributes =
-            new AudioAttributes.Builder().setUsage(AudioAttributes.USAGE_NOTIFICATION).build();
+            new AudioAttributes.Builder().setUsage(notificationChannelDetails.audioAttributesUsage).build();
```
MaikuB, line 975: This is the place to coalesce a null value from a scheduled notification to `AudioAttributes.USAGE_NOTIFICATION`

MaikuB (changes requested):

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -971,8 +971,11 @@ private static void setupNotificationChannel(
       notificationChannel.setDescription(notificationChannelDetails.description);
       notificationChannel.setGroup(notificationChannelDetails.groupId);
       if (notificationChannelDetails.playSound) {
+        Integer audioAttributesUsage = notificationChannelDetails.audioAttributesUsage != null
```
MaikuB, line 974: As I mentioned coalescing earlier but wasn't done, perhaps you're not familiar with what that means. Rather than having

```
Integer audioAttributesUsage = notificationChannelDetails.audioAttributesUsage != null
                 ? notificationChannelDetails.audioAttributesUsage
                 : AudioAttributes.USAGE_NOTIFICATION;
```

this can be expressed as

```
Integer audioAttributesUsage = notificationChannelDetails.audioAttributesUsage ?? AudioAttributes.USAGE_NOTIFICATION;
```

MaikuB: Apologies, I thought Java supported this but turns out it doesn't

flutter_local_notifications/lib/src/platform_specifics/android/enums.dart
```diff
@@ -262,3 +262,82 @@ enum NotificationVisibility {
...
+  static List<AudioAttributesUsage> get values => <AudioAttributesUsage>[
+        alarm,
+        assistanceAccessibility,
+        assistanceNavigationGuidance,
+        assistanceSonification,
+        assistant,
+        game,
+        media,
+        notification,
+        notificationEvent,
```
MaikuB, line 328: Was about to merge and then realised `notificationRingtone` is missing here. Are you able to add this in?

MaikuB (approved):

PR #1650, opened 2022-07-17
[flutter_local_notifications] WIP Custom intervals for periodicallyShow

micimize:
Changes `repeatInterval` in `periodicallyShow` to a `Duration` and treats older apple platforms without custom duration support as a special case. Also turned `RepeatInterval` into a class extension with static methods for backwards compatibility.

I haven't added assertions to enforce older platforms are using the right intervals or changed the docs yet, or tested much beyond iOS simulator – wanted to get feedback on the overall approach first.


MaikuB:
Thanks for the PR. It's a clever way of handling it so props for that :)

It looks like would cause an issue on Android with old scheduled notifications saved to shared preferences where the value for the repeat interval represents the index associated with the enum value. With the code changes done in the PR, those old notifications would have its repeat interval be interpreted as a value in milliseconds

micimize:
Ah I see – yes that could be a problem. I could also see there being a similar issue on iOS with old enum values cached.

Not sure what the best way of dealing with this is – adding legacy handling feels awkward, but I can't find any docs/utilities for doing at-update-time migrations/scripts for flutter.

MaikuB:
I'd suggest a new method to deal with this. It would also be cleaner in the sense in that there's a separate method that only works from certain versions of iOS/macOS. Coming up with a name for the new method could be difficult though given Dart doesn't support method/function overloading to make use of the existing name

neelayak:
Hey @MaikuB when can we expect this update as this will be a game changer for certain use cases?

PR #1658, opened 2022-07-30
[flutter_local_notifications] Add support for Android Tiramisu' `POST_NOTIFICATIONS` permission

### LINKED ISSUE #1597: Support for Android 13 Tiramisu (API 33)
[Android 13 introduces new runtime permission for notifications.](https://developer.android.com/about/versions/13/changes/notification-permission)

I've looked but haven't found any mention about it in the package's README on pub.dev.

So I suggest that `AndroidFlutterPluginImplementation` should get `requestPermissions` method, similar to [iOSFlutterPluginImplementation](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/IOSFlutterLocalNotificationsPlugin/requestPermissions.html)

bartekpacia:
Implements two new methods on `AndroidFlutterLocalNotificationsPlugin`, `requestPermission` and `isPermissionGranted`, to let developers work with the new [notification runtime permission](https://developer.android.com/about/versions/13/changes/notification-permission) added in Android 13 Tiramisu (API level 33)

Fixes #1597


flutter_local_notifications/android/build.gradle
```diff
@@ -31,6 +31,10 @@ android {
     lintOptions {
         disable 'InvalidPackage'
     }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
```
bartekpacia, line 37: This was automatically added my IDE (IntelliJ). Should I remove it?

bartekpacia:
I don't see why the Android integration tests are failing.

bartekpacia:
What's missing is returning a boolean result from `requestPermission()`. I'll implement this as well in this PR :)

MaikuB:
Thanks for the PR, was going to comment that it looks like that was missing when it comes to dealing with the listener.

Also the the Java code from the `isPermissionGranted` method remains after you had removed the code on the Flutter/Dart side.

I'll take a closer look once the changes are done

MaikuB:
Also just remembered about https://github.com/flutter/flutter/issues/12561 that can impact testing. A workaround is mentioned there

bartekpacia:
@MaikuB Now `AndroidFlutterLocalNotificationsPlugin.requestPermission()` returns a `Future<bool?`> :)

flutter_local_notifications/example/lib/main.dart
```diff
@@ -183,12 +182,11 @@ class _HomePageState extends State<HomePage> {
   @override
   void initState() {
     super.initState();
-    _requestPermissions();
```
MaikuB, line 186: Can this be kept so it request permissions so it's essentially done on startup? This was originally done as helps those using the example app to get started with testing the functionality to display notifications immediately

flutter_local_notifications/example/lib/main.dart
```diff
@@ -183,12 +182,11 @@ class _HomePageState extends State<HomePage> {
...
     _configureDidReceiveLocalNotificationSubject();
     _configureSelectNotificationSubject();
   }
 
-  void _requestPermissions() {
+  void _requestIOSPermissions() {
```
MaikuB, line 189: Based on what I mentioned above, it this can be invoked on startup then the old name can be kept. Note that the name of `_requestIOSPermissions()` is incorrect as it also handles requesting permissions on macOS

flutter_local_notifications/example/lib/main.dart
```diff
@@ -207,6 +205,21 @@ class _HomePageState extends State<HomePage> {
...
   }
 
+  Future<void> _requestAndroidPermissions() async {
+    final AndroidFlutterLocalNotificationsPlugin? androidImplementation =
+        flutterLocalNotificationsPlugin.resolvePlatformSpecificImplementation<
+            AndroidFlutterLocalNotificationsPlugin>();
+
+    final bool? granted = await androidImplementation?.requestPermission();
+
+    await showDialog<void>(
```
MaikuB, line 215: If the intent behind having a separate button to request permissions was done to display the result, if the example app can be updated to do this on startup via `initState` then an extra label could be added to the example to display the result. By that I mean like a `bool` field added and `setState()` can be called to save the result of the call and update the label that indicates if permissions were granted on Android

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -1217,40 +1228,31 @@ public void onDetachedFromActivity() {
...
+        break;
       case ZONED_SCHEDULE_METHOD:
-        {
-          zonedSchedule(call, result);
-          break;
-        }
+        zonedSchedule(call, result);
+        break;
+      case REQUEST_PERMISSION_METHOD:
+        requestPermission(result::success);
```
MaikuB, line 1249: I've saw that multiple calls to this method can result in a crash. To reproduce this, make two consecutive calls from the Flutter side and tap on "Allow" to grant permissions. App will crash and can see logs on requesting multiple permissions, reply already received. This can be avoided by limiting the plugin to only handle one permissions request at a time and return an error back to consumers if there are concurrent calls. Examples I've seen on how to handle this can be found in plugins like `google_sign_in` (see https://github.com/flutter/plugins/blob/e8a7abcd654f53d9111cb273dd24a7a7c5974b7a/packages/google_sign_in/google_sign_in_android/android/src/main/java/io/flutter/plugins/googlesignin/GoogleSignInPlugin.java#L284, I used this approach in `flutter_appauth`) and `permission_handler` (see https://github.com/Baseflow/flutter-permission-handler/blob/051828f0c5b572fde65ef3474af0de29407fe9aa/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java#L210)

MaikuB (changes requested):
Thanks for making the changes. I've left some comments on this, hope they make sense and happy to discuss further if required

bartekpacia:
@MaikuB Thanks for review, I'll make the changes soon :)

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
bartekpacia: Thanks, I'll look into this during the weekend.

MaikuB: It's probably becuase you were busy that this hasn't been addressed yet but in case it's because you forgot, a reminder that this issue remains

bartekpacia: Yes, I want to do this today. Thank you :)

bartekpacia: Done.

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -1520,6 +1523,44 @@ private void cancelAllNotifications(Result result) {
     result.success(null);
   }
 
+  public void requestPermission(@NonNull PermissionRequestListener callback) {
+    if (permissionRequestInProgress) {
+      callback.complete(false);
```
MaikuB, line 1528: Looks like this would complete and indicate to users of the API that the permission wasn't granted. This could be misleading as apps might think the user denied the permissions and the app updates the UI to present some text around this. IMO it would be better to return an error so apps can choose treat this differently e.g. swallow the error. The approach you're using here is similar to https://github.com/Baseflow/flutter-permission-handler/blob/051828f0c5b572fde65ef3474af0de29407fe9aa/permission_handler_android/android/src/main/java/com/baseflow/permissionhandler/PermissionManager.java#L210 so triggering an error callback could be done to solve this

MaikuB (changes requested):

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
bartekpacia: You're right, thank you. I fixed this.

MaikuB:
Thanks for making the change. I pushed a grammar fix for the error message you added in so I can try to merge sooner. One thing that this made me think about though is the ability to use the existing `initialize()` method to also request permissions too. This is something that's possible for iOS and macOS. I can try to look at that after this PR is merged

MaikuB (approved):

bartekpacia:
Thank you for your help with getting this merged @MaikuB :)

Regarding your idea with requesting permissions on `initialize()`, you have the example app in mind, not the plugin itself, right?

MaikuB:
No I'm referring to the plugin itself

stephanie-finch:
Thanks both for working on this! 👏🏻

Weighing in from the product side for requesting on `initialize()`, it's often important to prompt the user for notif permissions at a very specific step in the new user flow to optimize conversion for accepting. So I'd wanna make sure the plugin doesn't automatically prompt users for permission in a way that isn't within our control.

MaikuB:
If you look at how iOS/macOS works then there is control over that. This should be clearer if you look at the example app where it decides not to have permissions requested when calling `initialize`

PR #1659, opened 2022-07-30
[flutter_local_notifications] Update Gradle and other dependencies

bartekpacia:
It's always good to be on the newest Gradle & AGP:)

cc: @MaikuB

### DIFF
--- a/flutter_local_notifications/android/build.gradle
+++ b/flutter_local_notifications/android/build.gradle
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.3'
+        classpath 'com.android.tools.build:gradle:7.2.1'
     }
 }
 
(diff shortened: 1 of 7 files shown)


MaikuB:
I've been holding off on this to make sure the other PR for permissions is merged first. I'm inclined to think this shouldn't be merged in, at least not for Gradle and AGP changes. Whilst updating dependencies is good practice for apps, there have been issues with Flutter working with newer versions of Gradle/AGP. Having this plugin on newer Gradle/AGP versions could cause issues for those on older versions of Flutter as the tooling may not work with them. A number of official plugins are point to older versions too and that's most likely for the same reason

bartekpacia:
> Whilst updating dependencies is good practice for apps, there have been issues with Flutter working with newer versions of Gradle/AGP.

Could you give me some examples of such problems? I agree that when the new version of Gradle comes out, it's not immediately supported by Flutter, but [7.5 is supported](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/android/gradle_utils.dart#L166-L211). I maintain a package myself ([flutter_downloader](https://github.com/fluttercommunity/flutter_downloader)) and it's on the newest AGP and Gradle and my users have no problems with this configuration.

bartekpacia:
Thanks for this discussion. You're right that it doesn't make sense to bump Gradle now for this plugin. So I'm going close this PR.

PR #1668, opened 2022-08-09
[flutter_local_notifications] Adding calledAt parameter to periodicallyShow method

timurturbil:
When setting periodically scheduled notification, we may need to specify when start to call it. For example, Today is Monday but I want to set the periodically notification on Sunday to repeat on every Sunday.


MaikuB:
Thanks for the PR but will not look to merge this in as it only works on Android given makes use of an internal value (`calledAt`) that was specific to Android. `calledAt` exists so the plugin can know when the reschedule a notification after one has been fired and to try to be as precise with timing as possible as using platform channels on Android seems to have more a delay

PR #1686, opened 2022-09-01
[flutter_local_notifications] Check permissionRequestInProgress in onRequestPermissionsResult

nvx:
Other flutter plugins permissions requests can end up triggering onRequestPermissionsResult, so only handle callbacks when we're expecting one to be outstanding.

This fixes a `java.lang.IllegalStateException: Reply already submitted` crash that can occur when other flutter plugins request access to permissions on Android.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -1574,7 +1574,7 @@ public class FlutterLocalNotificationsPlugin
   @Override
   public boolean onRequestPermissionsResult(
       int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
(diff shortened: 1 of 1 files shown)


MaikuB:
Thanks for the PR. Do you by chance have details on how to reproduce the issue, preferably in the form of a repo hosting an app that can reproduce the issue? Asking as it looks related to https://github.com/MaikuB/flutter_local_notifications/issues/1688 but the exception is different. I've only been able to reproduce this if the permission requests aren't awaited but the error is different than what you reported. Furthermore, while the PR fixes the aforementioned issue, other plugins may still crash if the calls aren't awaited as awaiting avoid request codes from various plugins clashing

MaikuB:
Although I haven't heard of a response yet, will merge this in anyway as it definitely fixes an issue when multiple plugins handle permission requests

nvx:
I haven't got a publicly available demo, but the other plugin I was using is `flutter_local_notifications` which looking at it's code has a similar check to only handle the callback when it's got an outstanding permission request.

Cheers for merging the fix though!

PR #1706, opened 2022-09-17
[flutter_local_notifications] Update timezone

nohli:
This updates the `timezone` package.


MaikuB:
Thanks for the PR. Shame I didn't spot this before given I just pushed 10.0 and it looks like the timezone repository doesn't publish releases to make it easier to monitor for releases. I'll need to push this out as part of 11.0 to be safe as `timezone` going from 0.8.x to 0.9.0 is considered a breaking change

nohli:
@MaikuB thanks for merging that quickly, so no `dependency_overrides` were necessary for releasing a new app version and using the latest dependencies 🙂 

Yeah, technically `timezone` had a breaking change because its major version is <1, so you did everything right ❤️

PR #1714, opened 2022-09-22
Crash fix when notification actions is used for foreground service 

arnoldlaishram:
**Issue -** 
When NotificationAction is used for foreground service, getting the below error.
```
Failed to handle method call
Unhandled Exception: PlatformException(error, Parcelable encountered IOException writing serializable 
object (name = com.dexterous.flutterlocalnotifications.ForegroundServiceStartParameter), null, java.lang.RuntimeException: Parcelable encountered IOException writing serializable 
object (name = com.dexterous.flutterlocalnotifications.ForegroundServiceStartParameter
```

Upon debugging I found that NotificationAction was not implementing Serializable.

**Fix -** 

Making NotificationAction Serializable to support  notification actions when notification is started as foreground service


MaikuB:
Thanks!

Issue #1730, opened 2022-09-30
android: FlutterLocalNotificationsPlugin.initialize() in separate Isolate


emersion:
**Describe the bug**
Since a few versions, it seems like opening a notification in a separate Isolate will break the notification callbacks.

**To Reproduce**
1. Call `FlutterLocalNotificationsPlugin.initialize()` in main Isolate
2. Call `FlutterLocalNotificationsPlugin.initialize()` in other Isolate (e.g. Firebase messaging Isolate), show a notification from there
3. Tap on notification while the app is still open
4. Nothing happens, `onDidReceiveNotificationResponse` is not invoked

**Expected behavior**
`onDidReceiveNotificationResponse` should be invoked since the app is still in foreground.

Note, everything still works fine if the app is closed before the notification is tapped.

My theory is that the `FlutterLocalNotificationsPlugin.initialize()` call in the other Isolate will override the one done previously in the main Isolate. Shortly after, the other Isolate is stopped, and `flutter_local_notifications` no longer has a hold on the Isolate to which the callbacks should be sent. Sadly I don't think it's possible to open a notification without a call to `FlutterLocalNotificationsPlugin.initialize()`?

**Sample code to reproduce the problem**
Latest [Goguma](https://sr.ht/~emersion/goguma) APK when connected to a soju server.

MaikuB:
Do you mean to say it worked before? If so, besides changing the name and signature of the callback, the way it works has remained the same. I suspect what you're seeing is more related to https://github.com/flutter/flutter/issues/23904. I was looking to move away the callback to be more like a stream and have it hang off the plugin class instead i.e. it's not associated with the `initialize()` method. Not sure if this would solve your issue though. Would you be willing help submit a PR to help with these changes?

emersion:
> Do you mean to say it worked before?

Hmm, now that you mention it, I'm not sure it ever worked, maybe I always killed the app in my testing. Another Goguma developer also reports that it seemingly never worked. They also report that the callback gets called, but in the Firebase Isolate, not in the main Isolate. This doesn't match my experience, but I've been testing with UnifiedPush instead of Firebase. I'll investigate some more.

> I was looking to move away the callback to be more like a stream and have it hang off the plugin class instead i.e. it's not associated with the initialize() method.

This would be a purely cosmetic change though AFAIU?

emersion:
Would it be reasonable to:

[...]

PR #1744, opened 2022-10-08
[flutter_local_notifications] Skip setMethodCallHandler() on null callback

emersion:
If the user didn't provide a onDidReceiveNotificationResponse callback, there's no point in listening to method calls coming from the platform.

This is useful when using the plugin from a separate Isolate. The setMethodCallHandler() call is overwriting the main Isolate's callback without this patch.

See discussion in https://github.com/MaikuB/flutter_local_notifications/issues/1730

### DIFF
--- a/flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
+++ b/flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
@@ -140,8 +140,10 @@ class AndroidFlutterLocalNotificationsPlugin
     DidReceiveBackgroundNotificationResponseCallback?
         onDidReceiveBackgroundNotificationResponse,
   }) async {
-    _ondidReceiveNotificationResponse = onDidReceiveNotificationResponse;
-    _channel.setMethodCallHandler(_handleMethod);
+    if (onDidReceiveNotificationResponse != null) {
+      _ondidReceiveNotificationResponse = onDidReceiveNotificationResponse;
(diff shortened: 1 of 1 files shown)


MaikuB:
Thanks for the PR. I'll need to see if this will still work with the changes I'm planning to do that were mentioned at https://github.com/MaikuB/flutter_local_notifications/issues/1730#issuecomment-1266743806.  Conscious of adding this and this and it ends up being removed. This won't necessarily be just a cosmestic change as it would change from being a callback to a stream. This would allow apps to listen to notification events anywhere as the event would be broadcasted. This would be similar to how the FCM plugin works

MaikuB:
Apologies for late response. I don't believe this will work when the changes I mention are done as the broadcast stream controller would be a static property hanging off the `FlutterLocalNotificationsPlugin` class. This is so it's not tied the `initialize()` method. Not sure there's a way to handle this when it lands. A couple of other things

- the plugin calls `setMethodCallHandler()`  without checking if a callback has provided issimilar to how other plugins handle this so would suggest the problem is more widespread amongst the community
- do you know if this problem also occurs when the callback is to invoke a method on a background isolate? That is, does this also happen with the `onDidReceiveBackgroundNotificationResponse` callback. I would think it doesn't as different mechanism are used

emersion:
Yeah, that makes sense. You might want to consider to make `FlutterLocalNotificationsPlugin` a singleton since the `MethodChannel` is static.

PR #1745, opened 2022-10-08
[flutter_local_notifications] Drop registerWith in Android plugin

emersion:
registerWith() is a relic from Flutter embedding v1.


MaikuB:
Thanks for the PR. I'll check but wanted to ask if you know if this is actually safe to drop. Plugins have needed to support this for some time now as there could be apps that haven't migrated to v2 embedding. A lot of plugins I've seen still have this as well

MaikuB:
Checked with older SDKs by creating a new plugin and saw that the template no longer references v1 embedding so should be good. Thanks for the PR :)

Issue #791, opened 2020-09-06
Timer strategy


zommerberg:
Assuming I am building a timer / countdown that should be shown as a notification every second what is the best strategy to do so?
Is it possible to run it in the background?
Do I need to dispatch a new notification every second or I can update the already existing notification?

MaikuB:
The plugin allows you update an existing notifications. As for the strategy for your use cases, these are best asked on places like Stack Overflow and/or you'll need to research what's supported by the native APIs.

Issue #306, opened 2019-09-04
Feature request: Notification showing a Countdown timer


dpedrinha:
Would be very nice to have a notification showing a countdown timer like the new android system does.

If you start a the system countdown timer now it places a sticky notification with the countdown.

So it would be a notification that takes a Duration and shows a countdown timer once it is started. When it ends, it rings and vibrate and start counting forwards in red so the user knows for how long the timer have been set off. And of course, by this point, the notification should have a way to be dismissed.

Cheers and great job with this plugin.

MaikuB:
Thanks for the kind words. What you're asking for is very specific to an app IMO (I believe this will only work on Android as well) and isn't something I intend to work. As such I'll be closing this I'd suggest you look at using platform channels to implement this yourself for your app

ghenry:
I was looking to do this too like MS Teams does in their Android app. Android specific I think?

PR #1778, opened 2022-11-09
[flutter_local_notifications] Fixes #1486 :Adds ability to count down chronometer with chronometerCountDown

### LINKED ISSUE #1486: Timer in notification
Can I with this package leave a timer running on the notification like the android's native timer?
And when accessing the notification, open the app at the time set in the notification timer?

ebelevics:
Fixes https://github.com/MaikuB/flutter_local_notifications/issues/1486, https://github.com/MaikuB/flutter_local_notifications/issues/791, https://github.com/MaikuB/flutter_local_notifications/issues/306

**chronometerCountDown** will only work for Android devices that are 7.0 or newer. But this will do the trick for updating timer without updating notification every single second.
```dart
AndroidNotificationDetails(
          _channelKey,
          'Channel',
          channelDescription: 'Description',
          when: endAt.millisecondsSinceEpoch, // should be greater than DateTime.now()
          timeoutAfter: endAt.difference(DateTime.now()).inMilliseconds, // this will ensure that notification will hide after reaching 0:00
          usesChronometer: true,
          chronometerCountDown: true, // chronometer will go backwards from positive to negative
        ),
```


ebelevics:
Noticed typo at notification_details.dart. Should be `/// This property is only applicable to Android 7.0 and newer versions.`

MaikuB:
Thanks for the PR. Could you push an update to fix the issue in the API docs that you had mentioned in your previous message?

ebelevics:
Api doc corrected

MaikuB:
Thanks for the update. The unit tests are failing due to the updates made in this PR, could you look to address this as well?

ebelevics:
Sorry for delay, done.

Issue #1610, opened 2022-05-28
[macOS] Example not working


Fraa-124:
Hello, I tested it on android & ios and working fine, but I didn't have success running the example provided in this repository on MacOS. Nothing broke but the notifications didn't showed up. I don't know if it is really broke or I'm missing something.

MaikuB:
In my experience, this is something that happens when permissions haven't been granted. Perhaps you had dismissed it when it originally appeared or it didn't appear for some reason. Note the permissions prompt on macOS appears as a notification so it's easy to have missed or dismissed it. If you have trouble getting it to reappear after cleaning, re-running etc then can try granting the app permissions via macOS's system preferences

Fraa-124:
Yep, you are right, The system was blocking notifications, I must have missed the request at the beginning

PR #1785, opened 2022-11-14
[flutter_local_notifications] Implement options to hide or crop attachments in the thumbnail on iOS

noinskit:
This PR implements access to two platform-specific features for customizing attachments to notifications (c.f. https://developer.apple.com/documentation/usernotifications/unnotificationattachment?language=objc). Without them it's very hard to use image attachments at all, because by default the first attachment is used both for a tiny thumbnail (smaller than app logo) and a potentially full screen image on an expanded notification.

### DIFF
--- a/flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
+++ b/flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
@@ -67,6 +67,8 @@ NSString *const SOUND = @"sound";
 NSString *const ATTACHMENTS = @"attachments";
 NSString *const ATTACHMENT_IDENTIFIER = @"identifier";
 NSString *const ATTACHMENT_FILE_PATH = @"filePath";
+NSString *const ATTACHMENT_HIDE_THUMBNAIL = @"hideThumbnail";
+NSString *const ATTACHMENT_THUMBNAIL_CLIPPING_RECT = @"thumbnailClippingRect";
 NSString *const INTERRUPTION_LEVEL = @"interruptionLevel";
 NSString *const THREAD_IDENTIFIER = @"threadIdentifier";
 NSString *const PRESENT_ALERT = @"presentAlert";
@@ -912,6 +914,32 @@ static FlutterError *getFlutterError(NSError *error) {
             [NSMutableArray arrayWithCapacity:attachments.count];
         for (NSDictionary *attachment in attachments) {
           NSError *error;
+
+          NSMutableDictionary *options = [[NSMutableDictionary alloc] init];
+          if ([self containsKey:ATTACHMENT_HIDE_THUMBNAIL
+                  forDictionary:attachment]) {
+            NSNumber *hideThumbnail = attachment[ATTACHMENT_HIDE_THUMBNAIL];
+            [options
+                setObject:hideThumbnail
+                   forKey:UNNotificationAttachmentOptionsThumbnailHiddenKey];
+          }
+          if ([self containsKey:ATTACHMENT_THUMBNAIL_CLIPPING_RECT
+                  forDictionary:attachment]) {
+            NSDictionary *thumbnailClippingRect =
+                attachment[ATTACHMENT_THUMBNAIL_CLIPPING_RECT];
+            CGRect rect =
+                CGRectMake([thumbnailClippingRect[@"x"] doubleValue],
+                           [thumbnailClippingRect[@"y"] doubleValue],
+                           [thumbnailClippingRect[@"width"] doubleValue],
+                           [thumbnailClippingRect[@"height"] doubleValue]);
+            NSDictionary *rectDict =
+                CFBridgingRelease(CGRectCreateDictionaryRepresentation(rect));
+            [options
+                setObject:rectDict
+                   forKey:
+                       UNNotificationAttachmentOptionsThumbnailClippingRectKey];
+          }
+
           UNNotificationAttachment *notificationAttachment =
               [UNNotificationAttachment
                   attachmentWithIdentifier:attachment[ATTACHMENT_IDENTIFIER]
@@ -919,7 +947,7 @@ static FlutterError *getFlutterError(NSError *error) {
                                                fileURLWithPath:
                                                    attachment
                                                        [ATTACHMENT_FILE_PATH]]
-                                   options:nil
+                                   options:options
                                      error:&error];
           if (error) {
             result(getFlutterError(error));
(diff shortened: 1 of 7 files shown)


MaikuB:
Can you merge the latest changes from master to your branch/fork? This would fix the Kotlin plugin issue and allow seeing if tests are failing or not

flutter_local_notifications/lib/src/platform_specifics/darwin/notification_attachment.dart
```diff
@@ -18,4 +19,25 @@ class DarwinNotificationAttachment {
...
   final String? identifier;
+
+  /// Should the attachment be considered for the notification thumbnail?
+  final bool? hideThumbnail;
+
+  /// The clipping rectangle for the thumbnail image.
+  final DarwinNotificationAttachmentThumbnailClippingRect? thumbnailClippingRect;
+}
+
+class DarwinNotificationAttachmentThumbnailClippingRect {
```
MaikuB, line 30: - Can you add public API docs for these? This should've picked up by the linter
- Are you able to add an example that makes use of this?

MaikuB (changes requested):

flutter_local_notifications/lib/src/platform_specifics/darwin/notification_attachment.dart
noinskit: Both done!

MaikuB:
Good call out as I believe the API being used is also supported on macOS and my stance is that macOS support should be added for situations like this. On that note, are you able to make the macOS changes too or would you need support for that?

noinskit:
> Good call out as I believe the API being used is also supported on macOS and my stance is that macOS support should be added for situations like this. On that note, are you able to make the macOS changes too or would you need support for that?

I wanted to just implement, but I need help after all. I started the "example" app on my Mac (Ventura 13.0) and all the buttons silently do nothing, including `Show plain notification with payload` and `Request permission`. I know that e.g. Chrome sends me system notifications on the same system. Nothing in IntelliJ or XCode logs.

noinskit:
I'd add that I've found https://github.com/MaikuB/flutter_local_notifications/issues/1610 and tried a bunch of things to see a system dialog to ask me for permissions, incl. deleting `Derived Data`, `killall NotificationCenter` and some other things, with no luck. I cannot find it anywhere in `System Settings` either.

MaikuB:
Figured it out and will submit a PR to fix this soon. It looks like requesting critical alert permissions has prevented the permissions dialog appearing on macOS but was fine for me on iOS. Not sure why the behaviour is inconsistent

MaikuB:
If you update your branch/fork from the master branch then this should be in the change since https://github.com/MaikuB/flutter_local_notifications/pull/1809 is merged in

noinskit:
> If you update your branch/fork from the master branch then this should be in the change since #1809 is merged in

Thanks, I got the permission dialog now and I can trigger notifications! I'm looking at the thumbnail options on macOS now...

noinskit:
I pushed a version with macOS support. BUT it's not really working for me on my laptop, even though as far as I can tell it's being passed correctly (I looked at debug prints of the constructed `options` hashtable). I'm thinking that maybe these options aren't really supported on macOS yet. If that's the case, passing them along might still be correct to do in the plugin and some future macOS version might learn to respect them. That said, I'd appreciate if you took a look if something else could be wrong with the code!

MaikuB:
No worries, I'll take a look and see if I spot anything

MaikuB:
Whilst I take a look at this, could you add public API docs? The linter rules should highlight the ones that are missing

MaikuB:
Took a look and ran into same problem. I don't think there's anything wrong with what was implemented. The only thing that came to mind that didn't change the result is that the official docs for [UNNotificationAttachmentOptionsThumbnailClippingRectKey](https://developer.apple.com/documentation/usernotifications/unnotificationattachmentoptionsthumbnailclippingrectkey) say to use [this](https://developer.apple.com/documentation/coregraphics/1455760-cgrectcreatedictionaryrepresenta) to create the dictionary representation. Although it seems to have not made a difference, perhaps the code should change to use that function?

noinskit:
Sure, I made both changes! Still no impact on how the notification is shown on macOS, though :(

noinskit:
Or maybe not, looks like tests are now failing on:
`error: 'CGRectCreateDictionaryRepresentation' has been replaced by property 'CGRect.dictionaryRepresentation'`

I'm almost sure it worked for me locally; I'm guessing it's about compiler settings.

MaikuB:
Yeah it worked locally for me too. Perhaps the docs are out of date given the message. Looks like that particular change would need to be reverted as it sounds like what you did is more correct.

On a different note, in case you were squashing commits out of concern that it would affect the commit history when merged then don't worry about that GitHub provides the ability to do the maintainer's end before merging and it's what I select as the option too

noinskit:
Sure, it's back to `dictionaryRepresentation`.

flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
```diff
@@ -912,14 +914,40 @@ - (void)cancelAll:(FlutterResult _Nonnull)result {
...
+          if ([self containsKey:ATTACHMENT_THUMBNAIL_CLIPPING_RECT
+                  forDictionary:attachment]) {
+            NSDictionary *thumbnailClippingRect =
+                attachment[ATTACHMENT_THUMBNAIL_CLIPPING_RECT];
+            CGRect rect =
+                CGRectMake([thumbnailClippingRect[@"x"] doubleValue],
+                           [thumbnailClippingRect[@"y"] doubleValue],
+                           [thumbnailClippingRect[@"width"] doubleValue],
+                           [thumbnailClippingRect[@"height"] doubleValue]);
+            NSDictionary *rectDict =
```
MaikuB, line 935: for consistency and to avoid any potential issues down the line given what we saw on macOS, can this go back to using dictionaryRepresentation like it did before?

noinskit: I'm not sure if I understand... This part was always like that. The `dictionaryRepresentation` method on `CGRect` is there in Swift, but in ObjC this is it.

MaikuB: Apologies misremembered and turns out it's not even available in Objective-C

MaikuB (approved):

Issue #1801, opened 2022-11-22
README: Recommend android.window as Android dependencies


12people:
There is a desugaring-related bug on Android 12L and Android 13, where an app with desugaring enabled crashes on startup — see [this StackOverflow thread](https://stackoverflow.com/questions/74163927/flutter-chrash-in-android-12l-with-no-interface-method-addwindowlayoutinfoliste).

The fix for this desugaring bug is to add these two dependencies:
```
implementation 'androidx.window:window:1.0.0'
implementation 'androidx.window:window-java:1.0.0'
```

It'd be great if `flutter_local_notifications` recommended these two dependencies in the [guidelines on pubdev](https://pub.dev/packages/flutter_local_notifications#gradle-setup) alongside other necessary desugaring-related Gradle changes.

MaikuB:
Thanks for raising this. As this is open-sourced, would you be up for submitting a PR to add this information to the readme?

12people:
@MaikuB Submitted here: https://github.com/MaikuB/flutter_local_notifications/pull/1803

MaikuB:
Closing as the PR has been merged now :)

PR #1803, opened 2022-11-22
[flutter_local_notifications] Recommend WindowManager to fix Android 12L+ bugs

12people:
Fixes issue #1801

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -209,7 +209,19 @@ buildscript {
     }
 ```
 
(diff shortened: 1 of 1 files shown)


MaikuB:
Thanks for this. Personally I'd rather the readme removed mentioning using the launcher icon but I reckon there'd be those in the community that would keep doing so that keeping it with what you added in the PR will minimise the number of issues opened on this

MaikuB:
Thanks for this. Deleted my previous message in case you get an email/notification on it as I realised I responded about a different PR

PR #1894, opened 2023-02-13
[flutter_local_notifications_linux] Bump xdg_directories dependency

xvrh:
Also, I noticed that dependabot seems to not have run in while? Tried to add `enable-beta-ecosystems` to see if it improves the situation.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,6 @@
 version: 2
+enable-beta-ecosystems: true
+
 updates:
   - package-ecosystem: "pub"
     directory: "/flutter_local_notifications_platform_interface"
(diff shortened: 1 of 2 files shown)


MaikuB:
Thanks for the PR but I merged the one from Dependabot so will close this. It might be that you thought it hadn't run as it's configured to run daily. I don't know exactly what time it's scheduled to run though.

FYI the change from Dependabot is more safer to apply IMO as your bumps the minimum version of the library and can cause a breaking change as xdg_directories and/or issues resolving dependences as they bumped the minimum Flutter SDK requirement to 3. This plugin on the other hand has a minimum Flutter SDK requirement of 2.8

MaikuB:
Apologies, misread and forgot this plugin actually has minimum require of 3.0

FYI I'll be publishing a new release of the Linux plugin shortly

PR #1924, opened 2023-03-14
[flutter_local_notifications] Fix showMessagingNotification in example

michaelnew:
The notification ID was being incremented after creating the notification, which meant that trying to append a message 10 seconds later created an entirely new notification rather than just being added to the conversation.


MaikuB:
Good pickup and thanks for the PR!

PR #1940, opened 2023-03-25
[flutter_local_notifications] textstyle of theme changed in example

s4rath:
'subtitle1' is deprecated and shouldn't be used. 'titleMedium' is used instead.


MaikuB:
Thanks!

PR #1957, opened 2023-04-12
[flutter_local_notifications] Adds new Android foreground service types

Rexios80:
https://developer.android.com/about/versions/14/changes/fgs-types-required

This can be released now since you don't actually reference the APIs in native code, so we don't have to use preview APIs at all.


MaikuB:
Thanks!

Issue #1939, opened 2023-03-24
NSUserNotificationCenter Deprecation Warnings Despite Minimum macOS Target


cliftonlabrum:
Thanks for an awesome package! I have a minor bug to report.

The README for this package [states](https://pub.dev/packages/flutter_local_notifications#-supported-platforms):
> macOS 10.11+. On macOS versions older than 10.14, the plugin will use the [NSUserNotification APIs](https://developer.apple.com/documentation/foundation/nsusernotification). The [UserNotification APIs](https://developer.apple.com/documentation/usernotifications) (aka the User Notifications Framework) is used on macOS 10.14 or newer. Notification actions only work on macOS 10.14 or newer

I have a Flutter app for macOS with the minimum deployment set to `macOS 11.0`. Likewise my `Podfile` is also set for the same version:
```
//Podfile
platform :osx, '11.0'
```
Nevertheless, every macOS build of my app shows deprecation warnings:
```
1 warning generated.
#import "Headers/flutter_local_notifications-Swift.h"
        ^
.../build/macos/Build/Products/Debug/flutter_local_notifications/flutter_local_notifications.framework/Headers/flutter_local_notifications-Swift.h:270:33: warning: 'NSUserNotificationCenter' is deprecated: first deprecated in macOS 11.0 - All NSUserNotifications API should be replaced with UserNotifications.frameworks API
- (void)userNotificationCenter:(NSUserNotificationCenter * _Nonnull)center didActivateNotification:(NSUserNotification * _Nonnull)notification;
                                ^
@interface NSUserNotificationCenter : NSObject {
           ^
#import "Headers/flutter_local_notifications-Swift.h"
        ^
```
```
.../build/macos/Build/Products/Debug/flutter_local_notifications/flutter_local_notifications.framework/Headers/flutter_local_notifications-Swift.h:270:101: warning: 'NSUserNotification' is deprecated: first deprecated in macOS 11.0 - All NSUserNotifications API should be replaced with UserNotifications.frameworks API
- (void)userNotificationCenter:(NSUserNotificationCenter * _Nonnull)center didActivateNotification:(NSUserNotification * _Nonnull)notification;
                                                                                          ^
@interface NSUserNotification : NSObject <NSCopying> {
           ^
#import "Headers/flutter_local_notifications-Swift.h"
        ^
```
```
[...]

Issue #1295, opened 2021-08-25
IOS notification alert settings


mdikcinar:
I just wanted to ask how can i get alert notfication only on notfication screen without popup alert.

For example i want to close popup alert in first picture without close notification alert in 2nd picture.

<img src="https://user-images.githubusercontent.com/77580742/130706169-510ab5c9-c7a3-4f2d-a155-619bddc684a4.png" width="400" height="800">
<img src="https://user-images.githubusercontent.com/77580742/130706322-df3138b9-08b5-4446-ace1-3dd10d9d9b6f.png" width="400" height="800">

MaikuB:
This is currently not possible and requires an update from the plugin as Apple use to combine both together as one setting that is now deprecated (see https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions?language=objc). A pull request would be welcome for this

Ajay-jyothish:
Is there any update regarding this issue? or any other way this is possible?

okbye123:
This will help https://www.youtube.com/watch?v=aQTvI8L-mZs

PR #1965, opened 2023-04-17
[flutter_local_notifications] Fix macOS deprecation warnings

### LINKED ISSUE #1858: [flutter_local_notifications] suppress deprecation warnings on MACOS relating to old notification APIs #1740
Forgive me if this is the incorrect way to suggest this - I'm incredibly grateful for the work you did https://github.com/MaikuB/flutter_local_notifications/pull/1740 and was hoping the same could be done to suppress warning when building for MACOS as well as we use both in production.  Thanks for your time.

stevekohls:
This PR fixes #1858, also duplicated in #1939.

`NSUserNotification` was deprecated in macOS 10.14. This PR:
- Annotates the methods using that API with `@available(OSX, deprecated: 10.14)`
- Adds some `MARK` comments for improved navigability.

Warnings when building with 10.14 or higher should no longer occur.

### DIFF
--- a/flutter_local_notifications/macos/Classes/FlutterLocalNotificationsPlugin.swift
+++ b/flutter_local_notifications/macos/Classes/FlutterLocalNotificationsPlugin.swift
@@ -91,7 +91,7 @@ public class FlutterLocalNotificationsPlugin: NSObject, FlutterPlugin, UNUserNot
     public static func register(with registrar: FlutterPluginRegistrar) {
         let channel = FlutterMethodChannel(name: "dexterous.com/flutter/local_notifications", binaryMessenger: registrar.messenger)
         let instance = FlutterLocalNotificationsPlugin.init(fromChannel: channel)
-        if #available(OSX 10.14, *) {
+        if #available(macOS 10.14, *) {
             let center = UNUserNotificationCenter.current()
             center.delegate = instance
         } else {
@@ -103,7 +103,7 @@ public class FlutterLocalNotificationsPlugin: NSObject, FlutterPlugin, UNUserNot
 
     // MARK: - UNUserNotificationCenterDelegate
 
-    @available(OSX 10.14, *)
+    @available(macOS 10.14, *)
     public func userNotificationCenter(_ center: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
         if !isAFlutterLocalNotification(userInfo: notification.request.content.userInfo) {
             return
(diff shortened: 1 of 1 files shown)


MaikuB:
Thanks for the PR. I modified the example app to target macOS 11 like #1939 and can still see the warnings. For example, one of the entries is 

```
 warning: 'NSUserNotification' is deprecated: first deprecated in macOS 11.0 - All NSUserNotifications API should be replaced with UserNotifications.frameworks API
- (BOOL)userNotificationCenter:(NSUserNotificationCenter * _Nonnull)center shouldPresentNotification:(NSUserNotification * _Nonnull)notification SWIFT_WARN_UNUSED_RESULT SWIFT_AVAILABILITY(macos,deprecated=10.14);
```

I believe this is because even though Apple provided the User Notifications framework in 10.14 and that the plugin uses it if macOS 10.14 is used, the `NSUserNotification` APIs were actually deprecated in 11.

MaikuB:
Thanks for fixing that. Looks like that did the trick. Whilst not exactly what was reported in the original issue, given your expertise/knowledge, would you by chance know how to solve this one?

```
FlutterLocalNotificationsPlugin.swift:112:29: warning: 'alert' was deprecated in macOS 11.0
            options.insert(.alert)
```

Is it only possible to solve this one by having the plugin implement the replacements provided by https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions/3564813-list and https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions/3564812-banner?

stevekohls:
> Thanks for fixing that. Looks like that did the trick. Whilst not exactly what was reported in the original issue, given your expertise/knowledge, would you by chance know how to solve this one?
> 
> ```
> FlutterLocalNotificationsPlugin.swift:112:29: warning: 'alert' was deprecated in macOS 11.0
>             options.insert(.alert)
> ```
> 
> Is it only possible to solve this one by having the plugin implement the replacements provided by https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions/3564813-list and https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions/3564812-banner?

Thanks for reviewing my PR!

I did notice that warning a few days ago, and had a conversation with someone at my work about it also. There is a possibility to replace it with one of the new notification types.

Let me think on the best solution and I'll get back to you. If you can create an issue for it, I can put it on my stack of things to do.

There is a fix to the iOS/Objective-C code that I was planning to work on next –
The fix in #1740 suppresses the deprecation warning, but it goes ahead and compiles the code anyway. This could cause a compile-time error if the old API is ever removed.
A better approach would be to do what is done in the macOS Swift code, where there are two separate functions, one for the new API and one for the old API, and the functions are conditionally compiled based on API availability.

stevekohls:
@MaikuB Is there anything else I need to do in this PR for it to be merged? Do I have to update any version info?
I'm a long-time iOS developer, but new to Flutter so I'm still learning the toolchain, packaging, and best practices.

MaikuB:
No there isn't. Wanted to ask first before I merge it in. There is an existing issue already around the notification presentation options https://github.com/MaikuB/flutter_local_notifications/issues/1295

PR #1997, opened 2023-05-16
[flutter_local_notifications] Feature: Android - Add useAlarmClock mode to ScheduleMode

iballan:
Android Only

Add the useAlarmClock ScheduleMode to allow the ability to use AlarmManagerCompat.setAlarmClock method.

This is used for Alarms that should be shown event in low power mode. Used in Android native clock app.

This requires SCHEDULE_EXACT_ALARM permission


flutter_local_notifications/lib/src/platform_specifics/android/schedule_mode.dart
```diff
@@ -3,6 +3,11 @@
 /// This leverages the use of alarms to schedule notifications as described
 /// at https://developer.android.com/training/scheduling/alarms
 enum AndroidScheduleMode {
+  /// Used to specify that the notification should be scheduled to be shown at
```
MaikuB, line 6: This could be from lack of knowledge with the API but given the description, what is actually different between using this and `exactAlarmWhileIdle` then? Looking at the docs https://developer.android.com/reference/androidx/core/app/AlarmManagerCompat#setAlarmClock(android.app.AlarmManager,long,android.app.PendingIntent,android.app.PendingIntent), it says it would be same as `setExact` when using the `AlarmManagerCompat` APIs with `RTC_WAKEUP`. `RTC_WAKEUP` is already used as a flag in the exact scenario.

MaikuB (changes requested):
I left some comments there, if you could help answer/address them then that would be great. Another thing is could you update the example app to so there's a button that would make use of this. Would help show the difference as well

flutter_local_notifications/lib/src/platform_specifics/android/schedule_mode.dart
iballan: I actually tried all different versions of setting exact alarm and nothing worked for me to deliver notification and ignore the low power mode.
For example if you are making an alarm app, it doesn't make sense to fire the alarm after half an hour or ignore it when power is low. So I agree with you that the difference is RTC_WAKEUP but for APIs older than 21. This is the code of this method:

```java
    @SuppressLint("MissingPermission")
    public static void setAlarmClock(@NonNull AlarmManager alarmManager, long triggerTime,
            @NonNull PendingIntent showIntent, @NonNull PendingIntent operation) {
        if (Build.VERSION.SDK_INT >= 21) {
            Api21Impl.setAlarmClock(alarmManager,
                    Api21Impl.createAlarmClockInfo(triggerTime, showIntent), operation);
        } else {
            AlarmManagerCompat.setExact(alarmManager, RTC_WAKEUP, triggerTime, operation);
        }
    }
```

This requires SCHEDULE_EXACT_ALARM too like setExact. But this will show alarm icon on the status bar when you use it like this in the screenshot.

mmEq9


If you try to set an alarm on your device now, you will see that this icon will show up. Disabling the alarm will let it disappear only if there is no other app sets alarms.

iballan: @MaikuB  I added an example under a button called user alarm clock to schedule...

MaikuB: Thanks for the explanation

MaikuB (approved):
Thanks for making the changes. I made minor wording tweak on the button for consistency to get this in sooner.

PR #1998, opened 2023-05-16
[flutter_local_notifications] Feature: Android - Allow users to override broadcast receiver

iballan:
Android only

Allow reflection class creating to allow user to override broadcast receiver that is fired on each alarm and perform task there.

I personally needed this to update widget when the alarm is fired.

I extend receiver from class ScheduledNotificationReceiver 

Usage Example:

```kotlin
@Keep
class NotificationReceiver : ScheduledNotificationReceiver() {
    override fun onReceive(context: Context, intent: Intent) {
        super.onReceive(context, intent) // Super must be called, and marked as @CallSuper in ScheduledNotificationReceiver
        // Update widget on alarm fired
        try {
            val ids = AppWidgetManager.getInstance(context).getAppWidgetIds(
                ComponentName(
                    context,
                    MyHomeWidgetProvider::class.java
                )
            )
            val myWidget = MyHomeWidgetProvider()
            myWidget.onUpdate(context, AppWidgetManager.getInstance(context), ids)
        } catch (e: Exception) {
            e.printStackTrace()
        }
    }
}
```

While initializing on flutter:

```kotlin
var androidSettings = AndroidInitializationSettings('app_icon', 'com.mytest.package.receivers.NotificationReceiver' );
var iosSettings = DarwinInitializationSettings( ...
var initializationSettings = InitializationSettings(android: androidSettings, iOS: iosSettings);
```

Now on each alarm the widget will be updated, there might be other usages for this feature.


flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -1573,6 +1575,7 @@ private void initialize(MethodCall call, Result result) {
         applicationContext.getSharedPreferences(SHARED_PREFERENCES_KEY, Context.MODE_PRIVATE);
     SharedPreferences.Editor editor = sharedPreferences.edit();
     editor.putString(DEFAULT_ICON, defaultIcon).apply();
+    editor.putString(SCHEDULE_RECEIVER_REFLECTION_CLASS_NAME, scheduleReceiverReflectionClassName).apply();
```
MaikuB, line 1578: I've been seen a number of issues reported where people are running into exceptions around the default icon that is kept in shared preferences. Whilst no one has provided a way to reproduce this and how I've seen similar issues with other plugins, I strongly suspect that there are scenarios where shared preferences data has been cleared out and been considering to mandate specifying an icon for each notification. With that in mind, would like to avoid have this specified as part of the `initialize` and specified per Android notification. Would you be able to make that change make this parameter optional/nullable

MaikuB (changes requested):
Thanks for the PR. Left a comment and similar to what I wrote on the other PR, would you be able to update the example app to add a scenario to show how this could be used?

Another thing is since reflection is used where there's no direct reference to the specified receiver, won't there need to be extra instructions for app developers to update their R8/Proguard rules to ensure the receiver is kept even with a release build?

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
iballan: I can definitely make it optional/nullable. Regarding specifying the class name for each notification is not gonna be possible because the same class is used when you cancel the notification on Android. 

```java
  private void cancelNotification(Integer id, String tag) {
    Intent intent = new Intent(applicationContext, getReceiverClass(applicationContext)); // <- here we need the class as well
    PendingIntent pendingIntent = getBroadcastPendingIntent(applicationContext, id, intent);
    AlarmManager alarmManager = getAlarmManager(applicationContext);
    alarmManager.cancel(pendingIntent);
    NotificationManagerCompat notificationManager = getNotificationManager(applicationContext);
    if (tag == null) {
      notificationManager.cancel(id);
    } else {
      notificationManager.cancel(tag, id);
    }
    removeNotificationFromCache(applicationContext, id);
  }
```

iballan: Beside user should register the new receiver in manifest of course:
```java
<receiver
            android:name="com.mytest.package.receivers.NotificationReceiver"
            android:exported="false" />
```

iballan: @MaikuB do you think in order to have this feature we must make it adjustable for each notification? Or it may work with the manifest approach ?

MaikuB: What would your suggestion be? I'm not an Android developer but it would seem to me that the manifest approach is best in terms of the developer experience and have this used before. 

I've also been thinking more on this and correct me if I'm wrong but it would seem like your use case would've been possible if the plugin exposed a callback for when a notification is shown? Is this correct? I'm also noticing this use case is specific to Android so given there's no change on the iOS side, would sound like your app is only targeting Android. Bringing this up as if I understood this correctly, it would be better for the plugin to have a callback than to have this merged in that could end up being superseded

MaikuB: Thanks for the info. In light of what you said, I would say it's better for the community to have a cross-platform approach where they can deal with when a notification is being shown and then respond appropriately. With that in mind, will close the PR

PR #2001, opened 2023-05-18
[flutter_local_notifications] register android broadcast in init meth…

### LINKED ISSUE #1996: 安卓上架审核被拒说是sdk频繁自启动
com.dexterous.flutterlocalnotifications提示频繁自启动引起app上架审核失败

lwj1994:
register android broadcast in init function,  to fix #1996  and it is no need init in xml because export is false.

### DIFF
--- a/flutter_local_notifications/android/src/main/AndroidManifest.xml
+++ b/flutter_local_notifications/android/src/main/AndroidManifest.xml
@@ -6,15 +6,5 @@
     <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
     <application>
-        <receiver android:exported="false" android:name="com.dexterous.flutterlocalnotifications.ActionBroadcastReceiver" />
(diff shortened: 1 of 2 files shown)


MaikuB:
I tried this PR with the example app and it didn't work. Nothing got shown. Furthermore, upon looking at other existing libraries including those outside of the Flutter ecosystem, I can see that the manifest approach is also used. Makes sense as it improves the developer experience overall. With that in mind, I would say for that can't use it due to restrictions imposed by their local market like China should look at a fork

lwj1994:
> I tried this PR with the example app and it didn't work. Nothing got shown. Furthermore, upon looking at other existing libraries including those outside of the Flutter ecosystem, I can see that the manifest approach is also used. Makes sense as it improves the developer experience overall. With that in mind, I would say for that can't use it due to restrictions imposed by their local market like China should look at a fork

thanks.

 example it works. It could be something else。
image   

manifest is ok，It's just that it's too early to initialize and and uncontrolled.

MaikuB:
You chose to show a notification that is meant to appear immediately. Your PR relates to scheduled notifications

PR #2002, opened 2023-05-18
[flutter_local_notifications] Drop Flutter unsupported platforms from README.md

vbuberen:
Updated the README file to drop the information about older iOS versions as Flutter doesn't support iOS < 11. 
Additionally updated mentions of other platforms to mirror the official supported platforms list: https://docs.flutter.dev/reference/supported-platforms

The only part in README about iOS 10 and older is about the `uiLocalNotificationDateInterpretation` parameter as it is still a required parameter while package is used, so I believe that it should be removed after going through a deprecation cycle.

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -36,7 +36,7 @@ A cross platform plugin for displaying local notifications.
    - [Example app](#example-app)
    - [API reference](#api-reference)
 - **[Initialisation](#initialisation)**
-   - [[iOS (all supported versions) and macOS 10.14+] Requesting notification permissions](#ios-all-supported-versions-and-macos-1014-requesting-notification-permissions)
+   - [[iOS and macOS] Requesting notification permissions](#ios-all-supported-versions-and-macos-1014-requesting-notification-permissions)
    - [Displaying a notification](#displaying-a-notification)
    - [Scheduling a notification](#scheduling-a-notification)
    - [Periodically show a notification with a specified interval](#periodically-show-a-notification-with-a-specified-interval)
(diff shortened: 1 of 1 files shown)


MaikuB:
From what I could tell, you're right that as of the current Flutter stable release that it's iOS 11+. However, this plugin has minimum supported Flutter version of 3.0.0 that was released on the 12th of May 2022. The state of supported platforms at that time can be found at https://github.com/flutter/website/blob/709463f235c53566af5ec4b6d93e6d0426d9b35e/src/development/tools/sdk/release-notes/supported-platforms.md is that iOS 9 was still supported. With that in mind, this should that yes there is some incorrect info in the readme but I believe the fix should be based on what was listed on that page to align with what Flutter 3.0 supports

vbuberen:
Ok, makes sense, however, most of Google supported plugins (I mean these ones: https://github.com/flutter/packages/tree/main/packages) already moved to min Flutter 3.3 version, so it is highly likely that most users already moved to more recent Flutter versions. I also bumped Plus Plugins to min 3.3 recently to encourage more users to move in case there were some with older versions.
And there, in Flutter 3.3 iOS 11 already was the minimally supported version: https://github.com/flutter/website/pull/7481/files#diff-15bc8d52ef88ef94c935f92dc29910b7b712c7ce2aaa693d2be71806258eeaa9R32

MaikuB:
Thanks for the info. Aware of what the team's doing there and across the minimum version as of Flutter 3.3. This seems to be more driven by what makes it easier to maintain. This is different from how this plugin actually calls APIs available on older iOS versions. Granted it's a smaller market share but whilst maintaining this plugin, I had seen issues raised by devs using older versions of the plugin that can be indicative of those targeting older devices. There will be a point where the bump will happen as it would help maintenance. However, right now I don't see there being an urgent need to do e.g. based on new functionality available only available in newer Flutter/Dart SDK that the plugin absolutely needs

Issue #1102, opened 2021-03-25
[Request] Add option for provisional authorization (iOS/macOS)


seanmc86:
As per the [Apple documentation](https://developer.apple.com/documentation/usernotifications/asking_permission_to_use_notifications), rather than explicitly requesting authorization, we can use provisional authorization to send messages that the user can keep or turn off.

This just requires adding the `.provisional` option to` center.requestAuthorization`

Could this be added?

Does the `initialize ` method need to be adjusted in cases where we don't always want to use the default authorization request, or can it simply not be called?

MaikuB:
Could you submit a PR for it? Regarding `initialize`, this could default to not requesting it. Note there's another API that is specific to iOS/macOS for requesting permissions

amitm29:
Can you mention the name of the API to request provisional permission. 

Actually, I am using this plugin alongside firebase_messaging plugin. I am using following code in my main method to request provisional permission.

await FirebaseMessaging.instance.requestPermission(
      provisional: true,
    );

The firebase messaging plugin is working fine with provisional permission. But, i am not able to show notifications using this permission, granting notification permission.

Can you help me in understanding and resolving this issue. Thanks!

PR #2022, opened 2023-06-09
[flutter_local_notifications] Add provisional notification support for iOS 12+ and macOS 10.14+

Tokenyet:
### Related issue
[Add option for provisional authorization (iOS/macOS)](https://github.com/MaikuB/flutter_local_notifications/issues/1102)

### Description
This pull request implements the addition of provisional push notifications as a new feature. Provisional push notifications allow sending notifications without displaying pop-up permission requests to the user. You can refer to [this documentation](https://developer.apple.com/documentation/usernotifications/asking_permission_to_use_notifications) or take a quick look at [OneSignal's documentation on iOS provisional push notifications](https://documentation.onesignal.com/docs/ios-provisional-push-notifications) for more information.

#### User Story
As a product owner, I want the ability to send provisional notifications instead of regular ones. This eliminates the need to interrupt the user with permission prompts in a separate window.

### Notice
There are some limitations to consider with this option, which are outlined in [this quote](https://documentation.onesignal.com/docs/ios-provisional-push-notifications):

> As these are provisional notifications, they come with reduced functionality, including:
> 
> - No banner display
> - No associated sound
> - No user alert
> - Sent directly to the Notification Center


MaikuB:
Thanks for the PR. Would you by chance be able to help implement this on the macOS side as well?

Tokenyet:
@MaikuB 
Typically, I focus on developing projects related to iOS. If support for macOS is required, it might take a while. If it's not urgent, we can set this PR aside for now and I will add the necessary code and conduct testing when I have some free time. Alternatively, we can prioritize iOS support and delegate the implementation of the macOS portion to someone else.

In summary, if there is no urgency, please understand that I will need some time to address the macOS aspect.

MaikuB:
Understood, though as this is a cross-platform plugin and APIs and the Dart code (`DarwinInitializationSettings`) applies to both macOS and iOS, then the feature should apply to both. I don't have any concerns on waiting

Tokenyet:
Hi @MaikuB, I tested with macOS version 13.3（22E252） and discovered that macOS does support the UNAuthorizationOptionProvisional flag. However, during my testing, I did not receive any provisional notifications on macOS. Even after adding the flag, the system still displayed a dialog asking for permission. If we reject the permission, we won't be able to send any notifications. You can observe this behavior in my last commit. I believe it's unnecessary to add support for macOS since it has no effect. Therefore, I plan to drop that commit and push the changes again to remove it.

Let me know your thoughts. Should we still include the code even if it has no effect on macOS?

MaikuB:
Can you still add it? It's not the first time I've seen such behaviour and perhaps there was something missing that those of us less experienced with macOS don't know about

Tokenyet:
> Can you still add it? It's not the first time I've seen such behaviour and perhaps there was something missing that those of us less experienced with macOS don't know about

Thank you for your feedback, I have made the necessary change to support macOS versions in my pull request. However, I wanted to bring to your attention that when testing the code with Xcode, I received a warning advising to use this specific code snippet for iOS 12 and above. You can find the relevant line in https://github.com/MaikuB/flutter_local_notifications/pull/2022/files#diff-7cbfbb9097b07012487c85b50c6cb6973d58228b0286631a9a5c8349af7690faR514

image

I wanted to mention this to ensure that we address any potential compatibility issues with macOS and iOS versions correctly. Please let me know if there are any further modifications or considerations needed to ensure seamless compatibility across different macOS and iOS versions. I appreciate your guidance in this matter

MaikuB:
Ok I believe I know what happened now, you weren't seeing the changes working as  you were updating the code for the iOS implementation of the plugin. The macOS implementation where changes should be made is in [this file](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/macos/Classes/FlutterLocalNotificationsPlugin.swift)

Tokenyet:
> Ok I believe I know what happened now, you weren't seeing the changes working as you were updating the code for the iOS implementation of the plugin. The macOS implementation where changes should be made is in [this file](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/macos/Classes/FlutterLocalNotificationsPlugin.swift?rgh-link-date=2023-06-30T08%3A50%3A20Z)

Thanks for pointing me out! I thought iOS and macOS are the same code base. Let me try again.

Tokenyet:
Here is the updated code for macOS. Addressed at [894bedc](https://github.com/MaikuB/flutter_local_notifications/pull/2022/commits/894bedcfc4080fa6349a297c8800810216d4ca05) and [6e5d5eb](https://github.com/MaikuB/flutter_local_notifications/pull/2022/commits/6e5d5eb26186520c6e91af48d60cd37594005bd4).

Demo:

https://github.com/MaikuB/flutter_local_notifications/assets/6420527/fffff7be-91c0-41ac-b42b-e6aa441524c0

flutter_local_notifications/example/lib/main.dart
```diff
@@ -310,6 +310,29 @@ class _HomePageState extends State<HomePage> {
     }
   }
 
+  Future<void> _requestProvisionalPermissions() async {
```
MaikuB, line 313: How does this work within the context of the example app? From what I could tell, it doesn't align with current experience as the example app requests the permissions on startup. Once you give or don't give permissions at the prompt, the prompt no longer appears and requires going to app settings to change and calling this doesn't actually have an effect. Moreover, most users will grant permissions on startup. With that in mind, unless I have missed anything (do let me know if I have) then this looks like this would be something better left out of the example app. Don't want the example app to default to request provisional permissions on startup instead either as it's likely rare for apps to use this and could leave to more questions being raised around features not working properly

flutter_local_notifications/lib/src/platform_specifics/darwin/initialization_settings.dart
```diff
@@ -34,6 +35,11 @@ class DarwinInitializationSettings {
   /// Default value is true.
   final bool requestBadgePermission;
 
+  /// Request permission to send provisional notification for iOS 12+
```
MaikuB, line 38: Could you add information on the version of macOS that the setting applies to? Another thing is could you follow the existing convention in the API docs has on how to capture the supported OS versions? See `defaultPresentSound` for an example at https://github.com/MaikuB/flutter_local_notifications/blob/558b2ddbfa0c6360e5fbca33300c69e8938e1f7e/flutter_local_notifications/lib/src/platform_specifics/darwin/initialization_settings.dart#L68. What I'm referring to on following so there's consistency is the wording and how this is another paragraph after mentioning the default value

MaikuB (changes requested):
Thanks for this. I left a couple of comments on the PR

flutter_local_notifications/lib/src/platform_specifics/darwin/initialization_settings.dart
Tokenyet: Addressed at [82e55c4](https://github.com/MaikuB/flutter_local_notifications/pull/2022/commits/82e55c48d6b4093e014b01afbb3d427ee2f88878)

flutter_local_notifications/example/lib/main.dart
Tokenyet: You are correct. I will revert the changes to leave the file untouched.

When testing the provisional feature, we should follow these steps:

1. Comment out `requestPermission`.
2. Clear the permission in the system's settings -> notifications if it was previously installed.
3. Call `requestProvisionalPermission`.
If the user had previously rejected the permission before we called `requestProvisionalPermission`, then `requestProvisionalPermission` will not work until step 2 is completed.


Addressed at [77a2d07](https://github.com/MaikuB/flutter_local_notifications/pull/2022/commits/77a2d074bf80e3ac6b864f094f28eada82aa5e10)

MaikuB:
Thanks. Will take a look at this soon. Had to prioritise a hot fix so didn't look at this first

MaikuB (approved):

Issue #2055, opened 2023-07-17
Unable to run code onDidReceiveBackgroundNotification when app is terminated


asg1997:
Hi there. 
Thank you for your plugin!

I'm facing the following issue:

I need to make a transition on particular screen when my app is terminated. Everything works fine when the app isn't closed. This is what I've tried.

I scoured the plugins' docs but didn't find the answer. I'm I missing something?
Should use channels to run the actions when app is terminated?
```dart
    import 'dart:async';
    
    import 'package:firebase_messaging/firebase_messaging.dart';
    import 'package:flutter_local_notifications/flutter_local_notifications.dart';
    import 'package:flutter_native_timezone/flutter_native_timezone.dart';
    import 'package:like_cucumber/core/like_cucumber.dart';
    import 'package:like_cucumber/data/mappers/task_notification_payload_mapper.dart';
    import 'package:like_cucumber/presentation/tab_bar/tab_bar/tab_bar.dart';
    import 'package:like_cucumber/presentation/tab_bar/tasks_tab/task_info.dart';
    import 'package:timezone/data/latest_all.dart' as tz;
    import 'package:timezone/timezone.dart' as tz;
    
    @pragma('vm:entry-point')
    void onDidReceiveBackgroundNotification(NotificationResponse response) async {
      final payload = response.payload;
      if (payload == null) return;
      final task = TaskNotificationPayloadMapper.fromPayload(payload);
      navigatorKey.currentState?.push(TabBarPage.getRoute(initialTabIndex: 1));
      navigatorKey.currentState?.push(TaskInfoPage.getRoute(
        task: task,
      ));
    }
    
    class LocalNotificationsService {
      factory LocalNotificationsService() => _instance;
      LocalNotificationsService._();
      static final _instance = LocalNotificationsService._();
    
      final _plugin = FlutterLocalNotificationsPlugin();
      final _messaging = FirebaseMessaging.instance;
    
      Future<void> init() async {
        await _requestPermission();
        await _initCurrentTimezone();
        await _initPlugin();
      }
    
      Future<void> _requestPermission() => _messaging.requestPermission();
    
      Future<void> _initCurrentTimezone() async {
        tz.initializeTimeZones();
        final localName = await FlutterNativeTimezone.getLocalTimezone();
        tz.setLocalLocation(tz.getLocation(localName));
      }
    
      Future<void> _initPlugin() async {
        const DarwinInitializationSettings iosSettings =
[...]

PR #2057, opened 2023-07-22
Improve app launch instructions

JamesMcIntosh:
see conversation about improving docs: https://github.com/MaikuB/flutter_local_notifications/issues/2055

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -570,7 +570,7 @@ await flutterLocalNotificationsPlugin.initialize(initializationSettings,
     onDidReceiveNotificationResponse: onDidReceiveNotificationResponse);
 ```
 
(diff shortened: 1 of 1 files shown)


flutter_local_notifications/README.md
```diff
@@ -589,6 +589,20 @@ void onDidReceiveNotificationResponse(NotificationResponse notificationResponse)
 
 In the real world, this payload could represent the id of the item you want to display the details of. Once the initialisation is complete, then you can manage the displaying of notifications. Note that this callback is only intended to work when the app is running. For scenarios where your application needs to handle when a notification launched the app refer to [here](#getting-details-on-if-the-app-was-launched-via-a-notification-created-by-this-plugin)
 
+Use the `getNotificationAppLaunchDetails` method when the app starts if you need to handle when a notification triggering the launch for an app e.g. change the home route of the app for deep-linking.
```
MaikuB, line 592: Thanks for making the PR. Upon looking at what already exists, what got added here looks quite odd as there's now already a statement with a link to the section on dealing with when a notification causes an app to start (`For scenarios where your application needs to handle when a notification launched the app refer to [here](#getting-details-on-if-the-app-was-launched-via-a-notification-created-by-this-plugin)`). Besides how the readme actually mentions the dealing with this scenario requires doing something different, it reads awkwardly in that it is  immediately followed with a separate paragraph with instructions that is on the same topic. Perhaps it's better that what you added goes to the section that is linked?

MaikuB:
Closing due to lack of further response and docs already have a link to code snippet on calling the method needed

Issue #2033, opened 2023-06-25
Crashlytics error with useExactAlarm on version 14.1.1


mk-dev-1:
We just updated the flutter_local_notification dependency in our app to the latest version 14.1.1 and are slowly starting to see error messages come in as shown below. We have previously been using version 13.0.0, therefore suspect that this issue may have been introduct with version 14.0.0.

Note that our app does not declare `SCHEDULE_EXACT_ALARM` but uses `USE_EXACT_ALARM` since on-time notifications are absolutely paramount.

Does anyone have any suggestions?

```
Fatal Exception: java.lang.RuntimeException: Unable to start receiver com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.dexterous.flutterlocalnotifications.models.ScheduleMode.useExactAlarm()' on a null object reference
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:4458)
       at android.app.ActivityThread.access$1700(ActivityThread.java:261)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2165)
       at android.os.Handler.dispatchMessage(Handler.java:111)
       at android.os.Looper.loopOnce(Looper.java:238)
       at android.os.Looper.loop(Looper.java:357)
       at android.app.ActivityThread.main(ActivityThread.java:8090)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1026)

Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.dexterous.flutterlocalnotifications.models.ScheduleMode.useExactAlarm()' on a null object reference
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.setupAllowWhileIdleAlarm(:2)
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.scheduleNextRepeatingNotification(:48)
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.scheduleNextNotification(:20)
       at com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver.onReceive(:127)
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:4449)
[...]

PR #2076, opened 2023-08-22
[flutter_local_notifications] fixed an issue with null schedule mode for older notifications scheduled using periodicallyShow on Android

MaikuB:
Relates to #2033

### DIFF
--- a/flutter_local_notifications/CHANGELOG.md
+++ b/flutter_local_notifications/CHANGELOG.md
@@ -1,3 +1,7 @@
+# [16.0.0-dev.2]
+
+* [Android] fixed issue an issue similar to [2033](https://github.com/MaikuB/flutter_local_notifications/issues/2033) that was addressed in 15.0.1 where notifications on scheduled using older version of the plugin via the `periodicallyShow()` method would fail to have the next subsequent ones scheduled. This issue started occuring in 14.0 where support for inexact notifications was added using the `ScheduleMode` enum that was added and resulted in the deprecation of `androidAllowWhileIdle`. A mechanism was added to help "migrate" old notifications that had `androidAllowWhileIdle` specified but didn't account for how there are recurring notifications that were scheduled using older versions of the plugin prior to `androidAllowWhile` being added
+
 # [16.0.0-dev.1]
 
 * [Android] **Breaking change** renamed the `requestPermission()` associated with the `AndroidFlutterLocalNotificationsPlugin` class to `requestNotificationsPermission()`. This was done to be more explicit given another method (`requestExactAlarmsPermission()`) has been added that also requests a permission (more details below).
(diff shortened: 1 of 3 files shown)


nohli:
Thanks for the fix!

I don't know how to test it properly though?
It gladly only affects a few users.

MaikuB:
When you say you don't know how to fix, by that do you mean you don't know how to reproduce these issue or that you don't know how to update your app to point to the branch referenced by this PR?

nohli:
Don't know how to reproduce the issue, as it happens for <0.001% of users (with daily notifications).

MaikuB:
Based on the trace you shared before, it indicates to me that you have an app that has used version 1.x of this plugin and calls the `periodicallyShow()` method as the `androidAllowWhileIdle` parameter was added in 2.0. The issue happens as the changes done as part of https://github.com/MaikuB/flutter_local_notifications/pull/1881 didn't seem to account for this and is similar to the #2033 but for a different method provided by the plugin

nohli:
Oh, interesting. You mean, the users who experience this crash use the app for about 3 years, where it used plugin version <2? That would explain why the exception is so rare.

It's a bit hard to roll back the system to a state of 3 years ago, isn't it? I can try to downgrade Flutter etc., checkout an old commit, install the app, and then revert everything, install again and see if the crash happens.

But the app calls `cancelAll()` on a fresh app start and reschedules all notifications. Could it still cause this error?

On a side note, one user reported that he can't cancel a daily notification. His workaround is to disable notifications from the app's `channelId` in system settings. Could it be that it's him having the exceptions, and for some reason the plugin can't cancel very old notifications, which then cause this exception?
The app regularly calls `cancelAll()` and reschedules all notifications. Users can see their notifications (which get rescheduled after cancelling), and he reported the notification doesn't show up in the list.

MaikuB:
> It's a bit hard to roll back the system to a state of 3 years ago, isn't it? I can try to downgrade Flutter etc., checkout an old commit, install the app, and then revert everything, install again and see if the crash happens

Yeah it could be quite hard to roll back. May have to simulate by modifying the code so that the JSON representing the notification saved to shared preferences is same as what would get saved from version 1.x

> But the app calls cancelAll() on a fresh app start and reschedules all notifications. Could it still cause this error?

It should but perhaps there are users who haven't started another session?

> On a side note, one user reported that he can't cancel a daily notification. His workaround is to disable notifications from the app's channelId in system settings. Could it be that it's him having the exceptions, and for some reason the plugin can't cancel very old notifications, which then cause this exception?

I'm doubtful as I've seen a correlation between a channel and cancellation. Should be something that can be verified easily though

MaikuB:
@nohli FYI this has been released as part of 16.0.0-dev.2, 15.1.1 and 14.1.3

nohli:
> @nohli FYI this has been released as part of 16.0.0-dev.2, 15.1.1 and 14.1.3

Wow thank you so much!! ❤️

nohli:
15.1.1 is not yet available on pub.dev.

MaikuB:
Hmm thought I published it. Perhaps I forgot to hit the button. Hang on

MaikuB:
Ok done now. Got mixed up with when I pushed out 14.1.3+1

nohli:
Your support is really amazing! ❤️

Still can't see 15.1.1 nor 14.1.3 though 🙈

MaikuB:
I forgot to bump the version in pubspec.yaml file that I missed publishing failed. It's done but you have to go to versions tab to see 14.1.3

image

nohli:
Great, 15.1.1 is live, thank you again 😊

nohli:
FYI no crashes with the latest version 🙏🏽

PR #2096, opened 2023-09-14
[flutter_local_notifications] Update proguard-rules.pro

Fernandomr88:
Updating from https://github.com/google/gson/blob/main/examples/android-proguard-example/proguard.cfg


MaikuB:
Thanks for raising the PR but it doesn't add anything relevant to the example app. If you take a look at 

```
-keep class com.google.gson.examples.android.model.** { <fields>; }
```

This is only relevant to the example app for the gson library (hence the name of the namespace) so this was why it was left out

PR #2125, opened 2023-10-12
fixed documentation spell

corpiii:
Hello MaikuB. 👋 Thank you for providing this super useful library.
I am writing a PR because I have a suggestion while analyzing the code.

### initialise -> initialize
I know that initialise is correct. 
However, many parts of Flutter and Android use initialize, and I think keeping it consistent could look a little better.
So, I suggest changing initialise to initialize.

```dart
// in InitializationSettings.dart

/// Settings for initializing the plugin for each platform.
class InitializationSettings {
  /// Constructs an instance of [InitializationSettings].
  const InitializationSettings({
    this.android,
    this.iOS,
    this.macOS,
    this.linux,
  });

  /// Settings for Android.
  ///
  /// It is nullable, because we don't want to force users to specify settings
  /// for platforms that they don't target.
  final AndroidInitializationSettings? android;

  /// Settings for iOS.
  ///
  /// It is nullable, because we don't want to force users to specify settings
  /// for platforms that they don't target.
  final DarwinInitializationSettings? iOS;

  /// Settings for macOS.
  ///
  /// It is nullable, because we don't want to force users to specify settings
  /// for platforms that they don't target.
  final DarwinInitializationSettings? macOS;

  /// Settings for Linux.
  ///
  /// It is nullable, because we don't want to force users to specify settings
  /// for platforms that they don't target.
  final LinuxInitializationSettings? linux;
}

```

In the case above, initialize is also used.


MaikuB:
Thanks for the PR. Happy to concede on making the change in API docs to be consistent with the code where programming languages already follow American English but not looking to have these changes done for anything else. I can see in the PR that you made changes to the changelog and readme

Coedice:
I would also add that the commit message should be changed in my opinion, since non-US spellings are not typos

MaikuB:
Closing this due to lack of response/updates

PR #2145, opened 2023-11-03
[flutter_local_notifications] Check permissions for iOS and macOS

Vorkytaka:
With this PR we add ability to check notification permissions for both iOS and macOS.
We can check if notifications is just enabled or if there is some specific type is enabled (like alert, badge, etc).

---

For iOS we use `UNUserNotificationCenter` for version 10 or above and `UIApplication.sharedApplication.currentUserNotificationSettings` for 9 and less (with ignore of critical and provisional).

Both version was tested with iOS simulator.

---

For macOS we just use `UNUserNotificationCenter`.

---

In both OS we separately check for provisional, because there can be either authorized status or provisional.

### DIFF
--- a/flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
+++ b/flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
@@ -592,6 +592,12 @@ static FlutterError *getFlutterError(NSError *error) {
         [center getNotificationSettingsWithCompletionHandler:^(UNNotificationSettings * _Nonnull settings) {
             BOOL isEnabled = settings.authorizationStatus == UNAuthorizationStatusAuthorized;
             
+            if(@available(iOS 12.0, *)) {
+                if(provisionalPermission) {
+                    isEnabled = settings.authorizationStatus == UNAuthorizationStatusProvisional;
+                }
+            }
+            
             if(soundPermission) {
                 isEnabled = isEnabled && settings.soundSetting == UNNotificationSettingEnabled;
             }
@@ -602,15 +608,10 @@ static FlutterError *getFlutterError(NSError *error) {
                 isEnabled = isEnabled && settings.badgeSetting == UNNotificationSettingEnabled;
             }
             if(@available(iOS 12.0, *)) {
-                if(provisionalPermission) {
-                    isEnabled = isEnabled && settings.authorizationStatus == UNAuthorizationOptionProvisional;
-                }
-                
                 if(criticalPermission) {
                     isEnabled = isEnabled && settings.criticalAlertSetting == UNNotificationSettingEnabled;
                 }
             }
-            
             result(@(isEnabled));
         }];
     } else {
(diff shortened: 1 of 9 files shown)


MaikuB:
Thanks for the PR. Would you be able to change this so the API doesn’t require input arguments and returns an object with a boolean per permission that indicates if the associated permission is approved/enabled? The reason I brought this up is the because semantics of the API currently look a bit awkward around taking multiple boolean parameters and returning a single one that may cause confusion to some on what this may mean (note: I understand how it works). I believe taking the approach I mentioned would bring more clarity and the code using the API after would be more readable as an English sentence

Vorkytaka:
> Thanks for the PR. Would you be able to change this so the API doesn’t require input arguments and returns an object with a boolean per permission that indicates if the associated permission is approved/enabled? The reason I brought this up is the because semantics of the API currently look a bit awkward around taking multiple boolean parameters and returning a single one that may cause confusion to some on what this may mean (note: I understand how it works). I believe taking the approach I mentioned would bring more clarity and the code using the API after would be more readable as an English sentence

Good idea!
I'll try to get on it soon.

Vorkytaka:
> Thanks for the PR. Would you be able to change this so the API doesn’t require input arguments and returns an object with a boolean per permission that indicates if the associated permission is approved/enabled? The reason I brought this up is the because semantics of the API currently look a bit awkward around taking multiple boolean parameters and returning a single one that may cause confusion to some on what this may mean (note: I understand how it works). I believe taking the approach I mentioned would bring more clarity and the code using the API after would be more readable as an English sentence

Here's updated version.

In this version, I've left the consistent behavior, with splitting into multiple bool fields.
But maybe in the future it would be better to move to enum, which will be analogous to `UNAuthorizationStatus`, for `isEnabled` and `isProvisionalEnabled` fields.

numen31337:
Could we consider developing this feature for both iOS and Android? As a reference, [here's](https://github.com/Vanethos/flutter_notification_permissions/blob/master/android/src/main/java/com/vanethos/notification_permissions/MethodCallHandlerImpl.java) some Android implementation as an example.

Vorkytaka:
> Could we consider developing this feature for both iOS and Android? As a reference, [here's](https://github.com/Vanethos/flutter_notification_permissions/blob/master/android/src/main/java/com/vanethos/notification_permissions/MethodCallHandlerImpl.java) some Android implementation as an example.

This plugin already has support for a similar request on Android: [*click*](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart#L508)

flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
```diff
@@ -96,6 +97,13 @@ @implementation FlutterLocalNotificationsPlugin {
 NSString *const PRESENTATION_OPTIONS_USER_DEFAULTS =
     @"flutter_local_notifications_presentation_options";
 
+NSString *const CHECK_ENABLED = @"isEnabled";
```
MaikuB, line 100: thanks for making the change. minor thing but can the variables be renamed in the iOS and macOS implementation be renamed to be similar their values e.g. rename `CHECK_SOUND_ENABLED` to  `IS_SOUND_ENABLED`? the current naming reads like flags that the plugin should use check if the corresponding permission is enabled instead of a result

MaikuB (changes requested):

flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m
Vorkytaka: Done

Issue #1926, opened 2023-03-15
NotificationAppLaunchDetails not working as expected.


arv-aks:
void checkIfTheAppIsOpenedFromNotification() async {
    final NotificationAppLaunchDetails? notificationAppLaunchDetails =
        await NotificationService.flutterLocalNotificationsPlugin
            .getNotificationAppLaunchDetails();

    NotificationResponse? response;

    if (notificationAppLaunchDetails != null) {

      //check if the app is launched from the notification

      if (notificationAppLaunchDetails.didNotificationLaunchApp) {
        print("app launched from notification");

        //check if the notification contains the reponse or not
        if (notificationAppLaunchDetails.notificationResponse != null) {

          response = notificationAppLaunchDetails.notificationResponse;
          //check for the response or payload

          if (response != null && response.payload != null) {
            //navigate...

          } else {}
        }
      } else {
        print("app not lauched from notification....");
      }
    }
      }
    

-> When the app is killed and tap on notification -> The app is opened and **notificationAppLaunchDetails.didNotificationLaunchApp = true.** Thats working fine.
-> When the app is in memory (i.e. background) and I opened other apps. After sometime I click on this app icon to launch. Then the app restarted (from recents) and the value of notificationAppLaunchDetails.didNotificationLaunchApp = **true**. Which should be **false**. Because the app is not opened from notification

MaikuB:
What you've described is still working as expected as the purpose of the method is to find out if was launched/started from a terminated state by a notification. This did happen from what got described but the app wasn't terminated again in the scenario. It was put in the background and you resumed the app by using the recents screen so the fact that it reported that a notification was used to start the app still remains true

arv-aks:
@MaikuB 
But I'm navigating to another screen when the app opened from notification but currently it's also navigating to another screen every time app restarting from background. How can I solve this?

MaikuB:
I would suggest monitoring the app lifecycle events to help track that changes in the app or that you maintain state that you've dealt with the event. Was thinking recently to add a parameter that allows specifying if it should be consumed or not that would solve your issue. It would be something part of the current 14.0 prerelease and I'm happy to take a PR on this if you can help out on implementing this

arv-aks:
I'll provide a video asap. And yes the launchedActivityFromHistory is present. 
I tried to implement  a method for checking if the app is opened from recents or from notification, but still this method is giving me true with previous notification payload data.

MaikuB:
[...]

PR #2150, opened 2023-11-08
[flutter_local_notifications] solves the expected behavior for 'didNotificationLaunchApp'

mdAlvarenga:
Condition added to avoid 'didNotificationLaunchApp=true' when opening app from recent apps list in Android.
Related to #1926


MaikuB:
Thanks for the PR. Unfortunately this introduces an issue where on Android, subsequent calls to `getNotificationAppLaunchDetails()` returns information to indicate that the app wasn't launched from a notification and this would happen even if the app wasn't put in the background and resumed later on. This results in inconsistent behaviour on iOS and macOS. The approach used by you used seems to be to essentially consume the notification once read. If this is to be done, then I would say that the PR should change to make this a behaviour that the user can opt-into e.g. update `getNotificationAppLaunchDetails()` to take a `consumeNotification` boolean argument and implement the expected behaviour on all the supported platforms. Is this something you can do as part of the pR?

Issue #2160, opened 2023-11-20
[iOS] Conflict with OpenCV


MarkerwApk:
Hi,

I can't build app with plugin and OpenCv Library. I have conflicts:

`duplicate symbol '_OBJC_CLASS_$_Converters' in:
    /Users/user/Workspace/Flutter/libero/build/ios/Debug-gkDev-iphoneos/flutter_local_notifications/flutter_local_notifications.framework/flutter_local_notifications[4](Converters.o)
    /Users/user/Workspace/Flutter/libero/ios/opencv2.framework/Versions/A/opencv2[arm64][841](Converters.o)

duplicate symbol '_OBJC_METACLASS_$_Converters' in:
    /Users/user/Workspace/Flutter/libero/build/ios/Debug-gkDev-iphoneos/flutter_local_notifications/flutter_local_notifications.framework/flutter_local_notifications[4](Converters.o)
    /Users/user/Workspace/Flutter/libero/ios/opencv2.framework/Versions/A/opencv2[arm64][841](Converters.o)

d: 2 duplicate symbols
clang: error: linker command failed with exit code 1 (use -v to see invocation)
`

Can You help me found solution?

MarkerwApk:
I create PR: https://github.com/MaikuB/flutter_local_notifications/pull/2175

MaikuB:
Closing this as original PR hasn't had updates to it after leaving feedback

PR #2175, opened 2023-12-04
[flutter_local_notifications] Rename Converters - conflict with another libraries

MarkerwApk:
Fix: https://github.com/MaikuB/flutter_local_notifications/issues/2160

### DIFF
--- a/flutter_local_notifications/darwin/Classes/NotificationParser.h
+++ b/flutter_local_notifications/darwin/Classes/NotificationParser.h
@@ -1,5 +1,5 @@
 //
-//  Converters.h
+//  NotificationParser.h
 //  flutter_local_notifications
 //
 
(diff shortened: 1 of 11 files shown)


flutter_local_notifications/darwin/Classes/NotificationParser.h
```diff
@@ -8,7 +8,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface Converters : NSObject
+@interface NotificationParser : NSObject
```
MaikuB, line 11: Thanks for the PR. Rather than renaming from Converters to NotificationParser, would you be able to add the `FLN` prefix (an abbreviation of flutter local notifications) so that it's `FLNConverters` instead? Idea came about seeing the official plugins use `FLT` (presumably for Flutter)

MaikuB:
As it's been close to a year now without any updates, I'm going to close. Feel free to reopen if this gets updated

PR #2197, opened 2024-01-01
[flutter_local_notifications] Add custom repeat interval - periodicallyShowWithDuration method

mateuszluczak1996:
Add an option to set custom time for periodically showed notifications.

The problem that I have encountered on iOS devices is that repeating interval must be set to 60 seconds minimum, but I don't think that is the problem for real use case scenarios.

**Sample usage:**
```
    await flutterLocalNotificationsPlugin.periodicallyShowWithDuration(
      1,
      'repeating period title',
      'repeating period body',
      const Duration(minutes: 5),
      androidScheduleMode: AndroidScheduleMode.exactAllowWhileIdle,
    );
```

### DIFF
--- a/flutter_local_notifications/CHANGELOG.md
+++ b/flutter_local_notifications/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [17.2.0]
+
+* [Android]][iOS][macOS] added `periodicallyShowWithDuration()` method that allows for having a notification periodically shown based on a specified duration. The duration will need to be at least a minute. Thanks to the PR from [Mateusz Łuczak](https://github.com/mateuszluczak1996)
+
 ## [17.1.2]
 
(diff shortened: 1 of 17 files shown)


MaikuB:
Thanks for the PR. Will need time to give this one a good review compared to some other ones

MaikuB (changes requested):
Thanks. Got to test it out and works well. Besides the comment on the PR, do you think you'd be able to look at the following

- addressing the linter issues picked up by Dart analyzer
- I ran into the limitation you mentioned with Apple's APIs when the duration needs to be a minute. Rather than letting it run into a native exception that may not bubble up to be caught on the Dart side, could validation be added to ensure the duration is at least a minute on all the platform-specific implementations and throw an [ArgumentError](https://api.dart.dev/stable/3.3.0/dart-core/ArgumentError-class.html)? This would allow checking the argument up-front. If you're not sure how to do it, it'd be similar to how the ID is validated to ensure it's within range where there's a `validateId()` call in the plugin

MaikuB:
In the interest of merging the PR, I've made the changes and will have this merged in soon :)

yfujiki:
Great addition!! 

Only thing I want further is to specify initial time (currently it defaults to current time). But i guess it is limited by iOS side capability.

shrigashlesh:
Looks great!! About what @yfujiki said, can the offset datetime (say something like the zonedSchedule) be used in tandem with the periodic duration? It might be able to solve multiple scheduling scenarios.

pandey407:
Looks great!! About what @yfujiki said, can the offset datetime (say something like the zonedSchedule) be used in tandem with the periodic duration? It might be able to solve multiple scheduling scenarios.

Issue #2136, opened 2023-10-27
Crash: Attempt to invoke virtual method 'boolean com.dexterous.flutterlocalnotifications.models.ScheduleMode.useAllowWhileIdle()' on a null object reference


RuchiPurohit:
I have updated to version 16.1.0 from version 9.9.1, and I am getting hundreds of background crashes everyday. I am using zonedSchedule method. I have followed all the android setup guidelines in the readme section and have added the Proguard rules.

**I am in a fix, can't think of anything else to do. Please help.**

This is the background crash I am getting.

```
Exception java.lang.RuntimeException: Unable to start receiver com.dexterous.flutterlocalnotifications.ScheduledNotificationBootReceiver: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.dexterous.flutterlocalnotifications.models.ScheduleMode.useAllowWhileIdle()' on a null object reference
  at android.app.ActivityThread.handleReceiver (ActivityThread.java:4556)
  at android.app.ActivityThread.-$$Nest$mhandleReceiver
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2304)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loopOnce (Looper.java:240)
  at android.os.Looper.loop (Looper.java:351)
  at android.app.ActivityThread.main (ActivityThread.java:8377)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:584)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1013)
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.dexterous.flutterlocalnotifications.models.ScheduleMode.useAllowWhileIdle()' on a null object reference
  at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.setupAlarm
  at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.zonedScheduleNotification
  at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.rescheduleNotifications
  at com.dexterous.flutterlocalnotifications.ScheduledNotificationBootReceiver.onReceive
  at android.app.ActivityThread.handleReceiver (ActivityThread.java:4540)
```

[...]

PR #2199, opened 2024-01-03
[flutter_local_notifications] setupAlarm method: handle ScheduleMode.useAllowWhileIdle() called on null

RuchiPurohit:
…ence

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -598,7 +598,7 @@ public class FlutterLocalNotificationsPlugin
     AlarmManager alarmManager = getAlarmManager(context);
     if (notificationDetails.scheduleMode == null) {
       // This is to account for notifications created in older versions prior to allowWhileIdle
-      // being added so the deserialiser.
+      // being added to the deserialiser.
       // Reference to old behaviour:
       // https://github.com/MaikuB/flutter_local_notifications/blob/4b723e750d1371206520b10a122a444c4bba7475/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L569C37-L569C37
       notificationDetails.scheduleMode = ScheduleMode.exactAllowWhileIdle;
@@ -676,7 +676,7 @@ public class FlutterLocalNotificationsPlugin
 
     if (notificationDetails.scheduleMode == null) {
       // This is to account for notifications created in older versions prior to allowWhileIdle
-      // being added so the deserialiser.
+      // being added to the deserialiser.
       // Reference to old behaviour:
       // https://github.com/MaikuB/flutter_local_notifications/blob/4b723e750d1371206520b10a122a444c4bba7475/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L642
       notificationDetails.scheduleMode = ScheduleMode.inexact;
@@ -703,7 +703,7 @@ public class FlutterLocalNotificationsPlugin
 
     if (notificationDetails.scheduleMode == null) {
       // This is to account for notifications created in older versions prior to allowWhileIdle
-      // being added so the deserialiser.
+      // being added to the deserialiser.
       // Reference to old behaviour:
       // https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L735
       notificationDetails.scheduleMode = ScheduleMode.exact;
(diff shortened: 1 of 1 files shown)


flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -700,6 +700,13 @@ private static void setupAlarm(
       AlarmManager alarmManager,
       long epochMilli,
       PendingIntent pendingIntent) {
+
+    if (notificationDetails.scheduleMode == null) {
+      // Handle ScheduleMode.useAllowWhileIdle() called on a null object reference
+      // Issue reference: https://github.com/MaikuB/flutter_local_notifications/issues/2136
+      notificationDetails.scheduleMode = ScheduleMode.inexact;
```
MaikuB, line 707: Thanks for the PR but this looks like it hasn't been updated that I mentioned in https://github.com/MaikuB/flutter_local_notifications/issues/2136#issuecomment-1806769446 where the wrong schedule mode has been picked. It should be exact to retain the original behaviour as per https://github.com/MaikuB/flutter_local_notifications/blob/4b723e750d1371206520b10a122a444c4bba7475/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L549

MaikuB (changes requested):

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
RuchiPurohit: sorry, may be i didn't understand the requirement. do you want me to add this? 

`if (notificationDetails.scheduleMode == null) {
        checkCanScheduleExactAlarms(alarmManager);
        AlarmManagerCompat.setExact( alarmManager, AlarmManager.RTC_WAKEUP, epochMilli, pendingIntent);
    }`

MaikuB: That's not what I mean though is close. Currently you have specified inexact timing to be used via

```
notificationDetails.scheduleMode = ScheduleMode.inexact;
```

When I said this is wrong and should be exact instead, this means it should fixed to be

```
notificationDetails.scheduleMode = ScheduleMode.exact
```

This should be done instead of what you wrote as there would be code below to make sure exact alarm permissions have been granted as that needs to be done before trying to set an exact alarm based on that code you asked about.

Another thing is could you also change the code comment as mentioning the GitHub issue link it fixes isn't the important part. More important to link to https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L735 that has the code for the old behaviour. More specifically can you change it to be the following?

```
// This is to account for notifications created in older versions prior to allowWhileIdle
// being added so the deserialiser.
// Reference to old behaviour:
// https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L735
```

This would make the comment consistent with another similar fix

RuchiPurohit: great. Thanks. Will do.

RuchiPurohit: Hi. I have implemented the changes you suggested. Kindly review.

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -700,6 +700,15 @@ private static void setupAlarm(
       AlarmManager alarmManager,
       long epochMilli,
       PendingIntent pendingIntent) {
+
+    if (notificationDetails.scheduleMode == null) {
+      // This is to account for notifications created in older versions prior to allowWhileIdle
+      // being added so the deserialiser.
+      // Reference to old behaviour:
+      // https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java#L735
+      notificationDetails.scheduleMode = ScheduleMode.exact
```
MaikuB, line 709: Code's not compiling as you're missing a semicolon here

MaikuB (changes requested):

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
MaikuB: Left a comment elsewhere about how code isn't compiling

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
RuchiPurohit: I am so sorry. No idea how I missed a semicolon. I have pushed the changes.

MaikuB: Thanks. I've used your PR to fix existing typos I had in the comments too

MaikuB (approved):

PR #2212, opened 2024-01-16
[flutter_local_notifications] `AndroidFlutterLocalNotificationsPlugin` Consume selected notification

KirillBorodin:
Sets an extra to Android intent which indicates that selected
notification has been consumed.

The feature is a fix of the issue: #1926


giridat:
LGTM BUMP!

MaikuB:
Not sure if you sure this but a similar had already been raised at https://github.com/MaikuB/flutter_local_notifications/pull/2150 though your one is conveys what it does more clearly. With that in mind, I believe what I wrote https://github.com/MaikuB/flutter_local_notifications/pull/2150#issuecomment-1890351060 applies here too. Let me know what you think about changing how your PR works to match what I mentioned there

MaikuB:
@bitsydarel it applies to this PR as well as they both try to solve the same problem. What I'm referring to would also allow for an opt-in and the "cause" of this issue is developers have had assumptions before upon calling the `getNotificationAppLaunchDetails()`, a subsequent call to `getNotificationAppLaunchDetails()` wouldn't result in the same notification being returned  with the assumption that the plugin has consumed the notification. Part of this might be because the FCM plugin does this.

As there have been these expectations before, what I'm proposing is something similar to what you're saying but it's not based on updating the response and it also retains the opt-in behaviour. If you look at what I wrote in https://github.com/MaikuB/flutter_local_notifications/pull/2150#issuecomment-1890351060, I had described the proposal already. This was about updating the `getNotificationAppLaunchDetails()` to take an additional argument so that the plugin it would return and consume the notification in one go. This would be up to developers to opt-in and reduces the burden on developers using the API as there only be one API call to make compared the current approach in the PR that involves two separate API calls. If an app opted into this behaviour then wouldn't also be the need to check a `consumed` flag in the response as calling `getNotificationAppLaunchDetails()` again could just return null. Hope that is clearer now

PR #2249, opened 2024-03-04
Update README.md [flutter_local_notifications]

hassony105:
make highlight on the iOS and macOS custom notifications sound requirements because it does not clear.


MaikuB:
Thanks for the PR but highlighting like doesn't work for all browsers and I believe was Chrome-specific. Putting that aside, highlighting just that part of the text could cause readers to miss how there's text in the overview that talks about the location of file.  With in that mind, I would say it's better to have the person read everything so will close the PR

PR #2270, opened 2024-03-15
 [flutter_local_notifications] Add docs that desugaring is required for Android less than SDK 26

amrgetment:
Add docs that the desugaring step is required for Android less than Android 8 SDK 26 so the users can use Gradle 8 without any issues and new Java versions

Android 8 is used by 93% 
https://www.composables.com/tools/distribution-chart
image


amrgetment:
still it is required by the library to use desugaring even if I don't need it

Issue #1964, opened 2023-04-17
It's currently not possible to specify the intent (AudioAttributeUsage) when creating an AndroidNotificationChannel


jordyvanvorselen:
**Describe the bug**
It's not possible to specify the intent (AudioAttributeUsage) when creating a notification channel.

It however is possible to pass the intent when triggering a notification.

**To Reproduce**
1. Instantiate a new `AndroidNotificationChannel`
2. There is no way to specify the intent / `AudioAttributeUsage`

**Expected behavior**
There is the possibility to specify the intent / `AudioAttributeUsage`.

**Sample code to reproduce the problem**
 
        // Without a way of specifying the intent, we can not play the sound over the android alarm channel.
        // We need to create the notification channels explicitly, because in our usecase it should also be working
        // when the notification is received when the application is in the background.
        AndroidNotificationChannel(
          'alarm_notification_channel',
          'Alarm Notifications',
          description: 'This channel is used for alarm notifications.',
          importance: Importance.high,
          playSound: true,
          sound: RawResourceAndroidNotificationSound('alarm'),
          enableVibration: true,
          audioAttributesUsage: AudioAttributesUsage.alarm,  // This is currently not possible.
        ),

MaikuB:
Could you help submit a PR on this? Support for audio attributes originally came in via a community contribution (see https://github.com/MaikuB/flutter_local_notifications/pull/1649). Note that whilst it's not possible to specify it via the channel if it's created separately, as far as I know, it is possible to specify the usage/intent via `NotificationDetails` that if the channel was created as part of the process to trigger a notification, the channel will be created with the specified intent

jordyvanvorselen:
> Could you help submit a PR on this? Support for audio attributes originally came in via a community contribution (see https://github.com/MaikuB/flutter_local_notifications/pull/1649). Note that whilst it's not possible to specify it via the channel if it's created separately, as far as I know, it is possible to specify the usage/intent via `NotificationDetails` that if the channel was created as part of the process to trigger a notification, the channel will be created with the specified intent

Thanks for clarifying. Sure, I'll try to submit a PR for this soon.

MaikuB:
As it's been a very long time since this issue has been opened, I assume there's still no plan to submit a PR on this so will close

ditheshthegreat:
I have created a PR related this issue.

PR #2271, opened 2024-03-18
[flutter_local_notifications] Added AudioAttributesUsage for AndroidNotificationChannel

ditheshthegreat:
Added variable "audioAttributesUsage" in AndroidNotificationChannel  to create a channel with audioAttributesUsage in the beginning.
This is need when we need to alert sound on mute/silent.

Related to #1964

### DIFF
--- a/flutter_local_notifications/lib/src/platform_specifics/android/method_channel_mappers.dart
+++ b/flutter_local_notifications/lib/src/platform_specifics/android/method_channel_mappers.dart
@@ -53,6 +53,7 @@ extension AndroidNotificationChannelMapper on AndroidNotificationChannel {
         'ledColorRed': ledColor?.red,
         'ledColorGreen': ledColor?.green,
         'ledColorBlue': ledColor?.blue,
+        'audioAttributesUsage': audioAttributesUsage.value,
         'channelAction':
             AndroidNotificationChannelAction.createIfNotExists.index,
       }..addAll(_convertNotificationSoundToMap(sound));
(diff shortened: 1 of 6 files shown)


MaikuB:
Thanks for the PR. Would you be able to update the PR to fix the two unit tests that are failing?

flutter_local_notifications/lib/src/platform_specifics/android/method_channel_mappers.dart
```diff
@@ -53,6 +53,7 @@ extension AndroidNotificationChannelMapper on AndroidNotificationChannel {
         'ledColorRed': ledColor?.red,
         'ledColorGreen': ledColor?.green,
         'ledColorBlue': ledColor?.blue,
+        'audioAttributesUsage': audioAttributesUsage.value,
```
MaikuB, line 56: I had to leave a comment to mark the PR where changes are requested. FYI that these additions require updates to the unit tests that validate what gets passed to the platform channel

MaikuB (changes requested):

flutter_local_notifications/lib/src/platform_specifics/android/method_channel_mappers.dart
ditheshthegreat: I have created test case for audio attributes

MaikuB: Thanks, I pushed a commit directly to avoid back and forth. The commit was done as there was a change to the channel description that wasn't necessary to do

MaikuB (approved):

PR #2290, opened 2024-04-03
[various] Added bigText to ActiveNotification

vulpeep:


MaikuB:
Thanks for the PR. I made some updates so example app could also show the information and slight tweak to API doc for consistency with the docs for setting the `bigText` property

Issue #2177, opened 2023-12-07
iOS privacy manifest required


stuartmorgan:
This plugin is listed on https://developer.apple.com/support/third-party-SDK-requirements/ as requiring a privacy manifest. Some action will be required, either to clarify with Apple whether this plugin should be on the list, or to add a privacy manifest.

See https://github.com/flutter/flutter/issues/131940 for context.

MaikuB:
Will do. Unfortunately last time I got a generic answer to reference their docs even when I explained what my plugin does and asked if it was needed

yuk1ch1:
the reason might be that this package uses other packages listed by Apple here
Might be helpful to check this out:

https://developer.apple.com/support/third-party-SDK-requirements/

I'm not totally sure, but it looks like you might need to:
- update these dependency packages 
- signatures for SDKs
- prepare a PrivacyInfo.xcprivacy

[FYI]
> In [flutter/flutter#131940 (comment)](https://github.com/flutter/flutter/issues/131940#issuecomment-1881641272) it states, that plugins that do not collect data should include an empty privacy manifest. I'm assuming that holds true here too.

ref: https://github.com/pichillilorenzo/flutter_inappwebview/issues/1909#issuecomment-1883014512

MaikuB:
@yuk1ch1 it doesn't use other packages when it comes to iOS/macOS and only calls native APIs so is one of the reasons why I tried to ask for more details. I have seen the details of using an empty manifest so am aware of of that being a solution, just haven't gotten to it yet but will accept PRs on it

Miya49-p0:
@MaikuB 
hi

It seems that UserDefaults are being used in this package. It may still require PrivacyPrivacyManifest support. Could you please confirm this?

I'd appreciate it if you could check this.

MaikuB:
@Miya49-p0 Thanks for catching that. Sounds like updates to the manifest would be needed. If you happened to have already looked into this that you can submit a PR then that would be most welcome

PR #2294, opened 2024-04-12
[flutter_local_notifications] added reason for using User Defaults API to iOS privacy manifest

Miya49-p0:
In relation to #2177

# Reason
## flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m

I assumed that UserDefaults is being used for push notification settings in this code.

```objective-c
  NSDictionary *persistedPresentationOptions =
      [[NSUserDefaults standardUserDefaults]
          dictionaryForKey:PRESENTATION_OPTIONS_USER_DEFAULTS];
  bool presentAlert = false;
  bool presentSound = false;
  bool presentBadge = false;
  bool presentBanner = false;
  bool presentList = false;
  if (persistedPresentationOptions != nil) {
    presentAlert = [persistedPresentationOptions[PRESENT_ALERT] isEqual:@YES];
    presentSound = [persistedPresentationOptions[PRESENT_SOUND] isEqual:@YES];
    presentBadge = [persistedPresentationOptions[PRESENT_BADGE] isEqual:@YES];
    presentBanner = [persistedPresentationOptions[PRESENT_BANNER] isEqual:@YES];
    presentList = [persistedPresentationOptions[PRESENT_LIST] isEqual:@YES];
  }
```

## flutter_local_notifications/ios/Classes/FlutterEngineManager.m

It seems like it's storing handler information here.

```objective-c
- (void)registerDispatcherHandle:(NSNumber *)dispatcherHandle
                  callbackHandle:(NSNumber *)callbackHandle {
  [_persistentState setObject:callbackHandle forKey:@"callback_handle"];
  [_persistentState setObject:dispatcherHandle forKey:@"dispatcher_handle"];
}
```

# Result
Based on the usage scenarios, I believed UserDefaults would apply to "CA92.1", so I made that configuration. 
What do you think?

https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

### DIFF
--- a/flutter_local_notifications/ios/Resources/PrivacyInfo.xcprivacy
+++ b/flutter_local_notifications/ios/Resources/PrivacyInfo.xcprivacy
@@ -5,7 +5,16 @@
 	<key>NSPrivacyTrackingDomains</key>
 	<array/>
 	<key>NSPrivacyAccessedAPITypes</key>
-	<array/>
+	<array>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryUserDefaults</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>CA92.1</string>
+			</array>
+		</dict>
+	</array>
 	<key>NSPrivacyCollectedDataTypes</key>
 	<array/>
 	<key>NSPrivacyTracking</key>


MaikuB:
@Miya49-p0 thank you for creating a PR. I believe it should be C56D.1 as the plugin only wraps around the API and only uses access user defaults when the plugin APIs are called. From what I've seen at https://github.com/flutter/flutter/issues/139758#issuecomment-1852737278, the Flutter shared_preferences would've used it except it couldn't as its main purpose is the only wrap the APIs so doesn't meet the requirements

Miya49-p0:
@MaikuB 

I understand that Flutter's shared_preferences is being used, and the corresponding adjustments have already been made (https://github.com/flutter/packages/blob/main/packages/shared_preferences/shared_preferences_foundation/darwin/Resources/PrivacyInfo.xcprivacy).

The reason for this addition is due to the use of UserDefaults in flutter_local_notifications.


For example: 
・write
https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m#L366

・read
https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/ios/Classes/FlutterLocalNotificationsPlugin.m#L1287-L1288
These seem to be used independently of shared_preferences.


I thought the reason for this would be CA92.1, but what do you think about that?

MaikuB:
Based on your response, I suspect you have some misunderstanding of what I wrote. Your response would seem to indicate you think I'm saying the manifest update isn't needed but that's not what I'm saying. I mentioned raising a PR as I agreed that changes were needed.

The reason I brought up info on shared_preferences to tell you why I believe the reason should be C56D.1 and why the team maintaining shared_preferences couldn't pick C56D.1. They couldn't pick it as the main purpose of the shared_preferences plugin is that the only purpose for its existence is to act as a wrapper for the user defaults API. Apple's guidelines say if that is the case then it can't use C56D.1. However, flutter_local_notifications purpose is to wrap the Apple notification APIs. 

Hope that is clearer now but if you still don't understand (e.g. due to English not being your first language), then the summary is I'm saying that if PR is updated to use  C56D.1 instead of CA92.1 then I believe it's good to merge.

MaikuB:
Sorry actually, on further reading, I noticed C56D.1 says the following

> Information accessed for this reason, or any derived information, may not be used for your third-party SDK’s own purposes or sent off-device by your third-party SDK.

The data saved to user defaults is only by this plugin (i.e. Apple's definition of third-party SDK). I believe based on that the reason you picked is correct then so merge that in. Apologies for confusion

Miya49-p0:
Sorry. Understood now.

>  flutter_local_notifications purpose is to wrap the Apple notification APIs.
I see. Indeed

>  e.g. due to English not being your first language
that's right
Sorry for my unfamiliar English.



Thank you for your response!!

PR #2301, opened 2024-04-16
Dev atique

atiquehossain:
Hey there, it seems like my office is sticking with Flutter 2.2.3 for now. We're also using an older version of local_notification (6.1.1). I've noticed that errors occur unless we include android:exported="true" in the AndroidManifest. 
Thanks for making this awesome plugin!

### DIFF
--- a/flutter_local_notifications/CHANGELOG.md
+++ b/flutter_local_notifications/CHANGELOG.md
@@ -1,9 +1,8 @@
-# [7.0.0]
+# [6.1.1]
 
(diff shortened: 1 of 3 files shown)


MaikuB:
Thanks for the PR however, A PR can't be done specifically for the codebase that aligns with the older version of the plugin. The changes in this PR were already applied on newer versions and if you choose to remain on an older version then you need to use your own fork for it so will be closing this

atiquehossain:
Thank you, Michael Bui for your response.

Best regards,
Atique Hossain

PR #2308, opened 2024-04-21
[flutter_local_notifications] Reorganize the readme in to smaller digestible blocks

davetrollope:
When I tried to adopt the flutter_local_notifications package, I found it difficult to work through the readme. This PR proposes isolating the larger platform specific blocks in to separate files making the overall readme easier to follow.

I have updated to include the latest 17.1.0 updates to the README.


MaikuB:
Thanks for the PR. I understand the reason behind the changes and for it but  will take me some time to digest to see how the changes work. Do you by chance know how this will work on pub.dev and if there similar examples on pub.dev? Mentioning it as a readme will show up on the readme section and I don't know what would happen if a user was to click on, say, the link to the Android markdown file

davetrollope:
I am not sure about pub.dev, but I think it follows standard markdown. I will investigate and see if I can find some examples to verify.

davetrollope:
After doing some research, it looks like relative links are expected to be rewritten to github URLs. That means following the links would hop to the github view of the page, e.g. https://github.com/MaikuB/flutter_local_notifications/blob/49dc49d7ab19ceb6afcf99bfb84d5807c515aaf7/flutter_local_notifications/Android.md

I'm in two minds about this.

1) It might be a bit disconcerting to be forced over to github for subsections, and then have to hit back or close the tab to continue the overall flow of setup - at least as the subpages are written right now. Perhaps some nuanced updates recognizing this would make it intuitive.
2) perhaps keeping everything in one readme is better, but maybe we should reorganize the file in to complete sections/walkthroughs for a platform (perhaps with a badge that identifies common setup?

Open to thoughts and ideas.

davetrollope:
Yes, I like the idea of that. I will start playing with it.

Levi-Lesches:
I second the idea of a smaller main README linking to other pages with more details. Lets readers understand the bigger picture faster while still getting to choose when to zoom in on a specific platform, function, etc.

Levi-Lesches:
@MaikuB I think this should be re-prioritized as a lot of the issues opened recently are the result of people not reading the README. I understand that's just part of the human condition, but a much shorter and platform-specific guide should help

Levi-Lesches:
I started work at #2477 to not only separate the content by platform, but also reorganize each topic, rewrite it for clarity, include more links, more full examples, and removed content that applies to Flutter or Android versions this package no longer supports.

PR #2331, opened 2024-05-29
[flutter_local_notifications] Add "selectable" property for AndroidNotificationDetails to disable click actions on the notification

tranleo148:
This PR introduces a new option to disable click actions on notifications from Flutter. In Android, disabling click actions on a notification is achieved by setting the contentIntent to null when building the notification. Currently, there is no option in Flutter to disable the click action.

To address this, I have added a new `selectable` property that allows developers to control whether the notification is selectable or not. When selectable is set to false, the `contentIntent` will be set to `null`, disabling click actions on the notification.

* Default Behavior:
The default value of `selectable` is `true`, ensuring that existing behavior is not affected unless explicitly changed.
* Impact:
This change is backward-compatible and will not affect notifications unless the new property is utilized.


flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -274,7 +274,10 @@ protected static Notification createNotification(
                     : notificationDetails.body)
             .setTicker(notificationDetails.ticker)
             .setAutoCancel(BooleanUtils.getValue(notificationDetails.autoCancel))
-            .setContentIntent(pendingIntent)
+            .setContentIntent(
+                BooleanUtils.getValue(notificationDetails.selectable)
```
MaikuB, line 278: Compared to other parts of the code, I don't believe you should be using `BooleanUtils.getValue()`. That is for coalescing null values to false. The null values would happen when introducing new fields/properties and there are scheduled notifications kept in shared preferences. With the changes introduced in the PR, my understanding is that null `selectable` value be coalesced to `true` instead of `false`

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -237,6 +238,14 @@ class AndroidNotificationDetails {
...
   /// on it.
   final bool autoCancel;
+  
+  /// Specifies if the notification is selectable.
+  /// 
+  /// If `selectable` is `false`, the `contentIntent` will be set to `null`, 
+  /// disabling click actions on the notification.
+  /// 
+  /// The default value is `true`.
+  final bool selectable;
```
MaikuB, line 248: - Can you update the example app to include a scenario that makes use of this?
- I assume the `onSelectNotification` callback won't fire. Can you confirm? If so, can you update the API docs to explicitly add mention of this. Some developers may not realise the implication based on reading what's there

MaikuB (changes requested):
Thanks for the PR. Left a couple of comments on there. Could you take a look?

MaikuB:
@tranleo148 I saw a notification that updated your branch. In case you missed it, I requested changes on this PR a while back. Are you going to be able to take a look?

Issue #481, opened 2020-02-18
Flutter Web support


dawsonc623:
Are there plans to support Flutter Web at any point? There is a Notification API in JavaScript, so I imagine it is possible.

MaikuB:
Not likely at this stage. I looked at how the notifications API works on the web and every callback is tied to each instance of a notification so the model is quite different. It looked like the only thing that can be done is to show a notification immediately i.e. there's no way to schedule a notification. As such a developers would most likely need to put a guard in their code when running if web support was provided in the plugin.

This leads me to think there isn't much value in providing web support through this plugin as you could use the notifications API directly via Dart with full functionality and in a similar manner if the plugin were to support. Could be I missed seeing a use case for adding it the plugin so happy to be persuaded but note that web support would need contributions from others.

dawsonc623:
Thanks for the response. Admittedly, I have not looked into what it would take. I can think of a few "clever" things that could be done (such as using `setTimeout` for a "scheduled notification"), but I would actually have to spend some real time thinking through all of the nuances.

I think having at least partial support for notifications in web would be worthwhile - especially the instant notifications. The most fundamental use case I can think of is communication systems where the app may notify a user whenever they receive new communication.

Having a guard in place for detecting whether or not certain features are in place is not uncommon in the web, anyway. Feature detection is kind of a fact of life when running on many platforms. In fact, this library already has some Android-only and iOS-only things that need to be addressed by the user, so I do not think that is going to be a huge concern. The added benefit, however, is a consistent API for the features that are available across multiple platforms. There would just need to be solid documentation for what is available on which platforms.

As for getting it done, I have never written a Flutter plugin before, and I am admittedly new to Flutter and the ecosystem in general. However, since it is likely I will have to find a way to support web notifications for my project anyway (it is for work, so if it is a requirement then I have no choice) I figure I might as well do it in a way others can benefit from, as well.

hiashutoshsingh:
I am using this in web it works fine...is there any enhancement needed in this ? @MaikuB

MaikuB:
@hiashutoshsingh i'm not sure how you are using this plugin on the web when it doesn't support it to begin with.

gedw99:
Regarding the scheduling just use hive with a timer

CodeDoctorDE:
Any updates?

giorgio79:
It is definitely not working on web, getting this error
[...]

PR #2349, opened 2024-06-25
[flutter_local_notifications] Windows support

Levi-Lesches:
Hi, I'm picking this up from @kennethnym and @lightrabbit. From their work, we already have basic initialization set up. I'm referring to https://learn.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/adaptive-interactive-toasts?tabs=appsdk and https://learn.microsoft.com/en-us/uwp/schemas/tiles/toastschema/schema-root for the rest of my implementation. 

@MaikuB, There's a small TODO below, but feel free to start reviewing!

TODO: 

- [x] Validate message passing back to Dart
- [x] ~~Check Windows version before playing audio~~ 
- [x] update progress bars dynamically
- [x] validate that headers work as intended
- [x] Remove title, body, and payload from the C++ side
- [x] Implement getActiveNotifications and any other missing methods
- [x] Update example to use Windows notifications as well
- [x] Test `getActiveNotifications` on Windows MSIX
- [x] Test `pendingNotificationRequests`
- [x] Consolidate Notification --> XML --> Notification logic
- [x] Check message passing in the background
- [x] Clean up the C++ code
- [x] Support multiple progress bars
- [x] Remove `platformSpecifics` from method channel
- [x] Remove path dependency on platform interface package
- [x] Prevent recurring notifications in example app when on Windows
- [x] Think of safer ways to handle file URIs 
- [x] Look into `GET_NOTIFICATION_APP_LAUNCH_DETAILS` support
- [x] Test action image and image logic on exe and MSIX

Note: `WindowsImage` requires a Windows-formatted URI of an absolute path. Instead of putting that on the user, I accept a `dart:io` `File` object and do that internally. That marks the first use of `dart:io` inside this package, but: 
- it does not support web anyway (until #481)
- simply _importing_ `dart:io` does not crash a build anymore, but _using_ it does
- the file isn't actually used until `.toXml()` is called, which is only called by the Windows plugin.

If, in the future, this package supports web, people can still use `File` in their `WindowsNotificationsDetails`, and the app will compile, but the absolute path will not be looked up unless they're on Windows.


Levi-Lesches:
As an aside, @MaikuB, can I recommend disabling two lints? I found these _reduced_ the quality of the code at times: 
- `always_specify_types` leads to a lot of `final VeryLongClassName className = VeryLongClassName();`
- `lines_longer_than_80_chars`, especially when combined with the above, leads to simple lines getting split across many. I typically stick to ~100 characters per line with some rare exceptions when they help more than they hurt. 

Just a recommendation, feel free to ignore. My PR conforms to the lints you chose without disabling anything.

Levi-Lesches:
Also, I built off the work from Kenneth and lightrabbit, but I was wondering about a v2 of this plugin with pigeon or FFI. Since this version is already functional, let's try to get this out first, but just a thought for later.

Levi-Lesches:
@MaikuB All done! I don't anticipate making any more API changes from here, maybe just some better documentation.

MaikuB:
Thanks. Note I do want to prioritise a beta that includes changes like bumping the minimum Flutter version and potentially get rid of deprecated properties that are on Android side.

When I can look at this one, it's possible I reach out here on how to test as well as I've not been involved in Windows development since UWP was first introduced

Levi-Lesches:
> Thanks. Note I do want to prioritise a beta that includes changes like bumping the minimum Flutter version and potentially get rid of deprecated properties that are on Android side.

Sounds reasonable, I'll do my best to keep this PR up to date with any potential merge conflicts. 

> When I can look at this one, it's possible I reach out here on how to test as well as I've not been involved in Windows development since UWP was first introduced

I've updated the example to showcase all the new features version is capable of and hide unsupported examples when running on Windows, so if you have one, testing is as easy as `flutter run -d windows`. If you can't test on Windows, let me know and I can upload a video. 

Do you have an ETA on a review? No rush, because I'm using a pub override to my local copy, but I'm still thinking of possibly rewriting this as an FFI plugin (C++ logic would stay the same, but we'd drop all the serialization and method channel logic). However, I'm satisfied with how the plugin works now (adding FFI won't change the functionality), and it's about time I got back to the project that needed Windows notifications in the first place, so a rough timeline of when you want to go through this would be helpful.

MaikuB:
> I've updated the example to showcase all the new features version is capable of and hide unsupported examples when running on Windows, so if you have one, testing is as easy as flutter run -d windows. If you can't test on Windows, let me know and I can upload a video.

I mentioned I might reach as I think your original message mentioned some limitations that relate to MSIX (?)

> Do you have an ETA on a review? 

If there's a plan to rewrite this to use FFI then I'd prefer to wait until that is done. Reason for this is because I maintain as a spare-time effort (e.g. check PRs on weekends) so would be better for me to avoid reviewing and test functionality twice

Levi-Lesches:
> If there's a plan to rewrite this to use FFI then I'd prefer to wait until that is done. 

I pinged you there, but for people following, I pushed #2366 for the FFI version. TL;DR, it's much cleaner, less Flutter magic, safer, and faster to write with automatically generated bindings. The (minor) downsides are needing a C-compatible API (no vectors or maps for data that really wants it), and notification callbacks needing Dart 3.1 or higher.

Levi-Lesches:
Closing for now with the FFI implementation in #2366. I like the code there much better, but there's an important note there about Dart versions.

Issue #2329, opened 2024-05-25
Error bigPictureStyle.bigLargeIcon(null) v17.1.2


JaTocora:
flutter_local_notifications-17.1.2/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java:1033: error: reference to bigLargeIcon is ambiguous
      bigPictureStyle.bigLargeIcon(null);
                     ^
  both method bigLargeIcon(Bitmap) in BigPictureStyle and method bigLargeIcon(Icon) in BigPictureStyle match

godzyken:
pareil quelqu'un à t'il resolu ce probleme?

MaikuB:
Whilst I understand why you're seeing this, this isn't something I've seen that can reproduced and hasn't happened via the pipelines. I suspect it might be an issue specific with the Android SDK you're using

ducdx-0072:
I faced this problem too, and I don't understand the reason ;(

MaikuB:
Which versions of the Android SDK do you have installed? I only have the ones for Android 13 and 14 installed and don't have this issue

godzyken:
J'ai aussi 13 et 14

Envoyé à partir de Outlook pour Android<https://aka.ms/AAb9ysg>
________________________________

MaikuB:
@godzyken could you write in English please?

mark8044:
> @godzyken could you write in English please?

I just had this issue for the first time today.

I was on master branch `Flutter 3.23.0-13.0.pre.300` and 17.1.2 perfectly fine, then upgraded to master branch `Flutter  3.23.0-13.0.pre.370` and I got the error.

I moved to 17.2.0 and it still occurrs

I downgraded to master branch `Flutter 3.23.0-13.0.pre.300` and its gone.   

This only occurs on Android.  iOS remains ok.

I think something has changed in latest flutter framework, so I believe this needs attention

gnprice:
Very helpful, thanks @mark8044 for bisecting!

That commit changes `bin/internal/engine.version` from flutter/engine@d1506c12808ec42e1b277376c4c2b1bd4ec20ac7 to flutter/engine@ddd4814b9d407962eee4dc52919e3abddb0c3641.

Looking at the Git log between those, it's just three commits:
flutter/engine@3f40c6875 Fix AccessibilityFeatures.disableAnimations flag on Android 12+ (flutter/engine#53428)
[...]

PR #2355, opened 2024-06-28
[flutter_local_notifications] Fix crash on latest Flutter master due to ambiguous method call

gmackall:
Fixes https://github.com/MaikuB/flutter_local_notifications/issues/2329, by casting the null argument to be a `Bitmap` so that we choose

https://developer.android.com/reference/android/app/Notification.BigPictureStyle#bigLargeIcon(android.graphics.Bitmap)

of the two methods.

Verified by running the example app.


MaikuB:
Thanks for the PR. Was about to create a branch to ask others to test so much appreciated for this.

Would you be able to update your branch with latest changes from master though? I noticed PRs created weren't running th the formatting jobs. I merged some changes that should fix this now but would like to confirm it's firing not just for me but others too

mark8044:
Can confirm the fix works on latest flutter master.
Awesome turn around time!  Thanks much

PR #2362, opened 2024-07-03
[flutter_local_notifications_linux] Changed conditional import (required for WASM compilation) 

AlexBacich:


MaikuB:
Do you have more context on what this was supposed to solve? Is it do with https://github.com/dart-lang/sdk/issues/55948? If so, my understanding was to the 3.22 release where WASM was started being considered stable that the fix would be to update the Flutter SDK

AlexBacich:
@MaikuB , thanks for your input. 
With current Channel stable, 3.22.2 it is not working. But yes - eventually the fix will be in future releases. 
https://github.com/flutter/flutter/issues/149984

MaikuB:
Thanks for the reply. Going to close this in favour of the tooling being fixed. The comments I've seen would indicate this was more a problem with the conditional import not working as intended and returning the wrong response. Furthermore, I've seen more info on using `` as a way to detect FFI if is supported.  It's possible I've missed it but I've not seen info that indicates checking for `dart.library.io` is another way to confirm if a platform supports FFI. If anything, I've seen more to indicate `dart.library.io` is more generic and covers mobile use cases as well. The first-party plugins still make use of `dart.library.ffi`

PR #2366, opened 2024-07-11
[flutter_local_notifications_windows] Windows FFI plugin  

Levi-Lesches:
**Update**: I went in and simplified the code since yesterday, it's now in a reviewable state. If it would help, I can rebase + commit the first 50 or so commits and start with a cleaner slate.

This is a spin off of #2349, which implements the same logic but with an FFI plugin instead of method channels. I'm happy with it so far, here's the gist: 

### The C/C++ code

- `src/ffi_api.h`: a C API between C++ and Dart
- `src/plugin.hpp`: A C++ class that holds Windows-specific API handles
- `src/ffi_api.cpp`: C++ code to implement the C API using the C++ class

### The Dart code

- a new package, `flutter_local_notifications_windows` that implements the same functionality as before. 
- `lib/src/details` holds all the platform-specific details
- `lib/src/ffi` holds basic FFI stuff, like generated bindings and utils
- `lib/src/plugin` holds the Windows plugin interface, the real FFI implementation, and a stub, because `dart:ffi` will break web apps (note, the API does reference `dart:io` but it still compiles on web)

### Notes

- Overall, the only thing that changed was how to go from Dart <--> C++. Before, it was all JSON-based message passing, and a _big_ `HandleMethodCall` function that had to handle every possibility. I'm glad that's gone. Now we have a similar mechanism with Dart <--> C <--> C++ and `package:ffigen` sets everything up for us.
- It's all backed with compile-time safety as opposed to the runtime failures I was facing with method channels. For example, passing a string now looks like `"hello".toNativeString()`, but other types require allocating memory manually.
- Functionally, everything is working how it was before, with the method channels

### Benefits
  - the C++ code is ~150 lines smaller, and went from being one big 250 line file to two 150 line files, one of which can be skimmed as it's just Windows registry config. 
  - The C++ code no longer uses method channels or any of that associated logic. There's a lot of magic there involving serialization which also comes with a runtime cost. The new code directly passes values and pointers from Dart to C++, eliminating all the message passing. In fact, the new API is inherently synchronous instead of async (not that it matters in this context, since the other platforms are still async).
  - The Windows implementation now lives in `flutter_local_notifications_windows` instead of hijacking the original package and half-sharing some of the original implementations. That means no more platform checks, all the logic can be safely dealt with in a Windows-only context. 

### Summary of changes
  - Windows C/C++ code: +600 lines
  - Windows Dart API for customizing notification details): +800 lines
  - Windows Dart plugin code, excluding generated code: +300 lines
  - Main plugin integrations: < 50 lines
  - Example Dart code: +500 lines

That's only ~2,200 additions. the other half GitHub is reporting is automatically generated, like the example's new `windows` folder. 

> [!IMPORTANT]  
> FFI is a more recent development than method channels. As such, some more advanced FFI features are locked behind some more recent Dart versions. I already had to downgrade `package:ffigen` from a beta 13.0 to a whopping _7.0_, but more pressing is that for C to call a Dart function from another thread (eg, when a notification is pressed), you need to use features from Dart 3.1. The current constraint is only 2.17. 
> 
> Including this new implementation will force Dart ^3.1.0 on end-users. At this point, 2.17 is two years old, 3.1 is about one year old, and null safety is almost 3.5 years old. Age aside, I also don't believe going from 2.17 to 3.1 is so burdensome, especially since 2.17 is already after null-safety, and the Dart team's official position is that 3.0 is not really a breaking change in practice. In other words, I'm not sure I see a situation where someone is really stuck. They can either upgrade any pre-null-safety code, not upgrade this package further, or fork this package to add null Windows or future updates as needed. 
> 
> Overall, I'd recommend bumping the SDK version of the front-facing package to 3.1.0 as well to better advertise this, and to benefit from any features, since 2.17 and beyond -- better null promotion, FFI for mobile platforms, class modifiers, macros, and more.


MaikuB:
FYI I've merged in https://github.com/MaikuB/flutter_local_notifications/pull/2373 that bumps the minimum Flutter SDK requirement to 3.13. From my understanding, this uses Dart 3.1. The other breaking changes I mentioned I was going to do that relate to the Android side was merged in as well

Note I still need to setup Windows as a dev environment. Looks like Flutter doesn't support ARM for Windows yet for me to use my macBook. However, I do have a PC I could set setup. In the meantime, I'll see what I can comment on for the PR.

General comments I have so far

- Is it possible for the PR to update the GitHub workflow such that it can build the example app for Windows?
- Are there unit tests that can be added/run?

Levi-Lesches:
Oh, good to know about the breaking sdk change, I was afraid that would be a blocker. What motivated that? Just curious. 

I can work MSIX packaging into the CI, sure. Two features (cancel by ID and get active notifications) don't work without being an msix so having that could be useful. Seems like it's just a few lines in melos.yaml. Is that being stored or published anywhere? Or is it just to verify that the examples still build? 

I didn't add tests or major documentation yet. For docs, I added dart docs to all new code and some places that have different behavior on windows, but the main readme would still need to be reviewed again for windows details. 

How should I go about unit tests? Notifications are visual, and the tests can't check active notifications due to not being run as an msix app. I can do some smaller tests, like:
- assert invalid guids throw properly
- assert invalid XML throws properly
- try all notification details and make sure nothing throws 
- schedule a notification, check it's been scheduled. Wait, check that nothing is scheduled anymore 

Am I missing any bigger picture test cases here?

MaikuB:
> Oh, good to know about the breaking sdk change, I was afraid that would be a blocker. What motivated that? Just curious.

To make the plugin easier to maintain. This also factors in how Flutter SDK updates are done and handles changes to Android ecosystem too e.g. when new Android Studio and AGP versions are released.

> How should I go about unit tests? Notifications are visual, and the tests can't check active notifications due to not being run as an msix app. I can do some smaller tests, like

The visual side would be more of integration/UI testing so wouldn't be unit tests that I'm referring to. There are existing examples in repo and they check that appropriate details are passed via the platform channels. Note this was an open-ended question so it's not like I was calling out specifics but if possible/applicable, it's that kind of testing I was referring to. Based on what you shared, I think the first three bullet points align with that. The scheduled notification scenario sounds like an integration test. Not against having that but due to potential difficulty on covering integration tests, I hadn't asked about that

Levi-Lesches:
> The visual side would be more of integration/UI testing so wouldn't be unit tests that I'm referring to. There are existing examples in repo and they check that appropriate details are passed via the platform channels. Note this was an open-ended question so it's not like I was calling out specifics but if possible/applicable, it's that kind of testing I was referring to.

Got it. The nice thing with FFI is that with how much simpler it is VS method channels, the data is being passed along to the native sides just as if they were plain Dart functions. So I'm not sure how much testing is needed or even possible there, but I can definitely work on those sanity checks I mentioned in the bullets. Not sure if it's possible either to test that tapping on notifications triggers the callback", but like you said that's more of an integration test, and the example app should cover that pretty well. 

One more problem I'm noticing is that CI is rejecting this because `package:flutter_local_notifications_windows` is not on Pub yet. We can either publish now and not actually merge it into main branch until later, or introduce a pubspec override until we're ready to merge (either way, CI won't like it until there's no override and it's on Pub).

MaikuB:
The Windows package doesn't need to be published on pub.dev yet. The plugin and CI uses https://melos.invertase.dev to setup the appropriate "local" pubspec overrides when dealing with mono repos like this one. AFAIK, should be a matter of adding the Windows package to the melos workspace. See https://github.com/MaikuB/flutter_local_notifications/blob/master/melos.yaml where you can find the other packages listed

MaikuB:
Something else that came to mind regarding unit tests is that checking other FFI-based plugins can provide some pointers based on what they've done

Levi-Lesches:
@MaikuB I've added unit tests, configured things with melos to the best of my knowledge, and updated the example some more. Now it builds into a `.msix` file, which should help showcase the how to do that. 

If you need the MSIX for yourself I've uploaded one [here](https://github.com/Levi-Lesches/flutter_local_notifications/releases/download/v1.0/example.msix). Let me know if you need anything else to review.

Levi-Lesches:
Another update: I added docs to both READMEs, the one on the `_windows` repository and the one on the main repository. 

I also managed to remove the dependency on Flutter for the Windows package, which means anyone can use it for Dart-only applications!

A note on unit tests, they are run concurrently, but the logic isn't meant to be thread-safe. That leads to two things: 
- a flaky test with updating notifications, because sometimes the notification to be updated may be cancelled by a different test. I added 5 retries so it still passes. It only fails occasionally and its only ever needed 1 retry
- To fix this, I tried running `dart test -j 1`, but this still runs tests in different threads/isolates. 
- I added a file, `bin/test.dart` that shows an interesting crash that involves two isolates running two tests. Not sure what to do with it but I left it in

I don't think we need to _fix_ the failing test because the retries are enough and no user is going to be using this library in that way. Just something to be aware of.

MaikuB:
Thanks will try to take a look through this weekend. Whilst I have a PC, I'd prefer to get this setup on my MacBook Pro by having Windows 11 ARM running via Parallels. Do you happen to know if this would work with Flutter?

Levi-Lesches:
Seems they support building standard x64 executables and running in emulation mode. Here's the issue for building arm64 executables: https://github.com/flutter/flutter/issues/62597

I saw the ci issues, I'll get to those shortly

MaikuB:
Thanks. I was able to get this up and running though. I may not get to review everything this weekend so to will leave comments as I go through it

flutter_local_notifications_windows/lib/src/details/notification_details.dart
```diff
@@ -0,0 +1,162 @@
...
+
+  /// Custom bindings in the notification.
+  ///
+  /// Text elements can contains "bindings", which are entered as `{bindingName}` directly into the
+  /// string values. You can then update them while or after the notification is launched by
+  /// using the binding name as the key here, and the value as any string you want
+  final Map<String, String> bindings;
+
+  /// XML attributes for the toast notification as a whole.
+  Map<String, String> get attributes => {
```
MaikuB, line 119: From what I could tell, it looks like `attributes`, `toXml()` and `generateBindings()` is really only for internal use when the plugin converts the notification to the XML representation. However, all three are currently publicly exposed. I suspect these be made hidden and would reduce the chance of an inadvertent breaking change being made

flutter_local_notifications_windows/lib/src/details/notification_to_xml.dart
```diff
@@ -0,0 +1,39 @@
+import "package:xml/xml.dart";
+
+import "notification_details.dart";
+
+/// Converts a notification with [WindowsNotificationDetails] into well-formed XML.
+///
+/// For more details, refer to the [Toast Notification XML schema](https://learn.microsoft.com/en-us/uwp/schemas/tiles/toastschema/schema-root).
+String notificationToXml({
```
MaikuB, line 8: The entire file is being exported and means `notificationToXml()` is exposed by the plugin and suspect it's better off to not to do so unless I missed why this end for

flutter_local_notifications_windows/lib/src/details/notification_details.dart
MaikuB: Looks like there are other classes in a similar situation and I left a related comment on `notification_to_xml.dart'. Is it possible to move the conversion/mapping logic to that file and avoid exposing the methods in that class?

flutter_local_notifications_windows/lib/src/plugin/base.dart
```diff
@@ -0,0 +1,67 @@
...
+  /// Initializes the plugin. No other method should be called before this.
+  Future<bool> initialize(
+    WindowsInitializationSettings settings, {
+    DidReceiveNotificationResponseCallback? onNotificationReceived,
+  });
+
+  /// Releases any resources used by this plugin.
+  void dispose();
+
+  /// The raw XML passed to the Windows API.
```
MaikuB, line 20: Something seems off here. The comments say it's for the raw XML as though this method  returns the XML representation of the notification. However, it looks like it's to show a notification based on the raw XML representation. Even if it's for showing a notification based on the raw XML representation, is there a reason why this method exists? Would imply it's not possible to do other things like schedule a notification using the raw XML representation. I would've thought that the raw XML is an alternative representation. If so, semantically speaking  it would IMO be more appropriate that the `WindowsNotificationDetails` provides a named constructor that allows providing a raw XML

MaikuB:
FYI I've noticed that when I debug the example app that the following can be seen in debug output

```
Launching lib\main.dart on Windows in debug mode...
√ Built build\windows\x64\runner\Debug\example.exe
Connecting to VM Service at ws://127.0.0.1:54517/CrhXIgkgKHg=/ws
Unable to parse JSON message:
The document is empty.

════════ Exception caught by services library ══════════════════════════════════
The following assertion was thrown during a platform message callback:
Attempted to send a key down event when no keys are in keysPressed. This state can occur if the key event being sent doesn't properly set its modifier flags. This was the event: RawKeyDownEvent#f2796(logicalKey: LogicalKeyboardKey#f3b7b(keyId: "0x10000010a", keyLabel: "Num Lock", debugName: "Num Lock"), physicalKey: PhysicalKeyboardKey#5b142(usbHidUsage: "0x00070053", debugName: "Num Lock"), repeat: false) and its data: RawKeyEventDataWindows#32fe2(keyCode: 144, scanCode: 57413, characterCodePoint: 0, modifiers: 0)
'package:flutter/src/services/raw_keyboard.dart':
Failed assertion: line 862 pos 7: 'event is! RawKeyDownEvent || _keysPressed.isNotEmpty'

Either the assertion indicates an error in the framework itself, or we should provide substantially more information in this error message to help you determine and fix the underlying cause.
In either case, please report this assertion by filing a bug on GitHub:
  https://github.com/flutter/flutter/issues/new?template=2_bug.yml

When the exception was thrown, this was the stack:
#2      RawKeyboard.handleRawKeyEvent (package:flutter/src/services/raw_keyboard.dart:862:7)
#3      KeyEventManager.handleRawKeyMessage (package:flutter/src/services/hardware_keyboard.dart:1161:30)
#4      BasicMessageChannel.setMessageHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:235:49)
#5      _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:581:35)
#6      _invoke2 (dart:ui/hooks.dart:344:13)
#7      _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:45:5)
#8      _Channel.push (dart:ui/channel_buffers.dart:135:31)
#9      ChannelBuffers.push (dart:ui/channel_buffers.dart:343:17)
#10     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:750:22)
#11     _dispatchPlatformMessage (dart:ui/hooks.dart:257:31)
(elided 2 frames from class _AssertionError)
════════════════════════════════════════════════════════════════════════════════
```

flutter_local_notifications_windows/lib/src/plugin/base.dart
```diff
@@ -0,0 +1,67 @@
...
+    String? body,
+    TZDateTime scheduledDate,
+    WindowsNotificationDetails? details, {
+    String? payload,
+  });
+
+  /// Updates the progress bar in the notification with the given ID.
+  ///
+  /// Note that in order to update [WindowsProgressBar.label], it must
+  /// not have been set to null when [show] was called.
```
MaikuB, line 52: Presumably the a notification with a progress bar can be scheduled to be shown e.g. using `zonedSchedule()` and that it doesn't require the notification to be shown immediately via `show()`

flutter_local_notifications_windows/lib/src/plugin/ffi.dart
```diff
@@ -0,0 +1,196 @@
...
+
+extension on String {
+  bool get isValidGuid => length == 36
+    && this[8] == "-"
+    && this[13] == "-"
+    && this[18] == "-"
+    && this[23] == "-";
+}
+
+/// The FFI implementation of `package:flutter_local_notifications`.
```
MaikuB, line 22: Didn't know it's possible to reference a package so thanks for that. However, can this be updated to `The Windows implementation of [FlutterLocalNotificationsPlatform]` and then have a reference to the main plugin package in a separate paragraph (for example, like https://github.com/flutter/packages/blob/991c70177e8ff145f7d1c500d9dff86c6c7f92f4/packages/shared_preferences/shared_preferences_windows/lib/shared_preferences_windows.dart#L18)? Reason for pitching this is FFI  sounds too generic. I believe the Linux implementation could technically be said to use FFI as it depends on `dbus` package that in turn makes use of FFI. 

I just realised other parts of the plugin like the Linux one could do with similar changes so I'll note this to later as it's not tied to this PR

flutter_local_notifications/README.md
```diff
@@ -59,8 +59,9 @@ A cross platform plugin for displaying local notifications.
 * **iOS**. Uses the [UserNotification APIs](https://developer.apple.com/documentation/usernotifications) (aka the User Notifications Framework)
 * **macOS**. On macOS versions older than 10.14, the plugin will use the [NSUserNotification APIs](https://developer.apple.com/documentation/foundation/nsusernotification). The [UserNotification APIs](https://developer.apple.com/documentation/usernotifications) (aka the User Notifications Framework) is used on macOS 10.14 or newer. Notification actions only work on macOS 10.14 or newer
 * **Linux**. Uses the [Desktop Notifications Specification](https://specifications.freedesktop.org/notification-spec/)
+* **Windows** Uses the [C++/WinRT](https://learn.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/) implementation of [Toast Notifications](*https://learn.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/toast-notifications-overview)
```
MaikuB, line 62: There's an asterisk before the link https://learn.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/toast-notifications-overview. I believe is causing the link to not appear properly as shown on your fork's readme https://github.com/Levi-Lesches/flutter_local_notifications/tree/windows-ffi/flutter_local_notifications

MaikuB:
Putting aside how there's a way to provide XML, does this "implement" the entire that consumers can use the constructor to specify all the available options that the Windows SDK provides. Some examples that come to mind are how to set a hint crop or hero image like shown at https://learn.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/adaptive-interactive-toasts?tabs=appsdk

Levi-Lesches:
> Putting aside how there's a way to provide XML, does this "implement" the entire that consumers can use the constructor to specify all the available options that the Windows SDK provides.

Just to clarify, you're asking if there are Dart options for all the XML options? Yes! If something's missing I'd be happy to add it, but I went through the entire XML schema and hand-translated everything. The demo app, under Windows-specific examples, showcases almost all of these. Eg,

> how to set a hint crop

[`WindowsImage.crop`](https://github.com/Levi-Lesches/flutter_local_notifications/blob/68c03adce820d39b98ffd9791a4279459e744190/flutter_local_notifications_windows/lib/src/details/notification_image.dart#L45)

> or hero image

[`WindowsImagePlacement.hero`](https://github.com/Levi-Lesches/flutter_local_notifications/blob/68c03adce820d39b98ffd9791a4279459e744190/flutter_local_notifications_windows/lib/src/details/notification_image.dart#L42).

flutter_local_notifications_windows/lib/src/details/notification_details.dart
Levi-Lesches: I'll try to see if there's a better way to hide these on Monday

flutter_local_notifications_windows/lib/src/plugin/base.dart
Levi-Lesches: The original branch I worked with did have a `WindowsNotificationDetails.fromXml` constructor, however, I decided to remove it. The XML itself decides the title, body, and payload, and so I felt it would be off-putting to be able to say `show(title, body, payload, someOtherXmlThatOverridesEverything)`. 

That's why `notificationToXml` exists, because the XML is a combination of the title, body, payload, and all the other details -- even if none are provided. I figured it was fine to start off without the ability to schedule XML for now, but I can add that in too, it won't be much work.

Levi-Lesches:
> FYI I've noticed that when I debug the example app that the following can be seen in debug output

This is actually a common Flutter Desktop (at least, Windows) error I keep getting in all my apps, even fresh ones. I'm not sure exactly what causes it but it seems to appear more when I use the keyboard during app launch, hot reloads, or hot restarts, and that makes me think it's a debug mode problem. I wasn't able to find useful info on it online but it is an assertion which will get compiled out in release builds

flutter_local_notifications_windows/lib/src/plugin/base.dart
Levi-Lesches: I thought about that, but I figured it wouldn't make much sense to update a progress bar that was scheduled to be shown, because then you can't actually be sure you're updating anything at all. Besides, a progress bar is meant to represent real work done in real-time, not predicted to occur later. So it seemed to me that the only reasonable time to update a progress bar would be right after showing it,

flutter_local_notifications_windows/lib/src/plugin/ffi.dart
Levi-Lesches: > Didn't know it's possible to reference a package so thanks for that. 

Oh I don't think that does anything special in DartDocs, I just like `code style` in mardown. 

> However, can this be updated to The Windows implementation of [FlutterLocalNotificationsPlatform]

That would be `plugin/base.dart`

> Reason for pitching this is FFI sounds too generic. I believe the Linux implementation could technically be said to use FFI as it depends on dbus package that in turn makes use of FFI.

Right, but FFI crashes web builds. So I have `plugin/stub.dart` and `plugin/ffi.dart`, where the latter is the true, FFI-based implementation and the former is just a stub for web builds. `plugin/base.dart` is the base class that actually decides what the Windows plugin offers.

flutter_local_notifications/README.md
Levi-Lesches: Good catch! Not on my work laptop right now but I'll fix it tomorrow.

flutter_local_notifications_windows/lib/src/details/notification_to_xml.dart
Levi-Lesches: Fixed

flutter_local_notifications_windows/lib/src/plugin/ffi.dart
MaikuB: > That would be plugin/base.dart

I get what you mean but what's there is an abstract class and how it's used is more like an interface. It's not exposed publicly nor can it be constructed for it to be an implementation

MaikuB:
> Just to clarify, you're asking if there are Dart options for all the XML options? 

Yep that's what I mean. Thanks for clarifying as it looks like I missed so will need to check the XML options more thoroughly

flutter_local_notifications/example/lib/windows.dart
```diff
@@ -0,0 +1,391 @@
...
+    },
+  ),
+  PaddedElevatedButton(
+    buttonText: 'Show notifications with dynamic content',
+    onPressed: () async {
+      await _showWindowsNotificationWithDynamic();
+    },
+  ),
+  PaddedElevatedButton(
+    buttonText: 'Show notitification with activation',
```
MaikuB, line 66: notification has been spelt incorrectly here and in a few other places

flutter_local_notifications_windows/lib/src/details/notification_input.dart
```diff
@@ -0,0 +1,114 @@
...
+class WindowsTextInput extends WindowsInput {
+  /// Creates an input field in a notification.
+  const WindowsTextInput({
+    required super.id,
+    this.hintText,
+    super.title,
+  }) : super(type: WindowsInputType.text);
+
+  /// The hint text.
+  final String? hintText;
```
MaikuB, line 45: Whilst the name isn't wrong and is also what material uses, could this be renamed to `placeholderText` to better align with what the Windows API refers to it as? Presumably you picked this to avoid [placeHolderContent](https://learn.microsoft.com/en-us/uwp/schemas/tiles/toastschema/element-input) per the XML schema with it sounding too generic. If so, the WinRT APIs has it as [placeholderText](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.windows.appnotifications.builder.appnotificationbuilder.addtextbox?view=windows-app-sdk-1.5#microsoft-windows-appnotifications-builder-appnotificationbuilder-addtextbox(system-string-system-string-system-string))

flutter_local_notifications_windows/lib/src/details/notification_text.dart
```diff
@@ -0,0 +1,51 @@
...
+  /// The text being displayed.
+  final String text;
+
+  /// Whether to center this text. Only relevant if in an incoming call.
+  final bool centerIfCall;
+
+  /// Whether the text should be smaller like a caption.
+  final bool isCaption;
+
+  /// The placement of this text. Null indicates default.
```
MaikuB, line 33: Can you move second sentence to another paragraph?

flutter_local_notifications_windows/lib/src/details/notification_action.dart
```diff
@@ -0,0 +1,129 @@
+import "dart:io";
+
+import "package:xml/xml.dart";
+
+// NOTE: All enum values in this file have Windows RT-specific names.
+// If you change their Dart names, be sure to override [Enum.name].
+
+/// Decides how the [WindowsAction] will launch the app.
+///
+/// On desktop platforms, [foreground] and [background] are treated the same.
```
MaikuB, line 10: Apologies if I've missed something but is due to limitation on supporting background isolates or something else? Wanted to check as the sentence looked odd as I thought this would only run on a desktop environment to begin with

flutter_local_notifications_windows/pubspec.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+dependencies:
+  ffi: ^2.1.2
+  flutter_local_notifications_platform_interface: ^7.2.0
+  timezone: ^0.9.4
+  xml: ^6.5.0
+
+dev_dependencies:
+  ffigen: ^13.0.0
+  very_good_analysis: ^6.0.0
```
MaikuB, line 16: Could the use of this be removed for the moment? I recall you did raise a comment around linter rules used but for now would want to ensure there's consistency and can revisit the rules used later. If I remember correctly, if the Windows package has its `analysis_options.yaml` file removed, it should pull the one from the root of the repo. This may have required melos to bootstrapped

MaikuB: On a related note, I recall you mentioned another comment on Dart line length. I get where you're coming from and at work I work in a team that has chosen a larger value.  However, the default line length is 80. There's also a linter for it and is also being followed in a lot of repos to my knowledge e.g. the first-party packages. Looking to follow the defaults here for consistency

flutter_local_notifications/example/lib/windows.dart
Levi-Lesches: Whoops, fixed. FYI, apparently, my autocomplete suggested this because it found this spelling used in `ios/Classes/FlutterLocalNotifications.m`. I didn't change it in my commit, just thought you should know

flutter_local_notifications_windows/lib/src/details/notification_input.dart
Levi-Lesches: Done

flutter_local_notifications_windows/lib/src/details/notification_text.dart
Levi-Lesches: Done

flutter_local_notifications_windows/lib/src/details/notification_action.dart
Levi-Lesches: Taken from [here](https://learn.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=desktop-msix#handling-background-activation): 

> For desktop applications, background activations are handled the same as foreground activations (your OnActivated event handler will be triggered). You can choose to not show any UI and close your app after handling activation.

Whereas for UWP apps: 
> When you specify background activation on your toast (or on a button inside the toast), your background task will be executed instead of activating your foreground app.

So this was me being overly descriptive. But, since Flutter officially does not UWP, I'll remove the `background` option and comment.

flutter_local_notifications_windows/lib/src/plugin/ffi.dart
Levi-Lesches: Changed

flutter_local_notifications_windows/lib/src/plugin/base.dart
Levi-Lesches: I added `zonedScheduleRawXml` to the API, which is the same as `zonedSchedule` but XML replaces title, body, and payload. Resolving but feel free to re-open

flutter_local_notifications_windows/lib/src/details/notification_details.dart
Levi-Lesches: I put these in a private extension

flutter_local_notifications_windows/lib/src/plugin/base.dart
Levi-Lesches: Changed the comment

flutter_local_notifications_windows/pubspec.yaml
Levi-Lesches: Commit 9146d65 removes the custom lints and conforms to the 80 character line rule, as well as all the other lints. While I agree with conformity with the rest of the monorepo, I would strongly suggest reconsidering which lints the monorepo uses. Particularly, `cascade_invocations` forced a lot of imperative-style code like: 
```dart
builder.element(...);
builder.element(...);

plugin.show(...);
plugin.testDetails(...);
```

To be formatted as 
```dart
builder
  ..element(
    // ...
  )
  ..element(
    // ...
  );

plugin
  ..show(
    // ...
  )
  ..show( 
    // ...
  );
```

And of course, with a long name like `FlutterLocalNotifications` and prefixes/suffixes like `Windows` everywhere, the 80-character line length gets violated at almost every line. Even function signatures _before_ the parameters have to be split: 

```dart
  @override
  Future<List<PendingNotificationRequest>>
    pendingNotificationRequests() async =>
      using((Arena arena) {
```

This is made worse by `always_specify_types` requiring very long type names be used even when their constructors are in the same line. 

Again, I already committed the fixes for this PR, this is just a long-term recommendation advice for the rest of the repo. I do generally agree with following standards, these in particular aren't exactly agreed upon. While `always_specify_types` is part of [the Flutter style guide](https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md#avoid-using-var-and-dynamic), it [goes against Effective Dart](https://dart.dev/effective-dart/design#dont-redundantly-type-annotate-initialized-local-variables), and its opposite rule, [`omit_local_variable_types`](https://dart.dev/tools/linter-rules/omit_local_variable_types) is recommended instead. Character limits come from `dart format`, whose description explicitly recommends _not_ using it if your project needs configuration, and I'd say this package having a long name would justify that.

Levi-Lesches:
@MaikuB This should be ready to go now

flutter_local_notifications/example/lib/windows.dart
MaikuB: Thanks will take a look and address those outside of this PR :)

flutter_local_notifications_windows/pubspec.yaml
MaikuB: Yep noted, thanks

flutter_local_notifications_windows/lib/src/ffi/bindings.dart
```diff
@@ -0,0 +1,367 @@
...
+      _lookup<ffi.NativeFunction<ffi.Void Function(NativeLaunchDetails)>>(
+          'freeLaunchDetails');
+  late final _freeLaunchDetails =
+      _freeLaunchDetailsPtr.asFunction<void Function(NativeLaunchDetails)>();
+
+  /// EXPERIMENTAL: Enables multithreading for this application.
+  ///
+  /// NOTE: This is only to make tests more stable and is not intended to be used in applications.
+  void enableMultithreading() {
+    return _enableMultithreading();
```
MaikuB, line 277: Given the docs on this, I'm wondering if an assertion should also to also help act as another warning that developers. Some devs will not check API docs and would more likely run an app in debug mode that  if an assertion is done here, they can be given a more explicit warning/error

flutter_local_notifications_windows/lib/src/details/notification_action.dart
```diff
@@ -0,0 +1,134 @@
...
+  /// The style of the button. Null indicates a plain button.
+  final WindowsButtonStyle? buttonStyle;
+
+  /// The tooltip, useful if [content] is empty.
+  final String? tooltip;
+
+  /// Serializes this notification action as Windows-compatible XML.
+  ///
+  /// See: https://learn.microsoft.com/en-us/uwp/schemas/tiles/toastschema/element-action#syntax
+  void toXml(XmlBuilder builder) {
```
MaikuB, line 110: Can you move this out so it's not part of the public API surface?

flutter_local_notifications_windows/lib/src/details/notification_audio.dart
```diff
@@ -0,0 +1,164 @@
...
+  final bool shouldLoop;
+
+  /// Whether this notification should be silent.
+  final bool isSilent;
+
+  /// The source of the audio.
+  final String source;
+
+  /// Serializes this audio to Windows-compatible XML.
+  void toXml(XmlBuilder builder) => builder.element(
```
MaikuB, line 156: Can you move this out so it's not part of the public API surface?

flutter_local_notifications_windows/lib/src/details/notification_details.dart
```diff
@@ -0,0 +1,150 @@
...
+  /// Custom bindings in the notification.
+  ///
+  /// Text elements can contains "bindings", which are entered as
+  /// `{bindingName}` directly into the string values. You can then update them
+  /// while or after the notification is launched by using the binding name as
+  /// the key here, and the value as any string you want.
+  final Map<String, String> bindings;
+
+  /// Builds all relevant XML parts under the root `<toast>` element.
+  void toXml(XmlBuilder builder) {
```
MaikuB, line 107: Can you move this out so it's not part of the public API surface?

flutter_local_notifications_windows/lib/src/details/notification_details.dart
```diff
@@ -0,0 +1,150 @@
...
+      },
+    );
+    audio?.toXml(builder);
+    header?.toXml(builder);
+  }
+
+  /// Generates the `<binding>` element of the notification.
+  ///
+  /// See: https://learn.microsoft.com/en-us/uwp/schemas/tiles/toastschema/element-binding
+  void generateBinding(XmlBuilder builder) {
```
MaikuB, line 136: Can you move this out so it's not part of the public API surface?

flutter_local_notifications_windows/lib/src/details/notification_header.dart
```diff
@@ -0,0 +1,44 @@
...
+  final String title;
+
+  /// An application-defined payload that will be passed back when pressed.
+  final String arguments;
+
+  /// Specifies how the application will open.
+  final WindowsHeaderActivation? activation;
+
+  /// Serializes this header to XML.
+  void toXml(XmlBuilder builder) => builder.element(
```
MaikuB, line 35: Can you move this out so it's not part of the public API surface?

flutter_local_notifications_windows/lib/src/details/notification_image.dart
```diff
@@ -0,0 +1,68 @@
...
+  final File file;
+
+  /// Where this image will be placed. Null indicates below the notification.
+  final WindowsImagePlacement? placement;
+
+  /// How the image will be cropped. Null indicates uncropped.
+  final WindowsImageCrop? crop;
+
+  @override
+  void toXml(XmlBuilder builder) {
```
MaikuB, line 49: Can you move this out so it's not part of the public API surface?

flutter_local_notifications_windows/lib/src/details/notification_input.dart
```diff
@@ -0,0 +1,116 @@
...
+    required super.id,
+    this.placeHolderContent,
+    super.title,
+  }) : super(type: WindowsInputType.text);
+
+  /// A placeholder shown before the user enters input, like a hint text.
+  final String? placeHolderContent;
+
+  @override
+  void toXml(XmlBuilder builder) => builder.element(
```
MaikuB, line 49: Can you move this out so it's not part of the public API surface?

flutter_local_notifications_windows/lib/src/details/notification_input.dart
```diff
@@ -0,0 +1,116 @@
...
+  }) : super(type: WindowsInputType.selection);
+
+  /// The items that can be selected.
+  final List<WindowsSelection> items;
+
+  /// The default item that is selected.
+  final String? defaultItem;
+
+  @override
+  void toXml(XmlBuilder builder) => builder.element(
```
MaikuB, line 78: Can you move this out so it's not part of the public API surface?

flutter_local_notifications_windows/lib/src/details/notification_input.dart
```diff
@@ -0,0 +1,116 @@
...
+  });
+
+  /// A unique ID for this item.
+  final String id;
+
+  /// The content of this item in the UI.
+  final String content;
+
+  /// Serializes this item to XML.
+  void toXml(XmlBuilder builder) => builder.element(
```
MaikuB, line 109: Can you move this out so it's not part of the public API surface?

flutter_local_notifications_windows/lib/src/details/notification_input.dart
```diff
@@ -0,0 +1,116 @@
...
+  final String id;
+
+  /// The type of this input.
+  final WindowsInputType type;
+
+  /// The title of this input.
+  final String? title;
+
+  /// Serializes this input to XML.
+  void toXml(XmlBuilder builder);
```
MaikuB, line 33: Can you move this out so it's not part of the public API surface? Could you also see if this could be renamed? It doesn't return a result and methods named `to<...>` conventionally return a result due to how they're named

flutter_local_notifications_windows/lib/src/details/notification_part.dart
```diff
@@ -0,0 +1,17 @@
...
+// This class needs to be abstract so [WindowsNotificationText] and
+// [WindowsImage] can extend it. Specifically, this class is a marker
+// type for classes that are valid as part of a [WindowsColumn].
+// ignore: one_member_abstracts
+abstract class WindowsNotificationPart {
+  /// A const constructor.
+  const WindowsNotificationPart();
+
+  /// Serializes this part to XML, according to the Windows API.
+  void toXml(XmlBuilder builder);
```
MaikuB, line 16: Can you move this out so it's not part of the public API surface? Could you also see if this could be renamed? It doesn't return a result and methods named `to<...>` conventionally return a result due to how they're named

flutter_local_notifications_windows/lib/src/details/notification_progress.dart
```diff
@@ -0,0 +1,48 @@
...
+  /// Setting this to null indicates a indeterminate progress bar.
+  double? value;
+
+  /// Overrides the default reading as a percent with a different text.
+  ///
+  /// Useful for indicating discrete progress, like `3/10` instead of `30%`.
+  String? label;
+
+  /// Serializes this progress bar to XML.
+  void toXml(XmlBuilder builder) => builder.element(
```
MaikuB, line 39: Can you move this out so it's not part of the public API surface? Could you also see if this could be renamed? It doesn't return a result and methods named `to<...>` conventionally return a result due to how they're named

flutter_local_notifications_windows/lib/src/details/notification_row.dart
```diff
@@ -0,0 +1,41 @@
...
+/// See: https://learn.microsoft.com/en-us/uwp/schemas/tiles/toastschema/element-group
+class WindowsRow {
+  /// Makes a group of multiple columns.
+  const WindowsRow(this.columns);
+
+  /// The different columns being grouped together.
+  final List<WindowsColumn> columns;
+
+  /// Serializes this group to XML.
+  void toXml(XmlBuilder builder) => builder.element(
```
MaikuB, line 16: Can you move this out so it's not part of the public API surface? Could you also see if this could be renamed? It doesn't return a result and methods named `to<...>` conventionally return a result due to how they're named

flutter_local_notifications_windows/lib/src/details/notification_text.dart
```diff
@@ -0,0 +1,53 @@
...
+  /// The placement of this text.
+  ///
+  /// The default placement (null) is in the main body of the notification.
+  final WindowsTextPlacement? placement;
+
+  /// The language of this text.
+  final String? languageCode;
+
+  @override
+  void toXml(XmlBuilder builder) => builder.element(
```
MaikuB, line 42: Can you move this out so it's not part of the public API surface? Could you also see if this could be renamed? It doesn't return a result and methods named `to<...>` conventionally return a result due to how they're named

MaikuB:
Thanks for making the changes. I left one comment. Other thing was I wasn't too sure if you were done addressing the visibility of methods to with XML conversion. Since you mentioned it's good to go then FYI that there are still some methods that are publicly visible. I've commented on the parts that I could see are still visible to be more explicit. Apologies for spam on that but though it help make them easier to find.  FYI that the way to verify is to follow the steps [here](https://docs.flutter.dev/packages-and-plugins/developing-packages#api-documentation) to generate the API docs for the Windows plugin. 

Just noticed that there's also errors on the dependency/version conflict that wasn't there before. Presumably ffigen got bumped that it's no longer compatible with what was stated before on Dart 3.1 being the minimum

flutter_local_notifications_platform_interface/pubspec.yaml
```diff
@@ -1,15 +1,13 @@
...
-version: 7.2.0
+version: 7.3.0
 homepage: https://github.com/MaikuB/flutter_local_notifications/tree/master/flutter_local_notifications_platform_interface
 
 environment:
   sdk: ^3.1.0
   flutter: ">=3.13.0"
 
 dependencies:
-  flutter:
```
MaikuB, line 11: Just noticed this but is there a reason this got removed? Could it be restored? Whilst the `plugin_platform_interface` itself doesn't depend on Flutter, I can see the platform interfaces for other plugins have this included. The [PR](https://github.com/MaikuB/flutter_local_notifications/pull/2393/files) I have includes a fix to have API docs reference `PlatformException` and this required importing `'package:flutter/services.dart'`. Could be it still functions without it but IMO looks off to not include given the import and how this is for Flutter

flutter_local_notifications_windows/lib/src/ffi/bindings.dart
Levi-Lesches: This file is auto-generated and is not part of the public API. The public section is `lib/src/plugin/base.dart`, which is marked with `@visibleForTesting`. That will give a warning in the IDE/`dart analyze` if someone tries to use this.

flutter_local_notifications_platform_interface/pubspec.yaml
Levi-Lesches: Turns out, because I used FFI and not Flutter's Method Channels, the Windows version of this plugin is fully available for non-Flutter projects, and can be used in CLI apps (eg, start processing a file, send a notification when done). I think this is a valuable use-case and there are no plugins available on Pub to serve this need (and if there were one, it would ideally use FFI in the same way I did here). 

To allow this, I had to remove `flutter` as a dependency on the `_windows` plugin, and that means transitively removing it here as well. It doesn't impact functionality at all, but would instead allow even more developers to use this package. And the same could be done for Linux, since it is also FFI-based. If the MacOS version is re-implemented in FFI in the future it would benefit from this as well.

https://github.com/MaikuB/flutter_local_notifications/blob/d565daab613052fecaf723df865101b46ee5a26a/flutter_local_notifications_platform_interface/lib/flutter_local_notifications_platform_interface.dart#L87-L93

Reading these comments, it seems that it's basically the same as an `UnsupportedError`, which is unfortunately an `Error`, and not an `Exception`. `PlatformException` does seem like the right tool for the job, but I really don't think including the DartDoc link is worth totally breaking support for CLI apps on Windows and Linux. Maybe just link it the regular way?
```dart
/// [`PlatformException`](https://api.flutter.dev/flutter/services/PlatformException-class.html)
```

Levi-Lesches:
> when I checked the API docs for the Windows plugin and Linux one, I've noticed an existing problem where the docs for the stub is what is shown. Not a showstopper for the PR right now as it's an existing problem already that I've not seen anyone raise with the Linux implementation. Not sure how to solve this one but open to ideas.

When a subclass overrides a member, that member inherits its docs from the overriden version in the parent class, so long as docs are not explicitly provided on the override itself. This seems to be why the Linux version shows the stub docs -- the stub's overrides all have doc comments. To fix this, we should just not provide documentation on the stub class and let in inherit directly from the base class. 

That just leaves the documentation on the class itself. I think the simplest fix would then be to copy the base class's documentation onto the stub class, which is a lot less work than copying every single member. So now: 
- all three classes (base, stub, and FFI) all have the same one-liner on the class declaration itself
- any time a method is overriden, docs are _not_ provided, and are instead inherited from the parent class
- the plugin class shown on the docs site is the stub class, with comments from the base class
- there is no way to get DartDoc to show the FFI code snippets when using conditional imports, only the stub snippets

Levi-Lesches:
> Just noticed that there's also errors on the dependency/version conflict that wasn't there before. Presumably ffigen got bumped that it's no longer compatible with what was stated before on Dart 3.1 being the minimum

Yup, from `package:ffigen` v10.0.0 (currently at 13.0.0):

> Stable release targeting Dart 3.2 using new dart:ffi features available in Dart 3.2 and later.

Those specific features aren't used by us in this exact case, but the features from ffigen 10.0 -- 13.0 (soon to be 14.0) will be, especially exhaustive enum checking. I think it would be worth bumping the minimum SDK version to `3.2`. Now that we've passed `3.0`, I'm not sure I see much risk in bumping the minor versions, and most other packages tend to stay current with the latest Dart releases

Levi-Lesches:
> Thanks for making the changes. I left one comment. Other thing was I wasn't too sure if you were done addressing the visibility of methods to with XML conversion. Since you mentioned it's good to go then FYI that there are still some methods that are publicly visible. I've commented on the parts that I could see are still visible to be more explicit. Apologies for spam on that but though it help make them easier to find. FYI that the way to verify is to follow the steps [here](https://docs.flutter.dev/packages-and-plugins/developing-packages#api-documentation) to generate the API docs for the Windows plugin.

I actually figured it would probably be more convenient to leave them public. Since we offer a way to show notifications as raw XML, this would allow developers to build that XML using elements from the public API. Let's say they want to use their own `<image>`, they can still build the rest of the notification XML using `.toXml()` of everything else. It's not so important to me but I figured it would only reduce duplicate work and make custom XML easier. I'm not worried about stability or breaking changes because these methods are anyway compliant with the official Windows APIs which aren't breaking anytime soon. 

If we really wanted to keep everything private, I could move them all to extensions, and then `hide` them in the exports section. But I feel like that's going to extra lengths to prevent possibly useful code from being used or seen. Not to mention, having the code snippets available in the docs site would also be useful to those looking to build their own XML.

Regarding the naming convention, I switched everything to `.buildXml()`

flutter_local_notifications_platform_interface/pubspec.yaml
MaikuB: Gotcha and makes. I forgot the markdown approach for links could be used for links as well. Thanks

MaikuB:
> When a subclass overrides a member, that member inherits its docs from the overriden version in the parent class, so long as docs are not explicitly provided on the override itself.

Good callout as I admittedly forgot about that. Works for me and will update the approach used for Linux as well. Thanks!

> Those specific features aren't used by us in this exact case, but the features from ffigen 10.0 -- 13.0 (soon to be 14.0) will be, especially exhaustive enum checking. I think it would be worth bumping the minimum SDK version to 3.2. Now that we've passed 3.0, I'm not sure I see much risk in bumping the minor versions, and most other packages tend to stay current with the latest Dart releases. I bumped it in my latest commit so it should work now

Thanks though building for 3.13 will fail as the main plugin itself has a minimum of 3.13 so that and pipeline would need updating as part of the PR too. I've withheld bumping for this plugin for a while and would still need to be careful of it in the future as it could affect the minimum supported OS versions

> If we really wanted to keep everything private, I could move them all to extensions, and then hide them in the exports section

Yep would like to keep these hidden. These methods depend on an existing object passed through so AFAICT there's some prerequisites on what needs to be done before calling these `build` methods. When I was referring to breaking changes, it was more around how if these methods are part of the public API surface, updates to them (e.g. renaming) are considered breaking changes. The hiding is to help reduce the burden on a maintenance perspective. As for your point on `hide`, if all the extensions were moved to `notification_to_xml.dart` then they will be hidden without extra code. This is because the approach to determine what is publicly visible is done by determining which implementation files in the `src` folder are `export`ed

Levi-Lesches:
> Thanks though building for 3.13 will fail as the main plugin itself has a minimum of 3.13 so that and pipeline would need updating as part of the PR too. I've withheld bumping for this plugin for a while and would still need to be careful of it in the future as it could affect the minimum supported OS versions

Bumped `pubpsec.yaml` and GitHub workflow environments to use Dart 3.2 and Flutter 3.16 (which comes with Dart 3.2)

> Yep would like to keep these hidden.

Got it, moved all that to extensions in `lib/src/details/xml` that are not exposed publicly. It caused an issue in two cases where I use subclasses, since making `buildXml` a member of the parent class would make it public. I tried using `@internal` from `package:meta`, but then I decided to seal their abstract classes and use a switch case over the subtypes. 

----

On another note, I restricted the tests to run on Windows only, but the current workflows are running on `ubuntu-latest`. I'll leave that to you to handle since I think Windows CI is billed at a 2x rate compared to the Linux CI. But at least now, `dart test` on `ubuntu-latest` will skip the Windows tests automatically, and we can run the tests locally for now. I'm actually not even sure how notification privileges will work on GitHub servers. I don't see a specific issue, but it sounds like one of those things that could work differently compared to, say, my laptop.

abdelaziz-mahdy:
> > Thanks though building for 3.13 will fail as the main plugin itself has a minimum of 3.13 so that and pipeline would need updating as part of the PR too. I've withheld bumping for this plugin for a while and would still need to be careful of it in the future as it could affect the minimum supported OS versions
> 
> Bumped `pubpsec.yaml` and GitHub workflow environments to use Dart 3.2 and Flutter 3.16 (which comes with Dart 3.2)
> 
> > Yep would like to keep these hidden.
> 
> Got it, moved all that to extensions in `lib/src/details/xml` that are not exposed publicly. It caused an issue in two cases where I use subclasses, since making `buildXml` a member of the parent class would make it public. I tried using `@internal` from `package:meta`, but then I decided to seal their abstract classes and use a switch case over the subtypes.
> 
> On another note, I restricted the tests to run on Windows only, but the current workflows are running on `ubuntu-latest`. I'll leave that to you to handle since I think Windows CI is billed at a 2x rate compared to the Linux CI. But at least now, `dart test` on `ubuntu-latest` will skip the Windows tests automatically, and we can run the tests locally for now. I'm actually not even sure how notification privileges will work on GitHub servers. I don't see a specific issue, but it sounds like one of those things that could work differently compared to, say, my laptop.

i would like to note that public repos actions are free, so i think the rate is not a problem? https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions

MaikuB:
@Levi-Lesches sorry for late response, was recovering from surgery. It looks like there's still a conflict. Based on my understanding of the constraints set by ffigen 13.0, this actually needs Flutter 3.19/Dart 3.3 (reference: table under stable at https://docs.flutter.dev/release/archive). Are you able to help resolve this

>  I'm actually not even sure how notification privileges will work on GitHub servers.

I could be mistaken but presumably as they're unit tests, the priviledges to be concerned about. Given what @abdelaziz-mahdy has said, you think you could push an update to have the tests run on Windows? On a related note, is it possible to make updates so that so that the example app is built on Windows as part of the  GitHub workflow too? Would require updates to melos.yaml and validate.yml

I've also noticed there's a lot of files with changes picked up if I run `dart format .` from the root of the repo. Are you able to run that (or `melos run format`) and push the changes? I'll look into how to make this smoother separately in another PR to help others in the future (e.g. see if the workflow can help catch this)

Levi-Lesches:
> was recovering from surgery

Sorry to hear, hope that's going well 🙂

> Based on my understanding of the constraints set by ffigen 13.0, this actually needs Flutter 3.19/Dart 3.3 (reference: table under stable at https://docs.flutter.dev/release/archive). Are you able to help resolve this

That's annoying, I checked the changelog and found mention of `3.2` but not `3.3`! I changed everything to Dart 3.3 and Flutter 3.19 now.

> I could be mistaken but presumably as they're unit tests, the priviledges to be concerned about.

I didn't mock _everything_ out, as I was interested in maintaining tests that the XML does in fact conform to the Windows API. So the unit tests actually do send full XML to the Windows API and try to send notifications. If this doesn't work on the server, I'd still like to keep them, but maybe we'll just run them manually when changing XML code and configure the CI to only run "safe" tests. Speaking of, I added a global `retry: 5` to all the tests, given that there are some concurrency issues.

> you think you could push an update to have the tests run on Windows? On a related note, is it possible to make updates so that so that the example app is built on Windows as part of the GitHub workflow too? Would require updates to melos.yaml and validate.yml

Hit an annoying issue where "skip these tests on Windows" was actually causing Linux tests to _fail_. Filed https://github.com/dart-lang/test/issues/2277 to propose to change that, but in the meantime I exclude `_windows` in the standard unit tests and run it separately as `test:unit:windows`, only on `windows-latest`. Needs another workflow approval.

> I've also noticed there's a lot of files with changes picked up if I run dart format . from the root of the repo. Are you able to run that (or melos run format) and push the changes?

Done and pushed

----

On another note, I had previously tried to make the Dart package completely separate from Flutter so it can be used in CLI apps. I still believe there is value to that, but when running `dart pub lish -n` we get an error that is never an issue in practice: 
```console
Package validation found the following error:
* pubspec.yaml allows Flutter SDK version 1.9.x, which does not support the flutter.plugin.platforms key.
  Please consider increasing the Flutter SDK requirement to ^1.10.0 (environment.sdk.flutter)

  See https://flutter.dev/docs/development/packages-and-plugins/developing-packages#plugin
```
Basically, even though having a `flutter:` section in the Pubspec of a non-Flutter package is harmless, _theoretically_, someone can use a very early version of Flutter (1.9.x) which would have issues parsing this section. Really, this can never happen because we demand an `sdk: ^3.3` and Flutter 1.9 is pre null-safety.  So Pub wants us to demand `flutter: ^1.10.0`, but that means declaring a dependency on Flutter at all, which I was _really_ hoping to avoid. See also [this issue](https://github.com/dart-lang/pub/issues/3563).

My recommendation is to run `dart pub lish -n` first, see if there are any warnings besides the above, and address them first. Then, once this is the only warning left, running `dart pub lish --skip-validation` to ignore the warning, and preserve CLI ability.

MaikuB:
I'll take a look and come back to you on these but thought I'd reply in case you missed it since it wasn't brought up in your last message and mention that builds are failing due to the meta version specified

Levi-Lesches:
Good eye -- fixed

MaikuB:
> Sorry to hear, hope that's going well 🙂

Thanks minor surgery and still bit of discomfort but all good besides that :)

> On another note, I had previously tried to make the Dart package completely separate from Flutter so it can be used in CLI apps.

I forgot to ask but is this realistically possible? In the case of the platform interface, this would normally be a transitive dependency. In the case of the Windows plugin, I don't know if you could build a CLI app that shows notifications on the Windows platform? Presumably what you mean is the equivalent of using Visual Studio to create a console app and referencing the Windows APIs to display notifications

> My recommendation is to run dart pub lish -n first, see if there are any warnings besides the above, and address them first

This is what I get besides the Flutter one

```
    error - lib/src/plugin/ffi.dart:122:7 - The named parameter 'data' isn't defined. Try correcting the name to an existing named parameter's name, or defining a named parameter with the name 'data'. - undefined_named_parameter
    error - lib/src/plugin/ffi.dart:200:9 - The named parameter 'data' isn't defined. Try correcting the name to an existing named parameter's name, or defining a named parameter with the name 'data'. - undefined_named_parameter
```

I've taken a look at the issue you linked and conscious it's been open for a while. I do have a number of thoughts relating to this to and may be missing some additional context and other information so do let me know if I'm missing anything.  If issue linked addressed and there are valid cases for CLI apps, I wonder if it'd be better for the "greater good" to go ahead with the Flutter requirement and revisit this. Those who use this plugin would be those looking to build a multiplatform apps. Other aspect of this is if the issue you raised is addressed, then this presumably requires newer version of the Dart SDK. If so, this would cause the Flutter SDK requirement to go even higher and limit the reach even further.

If it's possible to build a Dart CLI app that can show notifications on Windows then this sounds like it one target a very small audience if such an audience even exists. I'm basing this on the assumption that this implies that this is a developer who works on the Windows platform and rather than building a Windows console application, has decided to use Dart. Is this what it would mean? If so, this seems like an odd choice to me. In my experience, companies that build dedicated solutions targeting Windows would employ devs and leverage based on .NET

On a different note, it looks like Windows example is failing to built I'm not sure what's the cause there. However, I'm noticing that the integration tests are failing to run for Android. It seems as though the deprecation warning is being treated as a error but can't reproduce this locally even if I run the integration tests via melos. Given I had raised a PR recently to update one of the workflows and didn't have this issue, I wonder if something got changed that impacted this?

MaikuB:
It's been a while since I've worked on Windows apps so didn't get to spend much time on it but ran into the following problems relating my previous post

- I had trouble getting a console app to create a notification. I was hoping to copy some of the code for the Windows App SDK samples for showing notifications to achieve this but trouble with having appropriate namespaces and NuGet packages resolve properly
- I tried to create Dart CLI app by creating a separate folder in the cloned repo, reference only the Windows plugin, bootstrap melos etc. Upon trying to run the CLI app, I saw `Error: Dart library 'dart:ui' is not available on this platform.` as an error message. It looks like this happens as the dependencies the plugin eventually need it and that's only available for apps that depends on Flutter

Based on the above it looks like trying to support CLI apps isn't actually possible, at least not in the current state either

flutter_local_notifications_windows/src/plugin.cpp
```diff
@@ -0,0 +1,227 @@
...
+  if (!IsWindows8OrGreater()) return false;
+  uint32_t length = 0;
+  auto error = GetCurrentPackageFullName(&length, nullptr);
+  if (error == APPMODEL_ERROR_NO_PACKAGE) return false;
+  else if (error != ERROR_INSUFFICIENT_BUFFER) return std::nullopt;
+  PWSTR fullName = (PWSTR) malloc(length * sizeof(*fullName));
+  if (fullName == nullptr) return std::nullopt;
+  error = GetCurrentPackageFullName(&length, fullName);
+  if (error != ERROR_SUCCESS) return std::nullopt;
+  free(fullName);
```
autoantwort, line 225: Use a std::vector. You currently have a memory leak in the case of an error.

gryznar:
@Levi-Lesches thank you for your work on this! I highly appreciate Windows support :)

MaikuB:
@Levi-Lesches no rush and could be that you've been busy. Wanted to ping in case you missed some comments/questions I left. Looks like @autoantwort left a comment on the PR as well

flutter_local_notifications_windows/src/plugin.cpp
Levi-Lesches: Addressed - thanks! (I copied almost all the code in this file from previous branches and forks, so the review is appreciated)

Levi-Lesches:
@MaikuB Thanks for the ping, I was kinda going through it with some personal matters.

- I fixed a bug that was causing Windows builds to fail before a certain Windows 11 version (ie, Windows 10)
- Merged `master` and resolved all merge conflicts
- Ran all tests on my end, including Android, so hopefully the workflows should stay green now
- I reverted the whole Flutter/pure Dart thing. I'm a bit of a purist but that unresolved Pub issue is really annoying

Should be ready to merge now! Note that it depends on `_platform_interface: 8.1.0`, not `8.0.0`, because my branch has changes that are not yet published, so both will need to be published to Pub

MaikuB:
Thanks @Levi-Lesches. I'll be trying to get the current prerelease out first and then come back to this one. In the interest of time. I may push directly to this fork if there are merge conflicts to resolve afterwards due to this

MaikuB:
@Levi-Lesches looks like the Windows build tasks failed. You able to see what the issue is?

Levi-Lesches:
The error is definitely new to me and doesn't occur when I run on my machine. From Googling around, the error `FileTracker : error FTK1011: could not create the new file tracking log file` seems to be linked to the Windows file length limit...

MaikuB:
@Levi-Lesches the approval is need for first-time contributors and in your case, it's due to the PR not having being landed. I won't get to make these changes for a while as I've only time to try to land https://github.com/MaikuB/flutter_local_notifications/pull/2442. If you have time, you may be able to test your proposal by temporarily changing the workflow (e.g. run on push) and see how it fixes the issue. A shame to see the file/path length limit is still a PITA for Windows :(

Levi-Lesches:
While it's true that the base path is `flutter_local_notifications\flutter_local_notifications` and cannot be changed, I'm experimenting with changing `flutter_local_notifications\example` into `f\e`, which should save some vital characters

Levi-Lesches:
@MaikuB So close! Not sure if you missed the CI turning green or want to wait for next week on this, so I'll just drop a reminder. Not sure if you noticed the chat going wonky during the shift to DST too or if that was just on my end.

MaikuB:
When I checked after your message, it was still red but the results may have been cached. I still need to retest before merging just in case. 

Regarding the timing issue on GitHub, yeah I noticed it but thought it was just me as I've seen a similar issue on Slack that was to do with Chromium browsers but sounds like it was a general problem

Levi-Lesches:
Thanks, lmk if you need anything else. On the bright side... I guess we know now that the scheduling features on Windows don't throw crazy errors during DST transitions!

flutter_local_notifications_windows/src/plugin.cpp
```diff
@@ -0,0 +1,225 @@
...
+};
+
+using RegistryKey = winrt::handle_type<RegistryHandle>;
+
+/// This callback will be called when a notification sent by this plugin is clicked on.
+struct NotificationActivationCallback : winrt::implements<NotificationActivationCallback, INotificationActivationCallback> {
+	NativeNotificationCallback callback;
+
+	HRESULT __stdcall Activate(LPCWSTR app, LPCWSTR args, NOTIFICATION_USER_INPUT_DATA const* data, ULONG count) noexcept final {
+		try {
```
MaikuB, line 29: A very minor thing I noticed just now. Looks like the formatting here is off? I don't work with C++ but from what I read [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html) can be applied. If so, would it be possible to include in this PR an update to [this workflow](https://github.com/MaikuB/flutter_local_notifications/blob/master/.github/workflows/format.yml) so that it could format C++ plugin introduced by the Windows implementation too?

Levi-Lesches: I set up clang-format with a custom `.clang-format` file, designed to make minimal changes while enforcing consistency.

Levi-Lesches:
@MaikuB I merged and formatted, want to give it another go?

flutter_local_notifications_windows/src/plugin.cpp
Levi-Lesches: Note: I used a pip-based installation of clang (19) for now since `ubuntu-latest` is `22.04` which can't download `clang-format-17` (jammy vs noble). I only now saw that you can manually select `ubuntu-24.0` in the workflow, which might work, but I don't have time to test that right now.

MaikuB: Thanks have a comment but I'll leave it to where it's more appropriate and leave this one as resolved

Levi-Lesches:
Looks like the Android tests are indeed passing, but the new version of Flutter (3.24) requires a new bootstrapping mechanism: 

```console
Running Gradle task 'assembleDebug'...                          
ERROR: ERROR: You are applying Flutter's app_plugin_loader Gradle plugin imperatively using the apply script method, which is deprecated and will be removed in a future release. Migrate to applying Gradle plugins with the declarative plugins block: https://flutter.dev/to/flutter-gradle-plugin-apply


ERROR: ERROR: 


ERROR: ERROR: You are applying Flutter's main Gradle plugin imperatively using the apply script method, which is deprecated and will be removed in a future release. Migrate to applying Gradle plugins with the declarative plugins block: https://flutter.dev/to/flutter-gradle-plugin-apply


ERROR: ERROR: 
```

These have always [appeared in the logs](https://github.com/MaikuB/flutter_local_notifications/actions/runs/11650031526/job/32438622393) but only in the new version are actually failing the tests. I can help with those in a separate PR if you'd like, but I think this one is ready to merge despite that.

.github/workflows/format.yml
```diff
@@ -34,3 +34,19 @@ jobs:
           which swiftlint || brew install swiftlint
           swiftlint --fix
           git diff --exit-code || (git commit --all -m "Swift Format" && git push)
+
+  windows_cpp_format:
```
MaikuB, line 38: Is it possible for this step to commit changes back like the `objc_format` job?

MaikuB:
> These have always appeared in the logs but only in the new version are actually failing the tests. I can help with those in a separate PR if you'd like, but I think this one is ready to merge despite that.

Yeah I left this as plugin needed a min version bump to migrate the example app. Once your PR is merged in then should be fine to migrate and yes happy to take a PR on that :)

.github/workflows/format.yml
Levi-Lesches: Done.

Levi-Lesches: @MaikuB  If there are issues running the workflows again (I can't see any on my side due to permissions), it may be because [commits authored by GitHub Actions don't trigger workflows](https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#commits-made-by-this-action-do-not-trigger-new-workflow-runs). If there are no issues. I guess it's good to go.

MaikuB:
Should also mention my plan right now is to release it after the example app issue is fixed and the release would be done as part of prerelease

Levi-Lesches:
Sounds good. If the builds are still green then I might push it off on my to-do list for now, so if you have more time on your hands then maybe don't wait for me.

MaikuB:
All good. I'll be on holidays for a while soon as anyway :)

flutter_local_notifications_windows/flutter_local_notifications_windows.dll
MaikuB, line 1: Sorry just noticed this as I was preparing a release. Did this DDL by chance accidentally make it when you ran the `build.bat` file or have I missed something?

MaikuB: Thanks you were right. I tested it by deleting the DLL, ran the unit tests locally and it failed. I was able to run the example app without it though. If it's only needed for the tests then is it possible to have the build generated as one of the steps before the tests are run? This would avoid having to manually remove the DLL before publishing to pub.dev. This in turn prevents the DLL from being downloaded by those using the plugin

Levi-Lesches: See #2458

MaikuB: Thanks will marked this as resolved and continue discussions there

flutter_local_notifications_windows/lib/src/details/xml/action.dart
```diff
@@ -0,0 +1,34 @@
...
+import '../notification_action.dart';
+
+/// Converts a [WindowsAction] to XML
+extension ActionToXml on WindowsAction {
+  /// Serializes this notification action as Windows-compatible XML.
+  ///
+  /// See: https://learn.microsoft.com/en-us/uwp/schemas/tiles/toastschema/element-action#syntax
+  void buildXml(XmlBuilder builder) {
+    if (image != null && !image!.isAbsolute) {
+      throw ArgumentError.value(
```
MaikuB, line 11: @Levi-Lesches  Just noticed this now whilst checking https://github.com/MaikuB/flutter_local_notifications/pull/2511 and noticed this. Could you shift this logic that checks this and throws an `ArgumentError` to the constructor itself? This keeps the logic co-located. Moreover, an `ArgumentError` here indicates an issue with what's passed to `buildXml()` but it's actually to do with what got passed to the constructor

MaikuB: Forgot to add I commented here as there wasn't a change around this file in #2511. Happy for #2511 to include this or be in a separate PR

Levi-Lesches: Good spot! On it now

lukemmtt:
@Levi-Lesches Just wanted to say thank you for your work on windows support!

PR #2371, opened 2024-07-14
[flutter_local_notification_linux] fixed definition and registration of Linux implementation

### LINKED ISSUE #2368: I was told to ask you to either avoid a reference or create a plugin (I am serious :)
I started getting this error/warning today, out of the blue.  Maybe related to upgrade of flutter master channel SDK, as I can't remember changing anything else.

App will still launch fine

```
Package flutter_local_notifications:linux references flutter_local_notifications_linux:linux as the default plugin, but the package does not exist.
Ask the maintainers of flutter_local_notifications to either avoid referencing a default implementation via `platforms: linux: default_package: flutter_local_notifications_linux` or create a plugin named flutter_local_notifications_linux.
```

MaikuB:
Closes #2368

### DIFF
--- a/flutter_local_notifications_linux/CHANGELOG.md
+++ b/flutter_local_notifications_linux/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [4.0.1]
+
+* Fixed pubspec so it defines that it implements the Linux implementation of `flutter_local_notifications`
+
 ## [4.0.0+1]
 
 * Bumped maximum Dart SDK constraint
--- a/flutter_local_notifications_linux/CHANGELOG.md
+++ b/flutter_local_notifications_linux/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## [4.0.1]
 
-* Fixed pubspec so it defines that it implements the Linux implementation of `flutter_local_notifications`
+* Fixed issue [#2368](https://github.com/MaikuB/flutter_local_notifications/issues/2368). This involved updating pubspec so it defines that it implements the Linux implementation of `flutter_local_notifications` and updating the code so it registers the Linux implementation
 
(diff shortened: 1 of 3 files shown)


Gustl22:
@MaikuB just chiming in and was curious, as it seems removing [`default_package`](https://github.com/MaikuB/flutter_local_notifications/blob/49c487f351185c8067d0378ea896cc0fc24b4fa9/flutter_local_notifications/pubspec.yaml#L36) in pubspec.yaml of `flutter_local_notifications` seems to be the more proper fix for this, as it doesn't use any native platform code which needs registering.

MaikuB:
@Gustl22 agreed though I would more than it doesn't have to be handled by main package than should. This is more because how the plugin is created where from a consumer's point of view, they're constructing an instance of the plugin so it would've been more natural to expect that the Linux implementation is registered then IMO.  I hadn't made the changes on the main plugin as it was less urgent that it could be done later (note: I also need to test it out) and AFAIK doesn't cause an issue. The changes I did here were to conform with the expectations that were enforced by the recent changes in Flutter. I am aware though that federated plugin spec documents certain things to be done.

Something this issue does bring to mind though is the message behind the warning should re-evaluated. I saw that you submitted the PR to the Flutter SDK around this. Perhaps something you could look into? I'm not sure how much of an impact this may have on the community on the future

Gustl22:
My introduced error message just should make aware, that the package uses a keyword `default_package`, which has no effect and should be used otherwise and bring up the options. But yes, the error message is kind of misleading, as there can exist packages, but they just don't serve as a native or Dart plugin.
So I added this as proposed by stuardmorgan: https://github.com/flutter/website/pull/10891

PR #2379, opened 2024-07-23
[flutter_local_notifications] updated plugin to no longer register Linux implementation

MaikuB:
Follow up to #2371 and relates to #2368 

As the Linux package now handles registering the Linux implementation, the app-facing package no longer needs to do this

### DIFF
--- a/flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
+++ b/flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
@@ -38,9 +38,6 @@ class FlutterLocalNotificationsPlugin {
     } else if (defaultTargetPlatform == TargetPlatform.macOS) {
       FlutterLocalNotificationsPlatform.instance =
           MacOSFlutterLocalNotificationsPlugin();
-    } else if (defaultTargetPlatform == TargetPlatform.linux) {
-      FlutterLocalNotificationsPlatform.instance =
-          LinuxFlutterLocalNotificationsPlugin();
     }
   }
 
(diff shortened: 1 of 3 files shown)


OlaiyaO:
It appears you've solved the problem and I've used the "flutter pub upgrade" command but it still reproduces the message...

Package file_picker:windows references file_picker:windows as the default plugin, but it does
not provide an inline implementation.
Ask the maintainers of file_picker to either avoid referencing a default implementation via
`platforms: windows: default_package: file_picker` or add an inline implementation to
file_picker via `platforms: windows:` `pluginClass` or `dartPluginClass`.

MaikuB:
Please take a closer look at the logs you're pasting

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
```diff
@@ -38,9 +38,6 @@ class FlutterLocalNotificationsPlugin {
     } else if (defaultTargetPlatform == TargetPlatform.macOS) {
       FlutterLocalNotificationsPlatform.instance =
           MacOSFlutterLocalNotificationsPlugin();
-    } else if (defaultTargetPlatform == TargetPlatform.linux) {
```
MaikuB, line 41: Hi @proninyaroslav. Wanted to see if I could get your help to confirm something in case I missed anything. With this PR and the change done in https://github.com/MaikuB/flutter_local_notifications/pull/2371, the stub implementation in the Linux package can be removed. Is this correct?

proninyaroslav: Well, if the plugin registration occurs in a separate package with its own method, then registration here is already redundant.

MaikuB: Thanks

PR #2406, opened 2024-09-17
[flutter_local_notifications] fix: handle exception while loading resource name

Goddchen:
As discussed in #2309

We deployed the fix to production and we see no more errors 🥳


MaikuB:
It's failing as it found the code wasn't formatted to meet, tried to apply but then failed to push the changes. I'm not sure if this is to do with permissions on your fork and if this could be solved by changing the settings on the PR so that changes could've been pushed.

If you want to try manually fixing it yourself then you could try following https://github.com/google/google-java-format?tab=readme-ov-file#using-the-formatter

Goddchen:
should be fixed now, thanks for the hint @MaikuB

MaikuB:
Thanks for a lot. Appreciate that you put the effort to test the changes and contribute back

PR #2418, opened 2024-10-04
[flutter_local_notifications] fix: Fix platform initialization to allow mocking

provokateurin:
My app initializes the plugin like this:
```dart
final localNotificationsPlugin = FlutterLocalNotificationsPlugin();
await localNotificationsPlugin.initialize(
    ...
);
...
```
To test this I want to mock the plugin like this:
```dart
import 'package:plugin_platform_interface/plugin_platform_interface.dart';
import 'package:mocktail/mocktail.dart';

class _LocalNotificationsPlatformMock extends Mock
    with MockPlatformInterfaceMixin
    implements FlutterLocalNotificationsPlatform {}

void main() {
    late FlutterLocalNotificationsPlatform localNotificationsPlatform;

    setUp(() {
        localNotificationsPlatform = _LocalNotificationsPlatformMock();
        FlutterLocalNotificationsPlatform.instance = localNotificationsPlatform;
    });

    test(...);
}
```

The problem with the current initialization is that calling `FlutterLocalNotificationsPlugin()` will override the `FlutterLocalNotificationsPlatform.instance` which makes it impossible to mock it.

Using the hybrid platform implementation to let the platform register the instance allows mocking the instance because it will not be overridden.

### DIFF
--- a/flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
+++ b/flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
@@ -25,21 +25,7 @@ class FlutterLocalNotificationsPlugin {
   /// Factory for create an instance of [FlutterLocalNotificationsPlugin].
   factory FlutterLocalNotificationsPlugin() => _instance;
 
-  FlutterLocalNotificationsPlugin._() {
-    if (kIsWeb) {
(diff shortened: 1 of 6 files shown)


MaikuB:
Thanks!

PR #2440, opened 2024-10-30
[flutter_local_notifications] Gson 2.12.0

koji-1009:
https://github.com/google/gson/releases/tag/gson-parent-2.11.0

Update Gson to 2.11.0. Gson 2.11.0 provides the default ProGuard / R8 rules, so there is no need to write a progurad-rules for Gson.

https://docs.flutter.dev/release/breaking-changes/android-kitkat-deprecation

Gson 2.11.0 only supports Android 21+. Also, starting with Flutter 3.22, support will be for Android 21+.
Since there are very few Andorid 19 devices on the market, updating the support version will have little impact.


MaikuB:
Thanks for the PR. This was something I was already thinking am holding out at the moment until some more testing can be done. In addition to that, Flutter 3.22 is still rather new and a lot of apps may not be ready to move to it yet. Flutter team themselves tend to have the minimum Flutter version be N-2 stable version.

Something else to note is that the GSON changelog mentions "you might not need any special Gson configuration anymore if your classes have a no-args constructor and use @SerializedName for their fields". This plugin doesn't have classes with a no-args constructor so this could very mean that a rules file is still needed. I don't know if you got to test this one out yourself already but it may be possible we'll see different behaviour on different devices

koji-1009:
Most classes in flutter_local_notification, especially `/models`, are given `@Keep`. Therefore, there is no need to set a proguard rule for flutter_local_notification.

https://developer.android.com/reference/androidx/annotation/Keep

Comparing the two rules below, it appears that they both keep the same content, with minor differences in description. Since there are no rules required by flutter_local_notification, I thought it would be a good idea to use the rules provided by GSON, which is why I created the PR.

https://github.com/MaikuB/flutter_local_notifications/blob/flutter_local_notifications-v18.0.0/flutter_local_notifications/example/android/app/proguard-rules.pro
https://github.com/google/gson/blob/gson-parent-2.11.0/gson/src/main/resources/META-INF/proguard/gson.pro

---

I ran the exmaple app in release mode and checked the `configuratin.txt` file.
The content of the pro file is merged in the `flutter_local_notification/example/build/app/outputs/mapping/release/configuration.txt` generated when the `flutter build apk` is done.

MaikuB:
Yep you're right. Thanks for the reminder as I forgot I did do a comparison on this before by checking the mapping as you did. However, I'd still like to hold out on this as I don't want to bump the minimum Flutter version yet

koji-1009:
> However, I'd still like to hold out on this as I don't want to bump the minimum Flutter version yet

Understood. I can reference gson 2.11.0 on the app side so there is no problem. If you are willing to drop support for KitKat, I would be happy to have gson updated. Thanks.

MaikuB:
@koji-1009 thanks for your patience. The bump was done in https://github.com/MaikuB/flutter_local_notifications/pull/2542. Are you able to revisit this PR? Seems like GSON had updates since then too

koji-1009:
@MaikuB 
Thanks for the mentions. I will check and rebase.

MaikuB:
Thanks for updating in a short time frame. Merged in now :)

PR #2456, opened 2024-11-12
[various] prepare for 19.0.0 dev.1 release

MaikuB:


Levi-Lesches:
> Summary: I treat it as breaking change if an update prevents a consumer from being able to compile their app based on their current setup

Still not at my computer so I can't test this right now, but I'm fairly certain pub won't get a package version that conflicts with your SDK version when it can find one that doesn't, which would be true for all current users of this plugin. The only time I can see there being an issue is someone using the package _starting now_ and needing to update their sdk immediately, which is true of breaking changes as well

> To use an example, from a consumer's perspective if I'm on Flutter 3.13, if there's no breaking compatibility issues with the next major release of the plugin, I'd expect to be able to get this next major release but that won't work.

I'm not sure semver really means that. It's expected that new features will depend on newer SDKs and dependencies. `dart pub outdated` often shows me that some packages have minor versions available that I can't have, and that's normal because once I upgrade my SDK, those warnings go away. 

> This also avoids the situation where taking the other approach could lead to being in a situation where a minor and patch updates to same major release cannot flow through to consumers unless they upgrade and may not be ready to upgrade.

Isn't that more true of a breaking change? With a minor change, all it takes is `flutter upgrade`, which many will find themselves doing for other projects that don't use this plugin. With a breaking change, users have to upgrade their SDKs _and_ manually look through the changelog to decide if using this version is worth the migration time, for every package in every project they use. So the minor bump would actually get this version out to more users faster, no?

MaikuB:
Summary: I'm still continuing treating it as a breaking change as not doing so has caused issues for the community before. I don't want a repeat of this, those using my plugin and would rather err on the side of caution

> Still not at my computer so I can't test this right now, but I'm fairly certain pub won't get a package version that conflicts with your SDK version when it can find one that doesn't

True it doesn't provided constraints are setup correctly. Even then I'm not referring to just using the Flutter tooling to automatically find the appropriate version. I'm also referring to other approaches such as manually update pubspec.yaml or using IDE plugins it detects a new version where it won't check the SDK versions you have installed. To clarify when I said "I treat it as breaking change if an update prevents a consumer from being able to compile their app based on their current setup"

> I'm not sure semver really means that.

This is why I said "From a pure API perspective it may not be". Even then, there's been debate before on what "API" means and how this could be elaborated upon further (see https://github.com/semver/semver/issues/468). Note I did use to take the approach of minor update but this has caused issues with the community explaining their point of view and have around to not taking "API" literally and consider other factors on the consumer/end user's experience e.g. support for old environment being dropped

> Isn't that more true of a breaking change? With a minor change, all it takes is flutter upgrade, which many will find themselves doing for other projects that don't use this plugin. With a breaking change, users have to upgrade their SDKs and manually look through the changelog to decide if using this version is worth the migration time, for every package in every project they use. 

- I'm not sure if my example here was understood properly but before I go further into that, what you said there sounds odd. If plugins released minor updates then I would expect to only need to run `flutter pub upgrade` not `flutter upgrade`
- Regarding checking changelog, this is true but a lot of people don't check changelogs. Bear in mind I also mentioned that not treating it as a breaking change has caused issues before. Flutter upgrades tends to bump other transitive dependencies it has too
- I'm going to elaborate more on the example where this goes out as a minor version causes problems. Suppose this went out as 18.1.0 and there's a critical bug to fix on Android side afterwards. This would go out as 18.1.1. However, there may be devs that were on 18.0.0 with Flutter 3.13 and not ready to get Flutter 3.19+. That patch release is being "gatekeeped" by a Flutter upgrade unless the patch is "backported" to 18.0.1. This is a process I already done before for hotfixes like this but highlights one of the missing nuances when it comes semantic versioning. It would end up in the situation where the same major version could get identical patch updates. Similarly, other updates to the same major version may not be end up being available due to the dependency on updating Flutter SDK. Bumping the major version instead helps provide a clearer delineation i.e. there's a 19.0.1 patch and 18.0.1 patch

PR #2466, opened 2024-11-15
Web Support

Levi-Lesches:
> [!note]
> This is a work in progress

Adds web support. The plan is as follows, taken from https://github.com/MaikuB/flutter_local_notifications/issues/481#issuecomment-2473307972:


Levi-Lesches:
@MaikuB I'll once again put in my request to disable the `always_specify_types` and `lines_longer_than_80_chars` lints, as they cause a lot of headache when trying to write readable code here. Disabling them now won't require changing old code, but will allow cleaner code going forward, and there may be a fair amount of code since this is a big feature. I'll also point out again that `always_specify_types` [contradicts Effective Dart](https://dart.dev/effective-dart/design#dont-redundantly-type-annotate-initialized-local-variables) and there are even replacements such as [`omit_obvious_local_variable_types`](https://dart.dev/tools/linter-rules/omit_obvious_local_variable_types) and [`specify_nonobvious_local_variabletypes`](https://dart.dev/tools/linter-rules/specify_nonobvious_local_variable_types). As for line lengths, the new formatter [will allow custom line lengths](https://github.com/dart-lang/dart_style/pull/1571), for which I find 100 is a reasonable limit, especially with variable names as long as `WebFlutterLocalNotificationsPlugin`.

Levi-Lesches:
Love this test feedback!
```console
 Failed to load ".../flutter_local_notification_test.dart": Instance of 'VmServiceDisappearedException'
```

therickys93:
Hi @Levi-Lesches, I would like to know how I can install this alpha version inside my project. Can you please write a simple procedure on how to use it? Thanks so much

Levi-Lesches:
Look into git dependencies in your pubspec, and set the URL to my repo and ref to my branch

therickys93:
> Look into git dependencies in your pubspec, and set the URL to my repo and ref to my branch 

thanks so much. Is there an example on how to show a basic notification (title and subtitle) like on the other platforms? I cannot find a simple example in your code.

Levi-Lesches:
It's been a while since I've touched this PR, but `show()` should work. Are you able to run the example app on chrome?

MichalNemec:
Hello @Levi-Lesches is there an update or ETA to have basic functionality merged?

Levi-Lesches:
I hear your calls! Foreground callbacks are now supported!

```dart
await plugin.initialize(
  initializationSettings,
  // same as other platforms, but now it'll work on web too!
  onDidReceiveNotificationResponse: myHandler,
);
```

Oh, but this commit also adds a new package to the mix, which is unpublished, so you'll need to clone this whole repository and use path overrides for `flutter_local_notifications_web`

----

Launch details are also now supported!

```dart
// Same as every other platform!
final launchDetails = await plugin.getNotificationAppLaunchDetails();
```

If a notification is tapped when your tab is closed, it will be re-opened with a bunch of query parameters. Calling this function will read them and clear them so the user doesn't see them for too long. 

----

I'm going to clean up the PR now. Here's what I did _not_ add: 
- background handlers: when a tab is closed, it is terminated. Browser tabs don't really have a standard suspended state as far as I know, and the code I wrote will just treat them as terminated and open a new tab
- scheduled notifications: The browser just doesn't offer the same flexibility as an app here, so I don't think this is reasonably possible. 
  - The Periodic Background Sync API looked promising, but
      - permission is granted when the browser "trusts" you, and there's no way to request it
      - the interval parameter is just a suggestion, and the browser will decide when to run your code based on how often the user interacts with your page. Being a few seconds or even minutes off is acceptable for this package, but this is too unreliable
      - Not supported by Firefox or Safari

Levi-Lesches:
@MaikuB I'm ready for a review! 

There a handful of lints that are coming up from the `master` branch that aren't related to this PR. I opened https://github.com/MaikuB/flutter_local_notifications/pull/2687 for those

I couldn't really think of a great way to unit test this since it's so close to the actual JS layer and there's barely any plain Dart, so I just did the usual in making the really detailed examples. Let me know if you have a specific test case in mind.

I also didn't touch changelogs because I see you have a lot of PRs open and didn't want to conflict. This release is non-breaking though! 

I added snippets in the web-only README, and made some light changes to the main README, but I want to save my main documentation effort for #2477 and adding a web guide to that PR

MaikuB:
Thanks I'll take a look when I can do so. Regarding changelogs, that is something I can deal as part of resolving merge conflicts. With that in mind, if your wanted to, you can add details under "vNext" to represent next release. This is optional though

flutter_local_notifications/example/pubspec.yaml
```diff
@@ -1,6 +1,7 @@
 name: flutter_local_notifications_example
 description: Demonstrates how to use the flutter_local_notifications plugin.
 publish_to: none
+version: 1.0.0+0
```
MaikuB, line 4: Was there a reason for this addition? It's only meant be given anything that gets published. This is also why it's standard for example apps for packages and plugins to omit these

flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
MaikuB, line 1: There are some other changes that appear to be missing from this file

- `periodicallyShow()` and `periodicallyShowWithDuration()` is missing code to resolve web implementation within the `kIsWeb` check and call the associated method
- `zonedSchedule()` should throw an `UnsupportedError` for web. In hindsight, plugin likely needs a review to make use of this error instead of `UnimplementedError` but better to tackle separately outside of PR

flutter_local_notifications/pubspec.yaml
```diff
@@ -12,8 +12,10 @@ dependencies:
     sdk: flutter
   flutter_local_notifications_linux: ^6.0.0
   flutter_local_notifications_windows: ^1.0.3
+  flutter_local_notifications_web: ^1.0.0
   flutter_local_notifications_platform_interface: ^9.1.0
   timezone: ^0.10.0
+  universal_platform: ^1.1.0
```
MaikuB, line 18: Can you remove this dependency? It's not actually being used by the main plugin and was only used by the example app

flutter_local_notifications_web/lib/src/plugin.dart
```diff
@@ -0,0 +1,204 @@
...
+          payload: payload.nullIfEmpty,
+          actionId: action.nullIfEmpty,
+        ),
+      );
+    }
+  }
+
+  @override
+  Future<List<ActiveNotification>> getActiveNotifications() async {
+    if (_registration == null) {
```
MaikuB, line 141: Looks like this only returns the IDs of each notification. Can it return more details than that? I've only done a quick test but looked like it could return the title and body. Not sure if payload could also be returned

flutter_local_notifications_web/lib/src/plugin.dart
```diff
@@ -0,0 +1,204 @@
...
+      return <ActiveNotification>[];
+    }
+    final List<ActiveNotification> result = <ActiveNotification>[];
+    final Set<int> ids = <int>{};
+    final List<Notification> jsNotifs =
```
MaikuB, line 146: This is a general comment that applies to rest of PR and I understand there's an element of subjectiveness...but are able to avoid shortening variable names and name them in a way that allows code to be read more closer to an English sentence? Ones that stick out 

- `result` -> `activeNotifications`: will give more reasoning for this one as others are more self-explanatory. `result` is generic and whilst I can see how the intent was that the result of the method is to return a collection, with how the code functions, it's awkward to say a notification was added to a result.
- `jsNotifs` -> `notifications` 
- `jsNotification` -> `notification`: note this also avoids inconsistency between how how the plural used for the collection wasn't `jsNotifications`
- `notif` -> `activeNotification`

flutter_local_notifications_web/lib/src/plugin.dart
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  @override
+  Future<void> cancel(int id, {String? tag}) async {
+    if (_registration == null) {
+      return;
+    }
+    final List<Notification> notifs =
+        await _registration!.getDartNotifications();
+    for (final Notification notification in notifs) {
```
MaikuB, line 167: A couple of comments on this one
- the logic on this one looks incorrect. The purpose of method was to either cancel a notification based on the ID or when *both* ID and tag matches. The or portion of the if statement is missing the ID check
- with the above in mind, iteration can be avoided and after importing the `collection` package, can use [`firstWhereOrNull`](https://api.flutter.dev/flutter/package-collection_collection/IterableExtension/firstWhereOrNull.html) that is more succinct for situations like this to find the matching notification to cancel

flutter_local_notifications_web/lib/src/details.dart
```diff
@@ -0,0 +1,83 @@
...
+
+  /// The timestamp of the event that caused this notification.
+  ///
+  /// This doesn't always have to be the time the notification was sent. For
+  /// exmaple, if the notification represents a message that was sent a while
+  /// ago, you can back-date this timestamp to when the message was originally
+  /// sent instead of when it was received.
+  final DateTime? timestamp;
+
+  /// An optional vibration pattern
```
MaikuB, line 71: Missing a full stop at the end of first sentence

flutter_local_notifications_web/lib/src/plugin.dart
```diff
@@ -0,0 +1,204 @@
...
+        .showNotification(
+          title ?? '',
+          details.toJs(id, body, payload),
+        )
+        .toDart;
+  }
+
+  /// Initializes the plugin.
+  Future<bool?> initialize({
+    DidReceiveNotificationResponseCallback? onNotificationReceived,
```
MaikuB, line 75: can you rename all instances of `onNotificationReceived` to be `onDidReceiveNotificationResponse`?

flutter_local_notifications_web/lib/src/details.dart
```diff
@@ -0,0 +1,83 @@
...
+  /// An optional URL to a monochrome icon to show next to the title.
+  final Uri? badgeUrl;
+
+  /// An optional URL to an image to show as the app icon.
+  final Uri? iconUrl;
+
+  /// An optional URL to an image to show in the notification.
+  final Uri? imageUrl;
+
+  /// The language code of the notification's content, eg `en-US`.
```
MaikuB, line 43:
```suggestion
  /// The language code of the notification's content  e.g. `en-US`.
```

Besides this, do you know what the default behaviour is when this is left null? If so, can you add a paragraph to describe?  Mentioning this though as `lang` is expressed as optional but a notification always has ties to a language. Presumably it defaults to relying on the language code/locale of the environment that is running the app.

flutter_local_notifications_web/web/notifications_service_worker.js
```diff
@@ -0,0 +1,62 @@
+'use strict';
+
+// Experimental function to save a notification to the IndexedDB database
```
MaikuB, line 3: As it's not used, can this be removed? This to expose on what is necessary. Besides that, I assume if this was left used by others, it creates additional maintenance overhead in tracking what could be a breaking change if removed

flutter_local_notifications/example/lib/main.dart
```diff
@@ -15,12 +15,17 @@ import 'package:image/image.dart' as image;
 import 'package:path_provider/path_provider.dart';
 import 'package:timezone/data/latest_all.dart' as tz;
 import 'package:timezone/timezone.dart' as tz;
+import 'package:universal_platform/universal_platform.dart';
```
MaikuB, line 18: Would it be possible to have the example app written in a way that doesn't depend on this package? Reasons behind this thought

- Attempting to minimise dependencies on packages that aren't official ones as much as possible as they can end up not being maintained, run into issues etc. Avoiding this will also provide guidance for developers on how to handle such scenarios without relying on third party packages
- It's easy to miss that `Platform` now points to `UniversalPlatform` that developers run into issues. It looks like an easy change but is actually more significant than it looks. Flutter is more commonly used for mobile but the example app has had a notable change around how platform detection is done just to deal with web and convenience of this example app. Given this is an example app, it can cause developers to think using the `universal_platform` is recommended even for mobile development when it's not necessary

flutter_local_notifications_web/lib/src/action.dart
MaikuB, line 1: As all of this is based on a proposal as opposed to a standard, could this be removed? It being part of a proposal means it's subject to change and breaking so want to avoid having the plugin in this situation that in turn adds more burden around maintenance

MaikuB (changes requested):
Thanks so much for your effort @Levi-Lesches. I left a number of comments throughout the PR. Let me know if you got any questions or comments

flutter_local_notifications_web/pubspec.yaml
```diff
@@ -0,0 +1,29 @@
+name: flutter_local_notifications_web
+description: Web implementation of the flutter_local_notifications plugin
+version: 1.0.0
```
MaikuB, line 3: Forgot to mention something I'd be changing but also happy for you to do is that whilst this PR has been open for a while, when web support is available, it'll start of as being part of a prerelease beginning with the `-dev.1` suffix

Gaurav-CareMonitor:
@Levi-Lesches is this still being worked on?

MaikuB:
Closing this one as the work has been carried on in https://github.com/MaikuB/flutter_local_notifications/pull/2757

Issue #2446, opened 2024-11-07
Upgrading from version 17.2.3 to 17.2.4 may have introduced crashing behavior


leaf-node:
**Describe the bug**

App randomly crashes every few hours, or crashes when trying to start it after the app is stopped by the system following app optimization. My app is using a foreground service notification that at least some of the time may be persisting when the app and its background process gets closed by the user.

Tested on Android 15 with 34 target SDK with flutter_local_notifications 17.2.4.

**To Reproduce**

1. Upgrade GrapheneOS version
2. Reboot phone
3. Wait for apps to optimize
4. Press GrapheneOS notification to restart apps
5. Start app manually
6. It crashes

Crashing also occurs randomly every few hours. I started noticing this issue in the last few days, about a week after I upgraded dependencies for my app, including this module from version 17.2.3 to 17.2.4.

**Expected behavior**

The app shouldn't crash.

**Sample code to reproduce the problem**

I don't have source code available for this yet, hoping that it's an easy bug to solve. If not, please let me know, and I'll consider creating a basic app with this problem. Thanks!

** Crash Logs **

```
type: crash
osVersion: google/shiba/shiba:15/AP3A.241005.015/2024110400:user/release-keys
flags: dev options enabled
package: com.example.foo:1, targetSdk 34
process: com.example.foo
processUptime: 92 + 323 ms

java.lang.RuntimeException: Unable to start service com.dexterous.flutterlocalnotifications.ForegroundService@c4128db with null: java.lang.NullPointerException: Attempt to invoke virtual method 'java.io.Serializable android.content.Intent.getSerializableExtra(java.lang.String, java.lang.Class)' on a null object reference
at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:5199)
at android.app.ActivityThread.-$$Nest$mhandleServiceArgs(Unknown Source:0)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2478)
at android.os.Handler.dispatchMessage(Handler.java:107)
at android.os.Looper.loopOnce(Looper.java:232)
at android.os.Looper.loop(Looper.java:317)
at android.app.ActivityThread.main(ActivityThread.java:8744)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580)
at com.android.internal.os.ExecInit.main(ExecInit.java:50)
[...]

PR #2475, opened 2024-11-28
[flutter_local_notifications] Fix crashes related to Android foreground service start type

EPNW-Eric:
See MaikuB/flutter_local_notifications#2446

### DIFF
--- a/flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
+++ b/flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
@@ -262,6 +262,8 @@ class AndroidFlutterLocalNotificationsPlugin
   /// Information on selecting an appropriate `startType` for your app's use
   /// case should be taken from the official Android documentation, check [`Service.onStartCommand`](https://developer.android.com/reference/android/app/Service#onStartCommand(android.content.Intent,%20int,%20int)).
   /// The there mentioned constants can be found in [AndroidServiceStartType].
+  /// WARNING: We strongly disadvise using [AndroidServiceStartType.startSticky]
+  /// because it can lead to crashes, see this [github issue](https://github.com/MaikuB/flutter_local_notifications/issues/2446).
   ///
   /// The notification for the foreground service will not be dismissible
   /// and automatically removed when using [stopForegroundService].
@@ -278,7 +280,7 @@ class AndroidFlutterLocalNotificationsPlugin
   Future<void> startForegroundService(int id, String? title, String? body,
       {AndroidNotificationDetails? notificationDetails,
       String? payload,
-      AndroidServiceStartType startType = AndroidServiceStartType.startSticky,
+      AndroidServiceStartType startType = AndroidServiceStartType.startRedeliverIntent,
       Set<AndroidServiceForegroundType>? foregroundServiceTypes}) {
     validateId(id);
     if (id == 0) {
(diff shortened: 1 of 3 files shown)


Levi-Lesches:
LGTM for changing the default and the docs, but I still think it's wrong that we don't try to handle the error condition at all. 

I know you mentioned that a crash would help people realize that they're missing behavior, but my question is, what _should_ we be doing for users when the intent is null? From what I understand, this isn't an error condition, just no work for us to complete at this point, so it should be safe to return. As far as I know, we never claimed to be able to restart work after sudden termination, so it's not like people are expecting this from us.

EPNW-Eric:
So [`onStartCommand`](https://developer.android.com/reference/android/app/Service#onStartCommand(android.content.Intent,%20int,%20int)) requires us to return one of the start flags `START_STICKY`, `START_NOT_STICKY` and `START_REDELIVER_INTENT`. Usually we are getting this from the intent, but without intent what should we return? Returning `START_STICKY` (which was most likely the start flag before system termination of the service) would return in an endless loop? So return `START_NOT_STICKY` instead? And maybe even call [`stopSelf`](https://developer.android.com/reference/android/app/Service#stopSelf())?

Levi-Lesches:
From `START_REDELIVER_INTENT`: 
> This Intent will remain scheduled for redelivery until the service calls [stopSelf(int)](https://developer.android.com/reference/android/app/Service#stopSelf(int)) with the start ID provided to [onStartCommand(Intent, int, int)](https://developer.android.com/reference/android/app/Service#onStartCommand(android.content.Intent,%20int,%20int))

Good point, it seems we must call `stopSelf()` when we're done or else the app may end up repeating the service infinitely. Seems we already have [`stopForegroundService`](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/AndroidFlutterLocalNotificationsPlugin/stopForegroundService.html) which calls `stopService`, which should accomplish the same as `stopSelf`. 

So, given all that: 
- if we use `startSticky`, we need to handle the null case. I can't think of a good thing to do here besides returning immediately. 
- if we use `startNotSticky` and the app was suddenly terminated while working, it won't restart automatically
- if we use `startRedeliverIntent`, it won't go away unless the dev calls `stopForegroundService`

Most importantly though, I just realized... a _real_ service is meant to do real work. This plugin only uses services to implement permanent notifications. That means the real work is still being done in Dart, which should really use `startService` and `stopService` as needed. **That also means restarting the service won't have any actual effect besides showing a notification _telling_ the user something is happening -- even though it isn't** (evidenced by `onStartCommand` only showing the notification and returning, even in the best case).

We should probably change the default to `startNotSticky` instead of `startRedeliverIntent` since that's the only mode we _actually_ support. We should mark the enum and parameter as `@deprecated`, as they will only crash or in the best case, do nothing. If someone is trying to use these to restart real work, they should get a warning that it isn't doing that, and they should be using a different package like [`flutter_foregound_task`](https://pub.dev/packages/flutter_foreground_task) or write their own native code, which will allow them to implement real logic. We should also change the documentation and README to reflect this.

EPNW-Eric:
I was asking @leaf-node in #2446 to test if using `START_NOT_STICKY` can achive the desired behaviour. While I agree there is no use for `START_STICKY` I would await the answer before proceeding here, maybe we get some additional insights if `START_REDELIVER_INTENT` is not usefull.

EPNW-Eric:
See #2479

Issue #2479, opened 2024-12-02
Android docs: Clarify the usage of `startForegroundService()`


Levi-Lesches:
The docs for [`startForegroundService()`](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/AndroidFlutterLocalNotificationsPlugin/startForegroundService.html) seems to be inaccurate in a few ways: 

1. The plugin does technically start a foreground service, but not in the normal way. A normal foreground service is associated with some work, like downloads, media playback, etc. In this plugin, the foreground service exits immediately after being created: its `onStartCommand()` returns after creating a notification, and its `onBind()` just returns null. 

2. The docs say to call the function multiple times to update the notification, but the standard way to update a foreground service notification is to call `show()` with the same ID (see #2407)

3. The docs claim to support any `startType`, but the plugin doesn't actually handle cases other than `START_NOT_STICKY` (see #2475)

4. The docs claim the notification will not be dismissible, but since API 33 (the plugin only supports 34+), [foreground service notifications are dismissable by default](https://developer.android.com/develop/background-work/services/foreground-services#user-dismiss-notification). 

5. The docs for the function show a lot of Android-specific setup. They could instead link to the setup steps of the README (see also #2477) 

But most importantly, when combining points 1 and 3, I've come to realize that this function does not actually start a "real" foreground service. It shows the notification, yes, but the point of a foreground service is to promote some logic-heavy code to the foreground to prevent it from being background by the system. As far as I can tell, this plugin isn't doing that, and in my opinion, that would be out-of-scope and better left to a plugin like [`flutter_foreground_service`](https://pub.dev/packages/flutter_foreground_service) or [`flutter_foreground_task`](https://pub.dev/packages/flutter_foreground_task)).

It seems from [this answer](https://stackoverflow.com/questions/72389156/does-an-empty-foreground-service-keep-an-android-app-alive) that even if a service does no work, the whole app is elevated to foreground status. This clarifies why the Java code isn't doing anything interesting, and why the Dart function doesn't take a callback to run, like the other plugins I linked do.

The best "service-like" functionality one can hope to achieve with this plugin is to
- start the foreground notification with `AndroidNotificationDetails(ongoing: false)`
- start running Dart code
- stop the foreground service

@epnw-eric, what are your thoughts? If you want to add these to your PR, that would be great, otherwise I can do them after yours is merged.

Levi-Lesches:
[...]

PR #2481, opened 2024-12-02
[Android] Remove `AndroidServiceStartType` and update docs 

Levi-Lesches:
See #2479 for details on why this was needed. Tl;dr, the function wasn't exactly doing what it was saying

### DIFF
--- a/flutter_local_notifications/CHANGELOG.md
+++ b/flutter_local_notifications/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [19.0.0-dev.2]
+* **Breaking change** [Android] Removed `AndroidServiceStartType` as it wasn't having any effect. Remove the parameter from your calls to `startForegroundService()`.
+
 ## [19.0.0-dev.1]
 
 * **Breaking change** bumped minimum Flutter SDK requirement to 3.19.0 and Dart SDK requirement to 3.3.0
@@ -156,7 +159,7 @@
 # [15.1.0]
 
 * [iOS][macOS] added the ability to request provisional permissions. On iOS, this is only applicable to iOS 12 or newer. On macOS, this property is only applicable to macOS 10.14 or newer. Thanks to the PR from [Tokenyet](https://github.com/MaikuB/flutter_local_notifications/pull/2022)
- 
+
 # [15.0.1]
 
 * [Android] fixed issue [2033](https://github.com/MaikuB/flutter_local_notifications/issues/2033) where notifications on scheduled using older version of the plugin would fail to have the next subsequent ones scheduled. This issue started occuring in 14.0 where support for inexact notifications was added using the `ScheduleMode` enum that was added and resulted in the deprecation of `androidAllowWhileIdle`. A mechanism was added to help "migrate" old notifications that had `androidAllowWhileIdle` specified but didn't account for how there are recurring notifications that were scheduled using older versions of the plugin prior to `androidAllowWhile` being added. This was also released part of the 14.1.2 hotfix release
@@ -165,7 +168,7 @@
 
 * **Breaking change** removed deprecated `schedule()`, `showDailyAtTime()` and `showWeeklyAtDayAndTime()` methods. Notifications that were scheduled prior to this release should still work
 * **Breaking change** removed `Time` class
-* [Linux] **Breaking change** calling `zonedSchedule()` on Linux will now throw an `UnimplementedError` to align with how their is a Linux implementation but the method hasn't been implemented 
+* [Linux] **Breaking change** calling `zonedSchedule()` on Linux will now throw an `UnimplementedError` to align with how their is a Linux implementation but the method hasn't been implemented
 * [iOS][macOS] **Breaking change** added supported for banner and list presentation options for iOS and macOS that is applicable for iOS 14.0 or newer and macOS 11 or newer. This is a breaking change as the values default to true and the alert presentation option is no longer applicable on these OS versions as Apple has deprecated it to be replaced by the banner and list presentations. Please ensure that if you target these OS versions that you configure the options appropriately for your application.
 * [Android] updated tags used when writing error logs. For corrupt scheduled notifications and error is logged the tag is now `ScheduledNotifReceiver` instead of `ScheduledNotifReceiver`. When logging that exact alarm permissions have been revoked the the tag is now `FLTLocalNotifPlugin` instead of `notification`
 * Updated API documentation related to the iOS/macOS notification presentation options to include links to Apple's documentations to show what they correspond to
@@ -261,10 +264,10 @@
      ...
  }
  ```
- 
+
 # [12.0.4]
 
-* Fixed issue [1796](https://github.com/MaikuB/flutter_local_notifications/issues/1796) where a `java.lang.ClassCastException` may be thrown on some Android devices when the `onDidReceiveBackgroundNotificationResponse` has been specified when calling `initialize()` 
+* Fixed issue [1796](https://github.com/MaikuB/flutter_local_notifications/issues/1796) where a `java.lang.ClassCastException` may be thrown on some Android devices when the `onDidReceiveBackgroundNotificationResponse` has been specified when calling `initialize()`
 
 # [12.0.3+1]
 
@@ -288,7 +291,7 @@
 
 # [12.0.1]
 
-* [Android][iOS] fixed issue [1721](https://github.com/MaikuB/flutter_local_notifications/issues/1721) where a crash occurs upon tapping on a notification action fbut the `onDidReceiveBackgroundNotificationResponse` optional callback hasn't been specified. 
+* [Android][iOS] fixed issue [1721](https://github.com/MaikuB/flutter_local_notifications/issues/1721) where a crash occurs upon tapping on a notification action fbut the `onDidReceiveBackgroundNotificationResponse` optional callback hasn't been specified.
 * [iOS] suppressed deprecation warnings where plugin was Apple's old notification APIs to support older iOS devices
 
 # [12.0.0]
@@ -322,7 +325,7 @@
   * `GET_ACTIVE_NOTIFICATION_MESSAGING_STYLE_ERROR_CODE` -> `getActiveNotificationMessagingStyle`
   * `PERMISSION_REQUEST_IN_PROGRESS` -> `permissionRequestInProgress`
 * [Android] **Breaking change** the `category` of the `AndroidNotificationDetails` now requires an instance of the newly added `AndroidNotificationCategory` class instead of a string. This was to improve the discoverability of the APIs and improve the semantics as the category can specified in a similar fashion to using an enum value
-* **Breaking change** callbacks have now been reworked. There are now the following callbacks and both will pass an instance of the `NotificationResponse` class 
+* **Breaking change** callbacks have now been reworked. There are now the following callbacks and both will pass an instance of the `NotificationResponse` class
   * `onDidReceiveNotificationResponse`: invoked only when the app is running. This works for when a user has selected a notification or notification action. This replaces the `onSelectNotification` callback that existed before. For notification actions, the action needs to be configured to indicate the the app or user interface should be shown on invoking the action for this callback to be invoked i.e. by specifying the `DarwinNotificationActionOption.foreground` option on iOS and the `showsUserInterface` property on Android. On macOS and Linux, as there's no support for background isolates it will always invoke this callback
   * `onDidReceiveBackgroundNotificationResponse`: invoked on a background isolate for when a user has selected a notification action. This replaces the `onSelectNotificationAction` callback
 * **Breaking change** the `NotificationAppLaunchDetails` has been updated to contain an instance `NotificationResponse` class with the `payload` belonging to the `NotificationResponse` class. This is to allow knowing more details about what caused the app to launch e.g. if a notification action was used to do so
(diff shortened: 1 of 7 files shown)


Levi-Lesches:
Seems the CI is failing because the example app is still using the old plugin syntax. I can open a PR for that sometime soon

EPNW-Eric:
Looks good to me 👍

martin-bertele:
> Seems the CI is failing because the example app is still using the old plugin syntax. I can open a PR for that sometime soon

fixed in here
https://github.com/MaikuB/flutter_local_notifications/pull/2482

leaf-node:
Wait, I think we shouldn't remove the `AndroidServiceStartType` option completely, because there are two valid values for this setting: the `START_NOT_STICKY` and `START_REDELIVER_INTENT`. The first would cause the foreground service to stop entirely if the system needs to pause that activity. And the latter would schedule the foreground service to resume when resources are available, correct? I don't think it's right to only allow use of the `START_NOT_STICKY` mode, because many people may want their notifications to persist.

If a value has to be hard-coded, it should be the `START_REDELIVER_INTENT` option so foreground services will continue to be reliable on Android. But I can see a use for both, so I suggest keeping the option in place, setting the default to `START_REDELIVER_INTENT`, and removing the `START_STICKY` option, because that one is not supported by this library.

MaikuB:
Not done testing on this but inclined to agree with @leaf-node and removing it seems more risky. I would suggest making this a required parameter. This way developers spend some time looking into the option that fits their scenario and the results are the consequence of the choice they made.

MaikuB:
@leaf-node thanks for the response. Based on my understanding, my preference would be to avoid removing the functionality as there could be cases where there are existing apps that choose to use this plugin to show the notification and start the operation independently. The one that seems to be problematic is the one tied to `START_STICKY`

By the way, I don't think I've seen it mentioned but what is your actual use case? I saw you mentioned that picking `START_NOT_STICKY` fixes the issue for you and makes sense as there wouldn't be a null intent. Even if a callback was being supported, avoiding that crash would still be needed. If you needed a callback so that work can resume on a restart, [`flutter_foreground_task`](https://pub.dev/packages/flutter_foreground_task) that @Levi-Lesches shared in #2479 looks to be more appropriate. For a callback to be supported, the only way that Dart code can be invoked in this scenario is through the use of background isolates. Background isolates are what allows Dart code to be run when the app isn't running and is the same mechanism used for notification actions that don't require interacting with a running app. `flutter_foreground_task` appears to support that already and would even appear to provide functionality to communicate back to the main isolate in the scenario and app is running

leaf-node:
@MaikuB The use case for my app is to fetch server alert statuses from third-party remote servers, and to notify the user if a new alert is found. I'm not using push notifications, so I create a foreground service, allowing my app to fetch from the network and create alert notifications.

As for START_STICKY vs START_NOT_STICKY, if I could, I would prerfer former for my app, because  (if I understand correctly) it will restart the app at the `.ForegroundService` class, but the Kotlin / Java glue code would have to function without access to the original `intent`. The START_NOT_STICKY option worked for me, but only in the sense that it didn't trigger a crash. START_REDELIVER_INTENT would be an ideal option, because with more work, it would allow the service to resume after the OS stops my app to clear up memory, etc. Currently, this notification package doesn't seem to recreate the foreground service after it is stopped.

As for background isolates, I'm pretty sure that they are implemented as separate threads on the same `FlutterEngine`, and platform code is capable of creating and stopping multiple of those. The background isolates exist in a FlutterEngine, so when the app stops, the flutter engine is stopped, and so are the isolates. When the service is running as a foreground service, I'm pretty sure that's mostly a notification that prolongs the life of the app even when not directly used. And the declaration of the foreground service in the AnroidManifest.xml file is used as an entry point for the OS to re-create a foreground service that was temporarily, but really and completely, stopped.

It's possible to create a FlutterEngine that calls into a specific entrypoint in the Dart code, skipping the `main()` function. So I'm trying to leverage this in my app. When I use that approach, the UI is not run, and the app process, including its service, uses less memory until the user launches the app. At that point, I kill the FlutterEngine that doesn't have a UI, because that Flutter engine was created with an Intent that doesn't offer the permission to run UI code.

And the sticky options are to tell the OS whether an app's foreground service should be restarted after it is killed by the OS for memory cleanup, etc. I suppose in some cases, it might be useful to have a foreground service that runs for a while, then somewhat arbitrarily stops, but in my case I would want something that always, or usually, comes back, even if it takes a long time.

I should also note that there is a 6 hour limit for foreground services in a 24 hour period, until the user interacts with the UI of the app again, unless the app uses a special permission that is harder to get into the Play Store. Some packages for background work operate by scheduling a series of timers. Between each tick, the app is not running, which might help with the time limit for the app.

I'm still new to this topic, so it's not all 100% clear to me, but I'm learning. @Levi-Lesches does that match what you understand about this topic?

MaikuB:
> @MaikuB The use case for my app is to fetch server alert statuses from third-party remote servers, and to notify the user if a new alert is found. I'm not using push notifications, so I create a foreground service, allowing my app to fetch from the network and create alert notifications.

What triggers this to run?

> It's possible to create a FlutterEngine that calls into a specific entrypoint in the Dart code, skipping the main() function. So I'm trying to leverage this in my app. When I use that approach, the UI is not run, and the app process, including its service, uses less memory until the user launches the app. At that point, I kill the FlutterEngine that doesn't have a UI, because that Flutter engine was created with an Intent that doesn't offer the permission to run UI code.

What you've described here is actually referring to background isolates and is what this plugin already leverages for handling notification actions that is required to trigger a callback on a background isolate. The reference in the official docs can be found [here](https://docs.flutter.dev/packages-and-plugins/background-processes) , which in turn refers to [this](https://medium.com/flutter/executing-dart-in-the-background-with-flutter-plugins-and-geofencing-2b3e40a1a124)

Levi-Lesches:
@leaf-node 

> Wait, I think we shouldn't remove the AndroidServiceStartType option completely, because there are two valid values for this setting: the START_NOT_STICKY and START_REDELIVER_INTENT
>
> ...
>
> START_REDELIVER_INTENT would be an ideal option, because with more work, it would allow the service to resume after the OS stops my app to clear up memory, etc. Currently, this notification package doesn't seem to recreate the foreground service after it is stopped.

@MaikuB 

> Based on my understanding, my preference would be to avoid removing the functionality as there could be cases where there are existing apps that choose to use this plugin to show the notification and start the operation independently. The one that seems to be problematic is the one tied to START_STICKY

So we all agree that `START_STICKY` is not useful in its current form, and is almost always causing null errors.

Now, about `START_REDELIVER_INTENT`: I haven't gotten around to testing this yet either (hence why this is still a draft), but my main concern is that this isn't doing what we want either. The _intention_ is that the OS should show the notification again and re-start the foreground service, along with any work it was doing. But, as far as I can tell, that's _not_ what's happening. It seems the foreground service's native side (ie, Java) is just starting the foreground service, showing the notification, and _not doing any work_. Because, as you mentioned, showing the notification is done (ultimately) in Java, and doing the work is done in Dart. And there's no link between the two like `package:flutter_foreground_service` or others have. 

If that's correct, then we should remove this option because it _sounds_ like something useful is happening, and a notification means the user _thinks_ something useful is happening, but nothing is happening at all. Again, haven't tested, but the code in `ForegroundService.java` is so short and contains no references to Dart code of any kind. It just shows the notification and quits immediately. This means that no matter what option you pass in today, `START_NOT_STICKY` is the _real_ behavior, and the APIs should reflect that.

leaf-node:
> What triggers this to run? Putting aside that there's an issue with START_STICKY for this plugin, currently I can't tell why would be preferred as well. Other thing is how do you handle this on iOS?

@MaikuB I'm running a timer in Dart that then calls code to perform an HTTP request each time data is fetched by the app. I don't have an iOS version, largely due to a lack of a foreground service feature there. As for `flutter_foreground_task`, it schedules work every 15 minutes, but I'd like more flexibility in the timing of work, and the reliability that a foreground service provides. Although, that plugin does support iOS, and a foreground service may have a limited lifespan. You make a good point, and hopefully I can take some of what I've learned and make some pull requests against existing plugins.

> This means that no matter what option you pass in today, START_NOT_STICKY is the real behavior, and the APIs should reflect that.

@Levi-Lesches Are you thinking that it would currently make sense to require the [`startNotSticky`](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/AndroidServiceStartType.html) parameter when launching a foreground service, for the sake of clarity, but then once there's support for calling back into Dart code from Java, it would make sense to allow the `startRedeliverIntent` option again?

Levi-Lesches:
> Are you thinking that it would currently make sense to require the [startNotSticky](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/AndroidServiceStartType.html) parameter when launching a foreground service, for the sake of clarity,

Yes, since anything else would be misleading, but

> then once there's support for calling back into Dart code from Java, it would make sense to allow the startRedeliverIntent option again?

I'm not sure this is ever going to happen, as we've mentioned that other packages like `flutter_foreground_task` do this better (and in a more cross-platform way), and I think it's better to leave it to them and consider it out-of-scope for this plugin, if @MaikuB agrees.

MaikuB:
>  I'm running a timer in Dart that then calls code to perform an HTTP request each time data is fetched by the app. I don't have an iOS version, largely due to a lack of a foreground service feature there. As for flutter_foreground_task, it schedules work every 15 minutes, but I'd like more flexibility in the timing of work, and the reliability that a foreground service provides.

@leaf-node thanks for the explanation. Have I missed something that's perhaps only in the implementation details or API docs? I've not taken a look at either but the only mention of 15 minutes  I've seen in the readme is specific to iOS and you don't have an iOS version for it be applicable

leaf-node:
> @leaf-node thanks for the explanation. Have I missed something that's perhaps only in the implementation details or API docs? I've not taken a look at either in depth but the only mention of 15 minutes I've seen in the readme is specific to iOS and you don't have an iOS version for it be applicable

@MaikuB Yeah, it looks like `flutter_foreground_task` supports other intervals, so I may have been thinking of a different plugin. I investigated a few, so I'm not sure why I passed on that one. Maybe it was because the isolate running my app's model needs to keep running while the UI is visible, even if the user chooses to disable all notifications, including the foreground service one. It should be possible to work around that issue, but I thought that I wouldn't need much Kotlin code in a custom approach. I ended up with almost 300 lines of Kotlin, which isn't too bad, and I feel like I learned from the process, so I'm happy with the result. We'll see about working out any subtle bugs though. : )

leaf-node:
(Sorry if I'm rehashing any of this)

The main caveat in my mind is that other plugins providing foreground services may rely upon this module to create the foreground notification. If the option to select `startSticky` or `startRedeliverIntent` is removed, then that code will have to be implemented in Java or Kotlin for each of those modules. It's possible  that won't be a lot of work, given that they might already have to write custom `onStartCommand()` code. For reference, here's [the code I ended up writing](https://github.com/okcode-studio/open_alert_viewer/blob/main/android/app/src/main/kotlin/studio/okcode/open_alert_viewer/OAVForegroundService.kt) for my app.

MaikuB:
@leaf-node yep but that is a similar kind of process that apps run would encounter if they make use of a dependency and there's a breaking change. More importantly though, hasn't it already been established that neither of these work properly and would therefore mean those dependencies would inherit the same kind of issues you reported?

leaf-node:
> More importantly though, hasn't it already been established that neither of these work properly and would therefore mean those dependencies would inherit the same kind of issues you reported?

Well, as Levi pointed out, the problem is more directly related to `onStartCommand()`. I suspected that various modules might currently use the foreground notification via this module, but create their own `onStartCommand()` method, however the two packages I checked use neither.

For instance, `flutter_foreground_task` module creates its own `onStartCommand()`, but also doesn't use `flutter_local_notifications`:

https://github.com/Dev-hwang/flutter_foreground_task/blob/a7f81365983f2fd6f90d4eb5e0686a204c48a4c0/android/src/main/kotlin/com/pravera/flutter_foreground_task/service/ForegroundService.kt#L127

The same goes for `flutter_foreground_service`:

https://github.com/cph-cachet/flutter-plugins/blob/f1f4ee07ca6cdd76f68be4ee3b98854916443712/packages/activity_recognition_flutter/android/src/main/java/dk/cachet/activity_recognition_flutter/ForegroundService.java#L29

MaikuB:
> Well, as Levi pointed out, the problem is more directly related to onStartCommand()

No, @Levi-Lesches  actually pointing out that it's to do with the service start types. That there is a crash to do with `onStartCommand` when `START_STICKY` is a *symptom* of using service start types that aren't relevant.  This is why @Levi-Lesches is proposing to remove support for this and `START_REDELIVER_INTENT` as based on how Android works with these options behind the scenes, they aren't really a fit for this plugin.

Another thing to note that any other plugin that would depend that currently depends on this one where `START_STICKY` is used would experience the same crash you reported yet there hasn't been any so there is more backing that removing these options is fine. Even if they did, I don't have a problem with them being removed

Levi-Lesches:
Right, @leaf-node just to clarify: while I'd also _like_ these features to be supported, the fact that you can use pass them as arguments right now is meaningless because _the feature doesn't actually work_. The crash you're seeing can be silenced, sure, but the important bit of "actually do meaningful work in the foreground" just isn't happening. Instead, all the work is happening in the Dart side, which is what we'd have to stick with because doing otherwise would be a very big change. 

So removing this should actually be harmless, because the feature never actually existed.

JinYi-Tsinghua:
Is this going to get merged? This has been causing my application to die for all users, multiple times a day, less-than-gracefully.

I mean even just merge an optional parameter so that startType can be passed would be a sufficient bandage.

PR #1970, opened 2023-04-26
Adds Android namespace

asaarnak:
Add a namespace attribute to the Android build.gradle, for compatibility with Android Gradle Plugin 8.0.
See:
https://github.com/flutter/packages/commit/6284c2d4e46a5d289e77cb03a9457543b97f750b

PR #2482, opened 2024-12-03
[flutter_local_notifications] Update gradle versions and declarations for example app

martin-bertele:
Followed https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply

and updated AGP and kotlin versions.

### DIFF
--- a/flutter_local_notifications/example/android/app/src/debug/AndroidManifest.xml
+++ b/flutter_local_notifications/example/android/app/src/debug/AndroidManifest.xml
@@ -1,5 +1,4 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.dexterous.flutter_local_notifications_example">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
     <!-- Flutter needs it to communicate with the running application
          to allow setting breakpoints, to provide hot reload, etc.
     -->
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -9,6 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: '17'
       - uses: axel-op/googlejavaformat-action@v3
         with:
           args: '--skip-sorting-imports --replace'
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -52,6 +52,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: '17'
       - uses: subosito/flutter-action@v2
         with:
           channel: stable
@@ -66,6 +70,10 @@ jobs:
       runs-on: ubuntu-latest
       steps:
         - uses: actions/checkout@v4
+        - uses: actions/setup-java@v4
+          with:
+            distribution: 'zulu'
+            java-version: '17'
         - uses: subosito/flutter-action@v2
           with:
             channel: stable
@@ -241,6 +249,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: '17'
       - uses: subosito/flutter-action@v2
         with:
           channel: stable
@@ -275,7 +287,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-java@v3
+      - uses: actions/setup-java@v4
         with:
           distribution: 'zulu'
           java-version: '17'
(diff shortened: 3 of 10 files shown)


flutter_local_notifications/example/android/app/src/debug/AndroidManifest.xml
```diff
@@ -1,5 +1,4 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.dexterous.flutter_local_notifications_example">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
```
Levi-Lesches, line 1: Nice! Just checking -- is this intentional? I found this in [the `<manifest>` docs](https://developer.android.com/guide/topics/manifest/manifest-element): 

> In the Gradle-based build system, starting with AGP 7.3, don't set the package value in the source manifest file directly. For more information, see [Set the application ID](https://developer.android.com/studio/build/configure-app-module#set-application-id).

Levi-Lesches (approved):

flutter_local_notifications/example/android/app/src/debug/AndroidManifest.xml
martin-bertele: I've run the Android Studio AGP Upgrade Assistant, any changes to .xml files came automatically, so good question.  In case take your edits :)

Levi-Lesches:
Hey @martin-bertele, Maiku ran the CI just now, seems you're getting [Java mismatches](https://github.com/MaikuB/flutter_local_notifications/actions/runs/12137271322/job/34172105150?pr=2482): 

```console
ERROR: ERROR: FAILURE: Build failed with an exception.

* Where:
Build file '/home/runner/work/flutter_local_notifications/flutter_local_notifications/flutter_local_notifications/example/android/app/build.gradle' line: 2

* What went wrong:
An exception occurred applying plugin request [id: 'com.android.application']
> Failed to apply plugin 'com.android.internal.application'.
   > Android Gradle plugin requires Java 17 to run. You are currently using Java 11.
      Your current JDK is located in /usr/lib/jvm/temurin-11-jdk-amd[64](https://github.com/MaikuB/flutter_local_notifications/actions/runs/12137271322/job/34172105150?pr=2482#step:5:65)
      You can try some of the following options:
       - changing the IDE settings.
       - changing the JAVA_HOME environment variable.
       - changing `org.gradle.java.home` in `gradle.properties`.
```

Notice how the Integrations tests are the only ones that passed on Android -- that's because they properly set up Java: 

```yaml
      - uses: actions/setup-java@v3
        with:
          distribution: 'zulu'
          java-version: '17'
```

Can you copy and paste that into the other failing tests as well?

MaikuB:
@martin-bertele I believe the build is failing as it needs changes that I tried to do in https://github.com/MaikuB/flutter_local_notifications/pull/2488 but looks like that would also need the changes done in this PR. Are you able to incorporate the changes I tried to do in this PR? Would help to ensure everything is green

Levi-Lesches:
@MaikuB If you can change your #2488 to be based on Martin's branch, that should allow you to test both changes at once.

Also, I wanted to point out that this PR upgrades the Android Gradle Plugin to 8.4, coming from 7.3.1. That shouldn't be an issue though. [This comment](https://github.com/flutter/flutter/issues/125621#issuecomment-1525995461) points out that it would break users of AGP 4.1, but... we've been mandating AGP >= 7.3.1 for a while now, and https://github.com/MaikuB/flutter_local_notifications/pull/1970 was merged in April 2023 with no major complaints.

MaikuB:
@Levi-Lesches yep I know. Provided the PR has permissions for it, I could've even pushed changes directly to @martin-bertele's but avoided doing so in case @martin-bertele was going to make changes since you mentioned what can be done already and could I end up wasting @martin-bertele's time. Since we've not heard back though I'll look to do so later if @martin-bertele hasn't gotten to do by the time I can to it

martin-bertele:
Unfortunately I got sick, if I need to change any permissions let me know this I can still do, otherwise I m available next week. Thank you

MaikuB:
It's quite possible you ticked it as I believe it's ticked by default but it's to do with this one https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

martin-bertele:
@Levi-Lesches @MaikuB you can go ahead and copy over any lines from this pr to the work you have, really use the way that works with minimum effort for you. It's been a coincidence that you were to upgrade the sample too.
I got it to run and thought then I ll contribute. :)

AGP and kotlin I was forced to upgrade since there was some compatibility complains.
After that, it built successfully.

Permissions on PR I am not sure how to set, let me know what you need there. Thanks for your great package!

MaikuB:
@martin-bertele all good as the permissions are fine as I've been able to push changes directly. Once it's all green and I've done some more testing/reviewing  then will merge it in

flutter_local_notifications/example/android/app/src/debug/AndroidManifest.xml
MaikuB: @Levi-Lesches  from what I saw, what you sent aligns with the change that was done as the application ID has been specified in the `build.gradle`. If you create a new Flutter app using a relatively recent of the SDK then you should find it doesn't have a package value either (note: I don't recall whenever Flutter started behaving like this but I remember it's been like this for a while). With that in mind, I'm going to mark this as resolved

Levi-Lesches (approved):
Compiles on Flutter 3.27.0 with no issues or warnings!

PR #2495, opened 2024-12-13
[windows] Bump ffigen to 16.1.0

Levi-Lesches:
Closes #2471 by manually bumping, re-running `ffigen`, then cleaning up the resulting Dart code.

The big change in ffigen here is that it better handles enums, by generating actual Dart enums instead of integers


Levi-Lesches:
Well there you have it: 

```console
Because flutter_local_notifications_windows depends on ffigen >=15.0.0 which requires SDK version >=3.4.0 <4.0.0, version solving failed.
```

So to merge this, we'd have to bump the SDK to `^3.4.0` and Flutter `^3.22.0`, which are the next versions of Dart and Flutter. It's not a terrible change, but this PR doesn't add any new functionality so its not a high priority.

MaikuB:
Happy to bump to Flutter 3.22 as that then makes the minimum Android API 21. https://github.com/MaikuB/flutter_local_notifications/pull/2440 is dependent on Flutter 3.22 as well but held off so that would allow Flutter 3.22 to be at least the two stable releases behind

Levi-Lesches:
Okay, confusingly, ffigen ^16.0.0 conflicts with Flutter 3.22 because of a conflict with the `analyzer` and `meta` packages. I've opened https://github.com/dart-lang/pub/issues/4488 to discuss how this could be caught sooner because my computer, with the latest Flutter, doesn't show these issues. Not sure what to do except wait in this case.

MaikuB:
I've made the change to bump the Flutter version to 3.22 but ffigen 16 is looking to not be compatible with it so going to close this one

Issue #2492, opened 2024-12-13
Refactor and redesign repeating notifications


Levi-Lesches:
I'm writing up the usage guide now (see #2477), and I find the scheduling notifications to be a bit long. Currently, if a user wants to schedule a repeating notifications, they can use: 

- `zonedSchedule()`, with a non-null `matchDateTimeComponents`
- `periodicallyShow()`, with a given `RepeatInterval`
- `periodicallyShowWithDuration()`, with a given `Duration`

Fundamentally, there are 3 options for a scheduled notification: 
- show once
- repeat given certain conditions, like date, day of week, etc
- repeat on a given interval
 
The only difference between `periodicallyShow()` and `periodicallyShowWithDuration()` is the ability to pass an arbitrary duration, but Android, iOS, and MacOS support an arbitrary duration, so the enum is not needed. In fact, on Android and Darwin, the enum is converted to milliseconds, and on Linux and Windows periodic notifications are not supported at all.

I believe all this can be simplified directly into `zonedSchedule`. I'm thinking something like this: 
```dart
sealed class NotificationRepeatDetails { }

class NotificationRepeatInterval extends NotificationRepeatDetails { 
  Duration duration;
}

class NotificationRepeatDate extends NotificationRepeatDetails { 
  DateTimeComponents components;
}

void zonedSchedule({
  required int id,
  required TZDateTime scheduleDateTime,
  required UILocalNotificationDateInterpretation iOSDateInterpretation, 
  required AndroidScheduleMode androidSchedulePrecision,
  NotificationRepeatDetails? repeatsOn,
  String? title,
  String? body,
  NotificationDetails? details, 
  String? payload,
);
```

Since this will be a breaking change anyway, I renamed some parameters. Here are some classic examples with the new approach: 

```dart
final location = getLocation(...);
final now = TZDateTime.now(location);

// Schedule a notification tomorrow at 8pm
plugin.zonedSchedule(
  id: 0,
  scheduledDateTime: now.copyWith(hour: 20, minute: 00),
  // title, body, details
);

// Schedule a repeating reminder, every day at noon
plugin.zonedSchedule(
  id: 1, 
  scheduledDateTime: now.copyWith(hour: 12, minute: 00),
  repeatsOn: NotificationRepeatDate(DateTimeComponents.time),
  // title, body, details
);

// Schedule a reminder every 30 minutes to take a break
plugin.zonedSchedule(
  id: 2, 
[...]

PR #2503, opened 2024-12-21
[flutter_local_notifications] Allow specifying start time of periodic notifications.

BadgerHobbs:
Firstly, thank your the continued work supporting and maintaining this package, it is much appreciated.

## Current Challenge

Currently if you schedule a notification using `periodicallyShow` the notification will be sent every day since the creation of the notification. As I understand this is the currently accepted and documented behaviour.

This means that you cannot set a periodic notification to start after a specific date, such as required in the below scenario.

### Example Scenario
- Current date is `2024-01-01 12:00:00`.
- Schedule a `zonedSchedule` notification for `2024-01-05 12:00:00` to tell the user that a task is due.
- Schedule a `periodicallyShow` notification from `2024-01-06 12:00:00` to tell the user that a task is overdue every day.

### Current Outcome
- User receives an overdue notification every day from the `2024-01-01 12:00:00` onwards.

### Desired Outcome
- User receives an overdue notification every day from `2024-01-06 12:00:00` onwards.

An unresolved question [here on StackOverflow](https://stackoverflow.com/questions/78393943/flutter-repeating-local-notifications) describes a similar issue.

## Changes

Inspired by the work/changes proposed by @henriklagergren in [this discussion](https://github.com/MaikuB/flutter_local_notifications/issues/92#issuecomment-1804514149), the following change has been made based off the latest master branch.

The existing `periodicallyShow` and `periodicallyShowWithDuration` methods have been updated with an optional parameter to replace the default `calledAt` value of `clock.now().millisecondsSinceEpoch`. This is to make the change non-breaking.

## Testing

I have tested and validated the change on Android, and can at least confirm that the notifications to start at the date time now provided to the `periodicallyShow` and `periodicallyShowWithDuration` methods. This results in the desired outcome of the example scenario being presented.

I do not currently have an iOS device, so am unable to confirm that the behaviour is identical. From the testing of @anilslabs mentioned [here](https://github.com/MaikuB/flutter_local_notifications/issues/92#issuecomment-2022566442) it appears that this may not be the case.

### DIFF
--- a/flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
+++ b/flutter_local_notifications/lib/src/flutter_local_notifications_plugin.dart
@@ -399,6 +399,7 @@ class FlutterLocalNotificationsPlugin {
     NotificationDetails notificationDetails, {
     required AndroidScheduleMode androidScheduleMode,
     String? payload,
+    int? repeatStartTime,
   }) async {
     if (kIsWeb) {
       return;
@@ -409,17 +410,22 @@ class FlutterLocalNotificationsPlugin {
           ?.periodicallyShow(id, title, body, repeatInterval,
               notificationDetails: notificationDetails.android,
               payload: payload,
-              scheduleMode: androidScheduleMode);
+              scheduleMode: androidScheduleMode,
+              repeatStartTime: repeatStartTime);
     } else if (defaultTargetPlatform == TargetPlatform.iOS) {
       await resolvePlatformSpecificImplementation<
               IOSFlutterLocalNotificationsPlugin>()
           ?.periodicallyShow(id, title, body, repeatInterval,
-              notificationDetails: notificationDetails.iOS, payload: payload);
+              notificationDetails: notificationDetails.iOS,
+              payload: payload,
+              repeatStartTime: repeatStartTime);
     } else if (defaultTargetPlatform == TargetPlatform.macOS) {
       await resolvePlatformSpecificImplementation<
               MacOSFlutterLocalNotificationsPlugin>()
           ?.periodicallyShow(id, title, body, repeatInterval,
-              notificationDetails: notificationDetails.macOS, payload: payload);
+              notificationDetails: notificationDetails.macOS,
+              payload: payload,
+              repeatStartTime: repeatStartTime);
     } else if (defaultTargetPlatform == TargetPlatform.windows) {
       throw UnsupportedError('Notifications do not repeat on Windows');
     } else {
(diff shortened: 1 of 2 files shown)


BadgerHobbs:
> Interesting! I'm also thinking about re-organizing those methods in general, see #2492. That research led me to realize that _iOS fundamentally doesn't support this feature_. Rather, like with Android, we'd need to work in a background task handler to schedule these tasks at the right time. I believe that's a fine approach, but it does require more code and I don't have an iOS device either :)
> 

Ah, classic iOS not supporting the feature. Sounds like a background task approach would in theory work, not sure how the implementation would be impacted by the handling of background tasks. 

From my experience they are run inconsistently at best, and sometimes not at all.

Would need someone with more experience/expertise to chime in on that one.

Levi-Lesches:
Seems like, really, no platform supports everything:
- iOS/MacOS support repeating notifications _starting now_
- Android, Windows, and Linux do not support repeating notifications at all

On Android, we use background tasks, so we probably need to do so on all platforms: 
- Android: https://developer.android.com/develop/background-work/services/alarms/schedule
- iOS/MacOS: https://developer.apple.com/documentation/uikit/using-background-tasks-to-update-your-app
- Windows: https://learn.microsoft.com/en-us/windows/uwp/launch-resume/create-and-register-a-winmain-background-task
- Linux: Can't find anything remotely native, so maybe spawning a thread and sleeping?

MaikuB:
The feature was never done and requests on this had been closed as Apple doesn't support it like @Levi-Lesches mentioned. 

@BadgerHobbs yep background tasks is an approach and I do remember consistency issues from it being subject to the OS optimisations e.g. around battery. If this is still the case then it'll likely not meet the use case. I've already seen a number of issues in Android with developers using inexact timing and still question why it doesn't fire at the exact timing.

Another thing to note is there is already a work manager plugin that makes use of background tasks. I've seen the community leverage this with this plugin already. My preference would be to keep the separation that way

BadgerHobbs:
> The feature was never done and requests on this had been closed as Apple doesn't support it like @Levi-Lesches mentioned. 
> 
> @BadgerHobbs yep background tasks is an approach and I do remember consistency issues from it being subject to the OS optimisations e.g. around battery. If this is still the case then it'll likely not meet the use case. I've already seen a number of issues in Android with developers using inexact timing and still question why it doesn't fire at the exact timing.
> 
> Another thing to note is there is already a work manager plugin that makes use of background tasks. I've seen the community leverage this with this plugin already. My preference would be to keep the separation that way

Thanks for the reply. Yeah I think this functionality is going to be very difficult to implement reliably on iOS. A different approach might just be to do network based notifications as that may be more reliable (though not something I've investigated).

As it stands my intention is to maintain the current functionality in my fork as it is useful in my app, and I can just disable the functionality for iOS users as it's not super critical.

What's the current approach for asymmetric features across platforms, sounds like it's something just generally avoided?

Levi-Lesches:
> What's the current approach for asymmetric features across platforms, sounds like it's something just generally avoided?

I updated my proposal in #2492 to address this with an API I'm really happy with, please check it out and leave feedback!

Issue #408, opened 2019-12-27
Unable to start receiver com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver


JCKodel:
**Describe the bug**
This is a Crashlytics report, can't pinpoint exactly what's going on.
I have >140k active users and this error affected 11 users.

```
Fatal Exception: java.lang.RuntimeException: Unable to start receiver com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:3623)
       at android.app.ActivityThread.access$1300(ActivityThread.java:240)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1808)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loop(Looper.java:214)
       at android.app.ActivityThread.main(ActivityThread.java:7097)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:975)
```

```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Integer.intValue()' on a null object reference
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.setupNotificationChannel(FlutterLocalNotificationsPlugin.java:584)
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.createNotification(FlutterLocalNotificationsPlugin.java:115)
       at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.showNotification(FlutterLocalNotificationsPlugin.java:634)
       at com.dexterous.flutterlocalnotifications.ScheduledNotificationReceiver.onReceive(ScheduledNotificationReceiver.java:45)
       at android.app.ActivityThread.handleReceiver(ActivityThread.java:3614)
       at android.app.ActivityThread.access$1300(ActivityThread.java:240)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1808)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loop(Looper.java:214)
[...]

PR #2552, opened 2025-02-17
[flutter_local_notification] Restore proguard-rules for v18 and lower

koji-1009:
Restore proguard-rules and doc for versions v18 and lower.

* https://github.com/MaikuB/flutter_local_notifications/pull/2440
* https://github.com/MaikuB/flutter_local_notifications/issues/408#issuecomment-2661744363

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -337,6 +337,12 @@ Developers should also be across Google's requirements on using full-screen inte
 
 ⚠️ Ensure that you have configured the resources that should be kept so that resources like your notification icons aren't discarded by the R8 compiler by following the instructions [here](https://developer.android.com/studio/build/shrink-code#keep-resources). If you have chosen to use `@mipmap/ic_launcher` as the notification icon (against the official Android guidance), be sure to include this in the `keep.xml` file. If you fail to do this, notifications might be broken. In the worst case they will never show, instead silently failing when the system looks for a resource that has been removed. If they do still show, you might not see the icon you specified. The configuration used by the example app can be found [here](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/example/android/app/src/main/res/raw/keep.xml) where it is specifying that all drawable resources should be kept, as well as the file used to play a custom notification sound (sound file is located [here](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/example/android/app/src/main/res/raw/slow_spring_board.mp3)).
 
+#### proguard-rules.pro
+
+For flutter_local_notificaiton v19 and higher, the proguard rules are automatically provided by the GSON. The following documentation is for v18 and lower versions.
+
+Before creating the release build of your app (which is the default setting when building an APK or app bundle) you will need to customise your ProGuard configuration file as per this [link](https://developer.android.com/studio/build/shrink-code#keep-code). Rules specific to the GSON dependency being used by the plugin will need to be added. These rules can be found [here](https://github.com/google/gson/blob/main/examples/android-proguard-example/proguard.cfg). Whilst the example app has a Proguard rules (`proguard-rules.pro`) [here](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/example/android/app/proguard-rules.pro), it is recommended that developers refer to the rules on the GSON repository in case they get updated over time.
+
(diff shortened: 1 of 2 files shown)


flutter_local_notifications/README.md
```diff
@@ -337,6 +337,12 @@ Developers should also be across Google's requirements on using full-screen inte
...
 
+#### proguard-rules.pro
+
+For flutter_local_notificaiton v19 and higher, the proguard rules are automatically provided by the GSON. The following documentation is for v18 and lower versions.
+
+Before creating the release build of your app (which is the default setting when building an APK or app bundle) you will need to customise your ProGuard configuration file as per this [link](https://developer.android.com/studio/build/shrink-code#keep-code). Rules specific to the GSON dependency being used by the plugin will need to be added. These rules can be found [here](https://github.com/google/gson/blob/main/examples/android-proguard-example/proguard.cfg). Whilst the example app has a Proguard rules (`proguard-rules.pro`) [here](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/example/android/app/proguard-rules.pro), it is recommended that developers refer to the rules on the GSON repository in case they get updated over time.
```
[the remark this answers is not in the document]
MaikuB: Yep you're right it's not going to be possible to reference by tag due to the existence of older readme files that pointed to master

flutter_local_notifications/README.md
```diff
@@ -337,6 +337,12 @@ Developers should also be across Google's requirements on using full-screen inte
...
 
+#### proguard-rules.pro
```
MaikuB, line 340: Since another (sub)section has been added. Could you update the TOCs so there's a link that allows jumping to this one? I would also suggest renaming this to `ProGuard rules`

flutter_local_notifications/README.md
```diff
@@ -337,6 +337,12 @@ Developers should also be across Google's requirements on using full-screen inte
...
+
+For flutter_local_notificaiton v19 and higher, the proguard rules are automatically provided by the GSON. The following documentation is for v18 and lower versions.
```
MaikuB, line 342: nit: `proguard rules` -> `ProGuard rules`

MaikuB:
Thanks for the quick turnaround. Left a couple of minor comments

flutter_local_notifications/README.md
koji-1009: Thanks! I fixed spell and added link. [60086c2](https://github.com/MaikuB/flutter_local_notifications/pull/2552/commits/60086c28705ad6b85874a43c6197e84284ff1ccb)

flutter_local_notifications/README.md
koji-1009: Sorry, I fixed. [60086c2](https://github.com/MaikuB/flutter_local_notifications/pull/2552/commits/60086c28705ad6b85874a43c6197e84284ff1ccb)

flutter_local_notifications/example/android/app/proguard-rules.pro
```diff
@@ -0,0 +1,29 @@
+# Note: This file is not required for flutter_local_notificaiton v19 and higher
```
MaikuB, line 1: typo: `flutter_local_notificaiton` -> `flutter_local_notifications`

koji-1009: Sorry,,, [38d7d09](https://github.com/MaikuB/flutter_local_notifications/pull/2552/commits/38d7d09ea34d2e3cd4c400e1da313e3f427c0227)

MaikuB (approved):

Levi-Lesches:
@MaikuB I'm a bit curious here, because the README people see on Pub is linked to the specific version that Pub is showing. Meaning, this restored section will only show for v19, but it says it's not needed for v19. Wouldn't it be better to leave this section out? That way, older versions of the plugin that need it would still have it, but going forward it would be absent.

MaikuB:
@Levi-Lesches that is correct but that assumes the only place developers check is on pub.dev. GitHub is another place that developers check and won't be aware of leveraging tags. Either way there's no harm in mentioning this and this will likely be a temporary thing. There are times that are libraries do something similar and mention when certain changes were introduced so this is similar. An example is https://pub.dev/packages/freezed/versions/1.0.0 that called out when null safety support was dropped or they temporarily include a migration guide

PR #2558, opened 2025-02-23
 [flutter_local_notifications] Add support for `providesAppNotificationSettings`

lukemmtt:
(will add later)


flutter_local_notifications/lib/src/platform_specifics/darwin/initialization_settings.dart
```diff
@@ -51,6 +52,19 @@ class DarwinInitializationSettings {
   /// Default value is 'false'.
   final bool requestCriticalPermission;
 
+  /// Request permission to provide custom notification settings UI.
+  ///
+  /// When enabled, the system displays a "Configure Notifications in {App}"
```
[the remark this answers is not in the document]
lukemmtt: 100%! In my head I think about brackets but <> is definitely more common 🙏🏻

lukemmtt, line 57:
```suggestion
  /// When enabled, the system displays a "Configure Notifications in <application name>"
```

flutter_local_notifications/lib/src/platform_specifics/darwin/initialization_settings.dart
```diff
@@ -51,6 +52,19 @@ class DarwinInitializationSettings {
...
+  /// button in the notification's context menu (when swiping on a
+  /// notification and selecting Options). This allows your app to
+  /// present a custom UI for notification settings by handling the
+  /// appropriate delegate callback.
+  ///
+  /// Corresponds to https://developer.apple.com/documentation/usernotifications/unauthorizationoptions/3040321-providesappnotificationsettings
+  ///
+  /// Default value is false.
```
lukemmtt, line 65:
```suggestion
  /// Corresponds to https://developer.apple.com/documentation/usernotifications/unauthorizationoptions/3040321-providesappnotificationsettings
  ///
  /// On iOS, this property is only applicable to iOS 12 or newer.
  /// On macOS, this property is only applicable to macOS 10.14 or newer.
  ///
  /// Default value is false.
```

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -858,13 +862,15 @@ class MacOSFlutterLocalNotificationsPlugin
     bool badge = false,
     bool provisional = false,
     bool critical = false,
+    bool providesAppNotificationSettings = false,
   }) =>
-      _channel.invokeMethod<bool>('requestPermissions', <String, bool?>{
+      _channel.invokeMethod<bool?>('requestPermissions', <String, bool?>{
```
lukemmtt, line 867:
```suggestion
      _channel.invokeMethod<bool>('requestPermissions', <String, bool?>{
```

In self-review, I see that this might not have been needed.

flutter_local_notifications/test/ios_flutter_local_notifications_test.dart
```diff
@@ -698,27 +701,29 @@ void main() {
...
-            alert: true,
-            provisional: true,
-            critical: true,
-          );
+              sound: true,
+              alert: true,
+              badge: true,
+              provisional: true,
+              critical: true,
+              providesAppNotificationSettings: true);
```
lukemmtt, line 718:
```suggestion
            sound: true,
            badge: true,
            alert: true,
            provisional: true,
            critical: true,
            providesAppNotificationSettings: true);
```

lukemmtt:
> Thanks a lot for the PR.

Sure thing! Glad to contribute 🙏🏻

> The addition of pictures adds a lot and makes reviews a lot more easier so thanks so much for being thoughtful about that.

Of course! I advocate for the same in my own project's PRs—pictures and videos are invaluable!

> I've left some comments that are minor and happy to discuss them further if needed.

Great stuff, thank you for the prompt and thoughtful review.

> Besides that, are you able to update example app so this setting is requested? It looks like it's a safe one to request so I'm thinking it's worth requesting it so that those running the example app will see the effect without making code changes

Sure thing, happy to 🚀 will report back shortly.

Thank you Michael!

flutter_local_notifications/lib/src/platform_specifics/darwin/initialization_settings.dart
MaikuB: Not familiar with suggested changes but lgtm

lukemmtt:
Btw @MaikuB, wasn't sure if you're aware of the handy PR suggestion tool, but I'm a big fan of it, especially for edits to code comments, makes nits really easy. Forgive me if you already know it well—I only found out about it last year after years of PR reviews!

https://github.com/user-attachments/assets/e6b1c940-0f6c-4d47-8455-4abba7afa72b

flutter_local_notifications/lib/src/platform_specifics/darwin/initialization_settings.dart
MaikuB: For consistency can you make the paragraph with the applicable OS versions last?

MaikuB:
No I wasn't aware of it so thanks for the tip  I'll need to look up how that works. Have used it for Azure DevOps but currently responding on phone (and will go to bed soon...) so suspect functionality will be limited. I imagine it'll be similar in highlighting a portion of the code to make a suggestion

flutter_local_notifications/lib/src/platform_specifics/darwin/initialization_settings.dart
lukemmtt: Done (edited pending suggestion above)

lukemmtt:
Glad I could help! In case it's not clear from the video, this button is the key:

image

lukemmtt:
I've added a demo to the iOS Example app. I've not touched the Mac example app, as I'm short on time. Hopefully this is adequate to merge.

https://github.com/user-attachments/assets/fec8f9e1-e2c3-41d3-874a-07b36b936c55

flutter_local_notifications/example/lib/main.dart
```diff
@@ -706,10 +746,9 @@ class _HomePageState extends State<HomePage> {
                       },
                     ),
                   ],
-                  if (!kIsWeb &&
-                      (Platform.isIOS || Platform.isMacOS)) ...<Widget>[
+                  if (Platform.isIOS) ...<Widget>[
                     const Text(
-                      'iOS and macOS-specific examples',
+                      'iOS-specific examples',
```
MaikuB, line 751: Was there a reason for this change? I recall there being some applicable to macOS. I assume you were trying to do this only for adding in `_buildConfigureInAppToggle()` where I've not seen how to bring up the context menu for macOS that this should only be for iOS

flutter_local_notifications/README.md
```diff
@@ -102,6 +102,7 @@ Note: the plugin requires Flutter SDK 3.22 at a minimum. The list of support pla
 * [Android] Ability to check if notifications are enabled
 * [iOS (all supported versions) & macOS 10.14+] Request notification permissions and customise the permissions being requested around displaying notifications
 * [iOS 10 or newer and macOS 10.14 or newer] Display notifications with attachments
+* [iOS 12.0+ and macOS 10.14+] Support for custom notification settings UI via "Configure Notifications in {App}" button in notification context menu
```
MaikuB, line 105:
```suggestion
* [iOS 12.0+ and macOS 10.14+] Support for custom notification settings UI via "Configure Notifications in <application name>" button in notification context menu
```

MaikuB (changes requested):
I've left some comments and suggestions around requested changes. Rest looks good to me

flutter_local_notifications/example/lib/main.dart
lukemmtt: Good catch, this was an accident, I was cleaning up my other edits where I inadvertently referenced Mac in the iOS directory files' comments and must have messed this up

flutter_local_notifications/example/lib/main.dart
```diff
@@ -783,6 +822,10 @@ class _HomePageState extends State<HomePage> {
                         await _showNotificationInNotificationCentreOnly();
                       },
                     ),
+                    _buildConfigureInAppToggle(),
```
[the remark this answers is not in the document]
lukemmtt: Certainly, I only went with this approach because I sought to improve readability without moving it too far away, and it's no worse than putting it in in-line re: performance. 

Happy to revise. I use DCM as well, so I'll follow its guidance for the revisions 👍🏻

flutter_local_notifications/example/lib/main.dart
lukemmtt, line 751:
```suggestion
                      'iOS and macOS-specific examples',
```

lukemmtt:
@MaikuB All PR comments and DCM lints are resolved

flutter_local_notifications/test/macos_flutter_local_notifications_test.dart
```diff
@@ -681,5 +688,95 @@ void main() {
...
+    test('macOS requestPermissions with all settings requested', () async {
+      await flutterLocalNotificationsPlugin
+          .resolvePlatformSpecificImplementation<
+              MacOSFlutterLocalNotificationsPlugin>()!
+          .requestPermissions(
+            sound: true,
+            badge: true,
+            alert: true,
+            provisional: true,
+            critical: true,
```
MaikuB, line 768: I missed this before but since the test description is `macOS requestPermissions with all settings requested`, it looks this is missing requesting passing in `providesAppNotificationSettings` as `true`. iOS test does this but not the macOS one

flutter_local_notifications/test/macos_flutter_local_notifications_test.dart
```diff
@@ -681,5 +688,95 @@ void main() {
...
+            'provisional': false,
+            'critical': false,
+            'providesAppNotificationSettings': true,
+          },
+        ),
+      ]);
+    });
+
+    test(
+      'check notification settings includes providesAppNotificationSettings',
```
MaikuB, line 747: based on the code for this test, this test is redundant as there's a `checkPermissions` test above that

flutter_local_notifications/test/ios_flutter_local_notifications_test.dart
```diff
@@ -760,5 +765,69 @@ void main() {
...
+          'badge': true,
+          'provisional': false,
+          'critical': false,
+          'providesAppNotificationSettings': true
+        })
+      ]);
+    });
+
+    test(
+      'check notification settings includes providesAppNotificationSettings',
```
MaikuB, line 821: this test appears to be redundant as it has the same code as the `checkPermissions` test

flutter_local_notifications/test/flutter_local_notifications_test.dart
```diff
@@ -14,6 +16,44 @@ void main() {
...
+      final NotificationsEnabledOptions options = NotificationsEnabledOptions(
+        isEnabled: true,
+        isSoundEnabled: true,
+        isAlertEnabled: true,
+        isBadgeEnabled: true,
+        isProvisionalEnabled: false,
+        isCriticalEnabled: false,
+        isProvidesAppNotificationSettingsEnabled: true,
+      );
+      expect(options.isProvidesAppNotificationSettingsEnabled, true);
```
MaikuB, line 54: it says it should be executing a permissions check and the test should be asserting what it returns but the test doesn't do this. is this test missing some logic?

MaikuB (changes requested):
Thanks for making the changes. I missed some parts that look odd with the unit tests. The recent changes that were push introduce linter issues as well. Can you take a look? Note that if you check the issues locally, the prerelease version of the plugin is set to have the minimum Flutter SDK as 3.22 so configuring that locally can show the relevant linter issues in your IDE

MaikuB:
Hey @lukemmtt, checking back on to see if you're able to look at the PR comments.

flutter_local_notifications/test/flutter_local_notifications_test.dart
lukemmtt: Thanks for the catch @MaikuB; this was a mistake in naming. To address this, I've renamed (https://github.com/MaikuB/flutter_local_notifications/pull/2558/commits/fe17dadf18df38aa43f338fb9eefe49d8c070d25) this test from:
`permission check returns correct value`
to
`NotificationsEnabledOptions stores providesAppNotificationSettings value`.

The actual `checkPermissions()` method is tested in the platform-specific test files.

flutter_local_notifications/test/ios_flutter_local_notifications_test.dart
lukemmtt: Thank you, I've removed it from both platform test files here https://github.com/MaikuB/flutter_local_notifications/pull/2558/commits/f82c4ab8351f791dc5c6d03466d25a3c86f78a55

flutter_local_notifications/test/macos_flutter_local_notifications_test.dart
lukemmtt: Good catch; fixed here https://github.com/MaikuB/flutter_local_notifications/pull/2558/commits/65d17160d85fb64441c77e7520c2e2e814b3233d

flutter_local_notifications/test/macos_flutter_local_notifications_test.dart
lukemmtt: Thank you; removed the redundant tests here: https://github.com/MaikuB/flutter_local_notifications/pull/2558/commits/f82c4ab8351f791dc5c6d03466d25a3c86f78a55

lukemmtt:
Thanks for your patience @MaikuB and thank you for taking the time to review my PR in the first place, I appreciate it!

I've merged in `master`, resolved conflicts, and resolved the issue you pointed out as best as I could. I also resolved the lints and ran dart format against the Flutter 3.22.0 SDK. I don't have a DCM subscription at the moment, but let me know there are any lingering issues and I'll be happy to address. 

Thank you!

flutter_local_notifications/example/lib/main.dart
```diff
@@ -760,8 +784,7 @@ class _HomePageState extends State<HomePage> {
                       },
                     ),
                   ],
-                  if (!kIsWeb &&
-                      (Platform.isIOS || Platform.isMacOS)) ...<Widget>[
+                  if (Platform.isIOS) ...<Widget>[
```
MaikuB:
```suggestion
                  if (!kIsWeb &&
                      (Platform.isIOS || Platform.isMacOS)) ...<Widget>[
```

Change should be reverted as it is hiding the examples that use to be visible on macOS. `!kIsWeb` prevents example app from misbehaving on macOS as well

flutter_local_notifications/example/lib/main.dart
```diff
@@ -848,6 +871,13 @@ class _HomePageState extends State<HomePage> {
                         await _showNotificationInNotificationCentreOnly();
                       },
                     ),
+                    ConfigureInAppToggle(
```
MaikuB: The `if (Platform.isIOS)` check should be applied around here so that the toggle and sized box are only shown on iOS

lukemmtt:
> Thanks for making those changes. I picked up a problem in the example app. Can you take a look?

Thanks for the latest review @MaikuB; I've [committed](https://github.com/MaikuB/flutter_local_notifications/pull/2558/commits/0cf6b5ba14f4f1fab74cb6943be7e19d46a57a1e) your [suggested edit](https://github.com/MaikuB/flutter_local_notifications/pull/2558#discussion_r2423406569); I'll test the example app manually as well.

> In the interest of time and to help out, I thought I could make the changes myself. However, I believe you haven't enabled the setting that would've allowed me to push the changes myself.

Thanks for bringing this to my attention—I didn't even realize that such a setting existed. I looked into it, and found out why I can't see this setting—it [isn't an option for organization-owned repositories](https://github.com/isaacs/github/issues/1681), and I happened to create this fork in my own org, so I'm out of luck, even though I'm the sole owner of the org. Bummer. I'll make my forks in my personal repo from now on. 

> FYI for future reference regarding linter issues, these do not need a DCM subscription. I don't have one myself and linter issues will show up with your IDE. I picked these up by opening your PR on VS code

Also great to know! I used to have a DCM subscription myself but cancelled it for cost reasons; I've never tried using it without a sub. I appreciate the insight and will give this a try. 

Thanks and I'll let you know when it's ready for re-review shortly 👍🏻

lukemmtt:
Hey @MaikuB,

I've implemented the edits you've requested.  I haven't yet tested or fixed any possible DCM lints—having trouble getting it set up, it's complaining about my expired license.

Additionally, one other thing to note: in my latest iteration, I attempted to verify this feature actually works on macOS (since it has the API support, but I had omitted it from the example app before). In my attempt, in which I temporarily added support to the macOS AppDelegate and the example app, the UI button **did not seem to actually appear in notification context menu on macOS upon right-clicking**. I also couldn't find any examples online of this feature working on macOS either.

This seems to suggest that while the macOS API supports the `providesAppNotificationSettings` API, the feature may not actually implemented on the OS level for macOS. 

Commit links:
- [d2e06fa](https://github.com/MaikuB/flutter_local_notifications/pull/2558/commits/d2e06fa807fc517e940c3ab76fb77e213e510a9a): macOS POC attempt with full AppDelegate implementation (documents it doesn't work)
- [0dab96c](https://github.com/MaikuB/flutter_local_notifications/pull/2558/commits/0dab96c720cb11368a7554ccccd14416b50566d0): Revert of POC
- [9a2b243](https://github.com/MaikuB/flutter_local_notifications/pull/2558/commits/9a2b243f865874fef6f2e09e1ac90be77c727f09): Updated API documentation to reflect findings

I've left the macOS API support in place in this PR (as it was before), mostly for completeness, and also in case the OS changes to actually implement the UI in the future (or someone finds evidence to the contrary that suggests that macOS DOES support this in its UI). Let me know if you'd prefer to strip it out completely.

Thanks!

MaikuB:
Hmm I merged this but I just thought of a question. Would it have been possible for the plugin to handle the [method](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:opensettingsfor:)?language=objc) instead of requiring developers to implement it in their own `AppDelegate`? I believe it would have been possible given the plugin handles other ones but didn't get to test in this context. If it's possible, this is what comes to mind and let me know what you think
- [This enum](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/NotificationResponseType.html) is updated to support an additional type that is iOS-specific and could be called something like `selectedShowAppNotificationSettings` specific to the scenario in this PR
- Presumably picking the option in the context menu can launch the app. If so, plugin users could make use [this method](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/FlutterLocalNotificationsPlugin/getNotificationAppLaunchDetails.html) to know that the app was launched by that option. The new response type should become in the object returned and along with the notification details
- Similarly, if the app is running,  it should fire the `onDidReceiveNotificationResponse` callback that is part of the [initialisation](https://pub.dev/documentation/flutter_local_notifications/latest/flutter_local_notifications/AndroidFlutterLocalNotificationsPlugin/initialize.html) method

lukemmtt:
Thanks for the merge @MaikuB! I'm grateful for your diligence in reviewing my PR and maintaining this plugin, and for your offering the plugin to the community in the first place!

As for your question about whether the plugin could have handled the method instead of requiring devs to implement it on their own, it's a really thoughtful idea! I'd much prefer something like that myself. Unfortunately though, I looked into it, and it seems that it might not be possible at all, due to a Flutter engine limitation.

Specifically, Flutter's `addApplicationDelegate` forwarding system only forwards **two specific delegate methods**:

1. `userNotificationCenter:willPresentNotification:withCompletionHandler:`
2. `userNotificationCenter:didReceiveNotificationResponse:withCompletionHandler:`

`openSettingsFor` is **not forwarded**, so plugins registered via `addApplicationDelegate` (like flutter_local_notifications) would never receive this call.

See: [FlutterAppDelegate.mm](https://github.com/flutter/engine/blob/main/shell/platform/darwin/ios/framework/Source/FlutterAppDelegate.mm) and [FlutterPluginRegistrar Protocol](https://api.flutter.dev/ios-embedder/protocol_flutter_plugin_registrar-p.html)

As I understand it, that makes such an implementation on iOS impossible, at least without Flutter engine changes. macOS of course doesn't use this forwarding system, so we could theoretically implement some built-in support for this feature on macOS, but that's pointless of course, since macOS doesn't seem to have a UI implementation for this capability to begin with.

Let me know if this adds up; this is all new to me and I'm certainly no expert in this domain, I simply did some digging.

Thank you!

MaikuB:
Thanks for the response and references. Yep I took a look and you're right. What you shared would indicate Flutter isn't forwarding the calls. If it was then there should be some code that invokes the method like it does for the ones you highlighted

PR #2562, opened 2025-02-25
[flutter_local_notifications] Support bypassing DnD-settings on Android

### LINKED ISSUE #1211: [Feature] Add setBypassDnd option to Android Notificationchannel creation
Hello,
I like flutter_local_notifications much, but there is one feature I need and is missing.
The feature I need is the option to set "setBypassDnd" by creating a Android Notificaitonchannel.

If I create a notification channel in java, I can set the option and after allowing this feature in notificaiton settings of Android, Notifications of this channel are also loud if dnd is on.

example code:
```
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
            NotificationManager notificationManager = getSystemService(NotificationManager.class);

            String CHANNEL_ID = "alarm";
            CharSequence name = "alarm";
            String description = "alarm";

            int importance = NotificationManager.IMPORTANCE_HIGH;
            NotificationChannel notificationChannel = new NotificationChannel(CHANNEL_ID, name, importance);
            notificationChannel.setDescription(description);

            notificationChannel.setBypassDnd(true);

            if (notificationManager != null) {
                notificationManager.createNotificationChannel(notificationChannel);
            }
        }
```

Through adding this feature I don't need duplicate code anymode.

Thank you, for your good work :)

mvarendorff2:
Closes #1211 

- Adds new `bypassDnd` and `channelBypassDnd` options for `AndroidNotificationChannel` and `AndroidNotificationDetails` respectively
- Adds `hasNotificationPolicyAccess()` and `requestNotificationPolicyAccess()` methods to `AndroidFlutterLocalNotificationsPlugin`
- Adds relevant buttons to example
- Adds README-entry

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -431,11 +431,11 @@ Developers should also be across Google's requirements on using full-screen inte
 
 ### Bypassing Do Not Disturb (DnD)
 
-If your application needs the ability to send notifications that ignore the device's DnD mode settings, you must request these permissions from Android 6.0 onwards. You can do this using `AndroidFlutterNotificationsPlugin` by calling the `requestNotificationPolicyAccess()` method which will redirect the user to a settings page where your application may be explicitly whitelisted to bypass DnD.
+If your application needs the ability to send notifications that ignore the device's DnD mode settings, you must request these permissions from Android 6.0 onwards. You can do this using `AndroidFlutterNotificationsPlugin` by calling the `requestNotificationPolicyAccess()` method which will redirect the user to a settings page where your application may be explicitly whitelisted to bypass DnD. **This method _must_ be called before attempting to create a notification channel with `bypassDnd: true`.** Failing to do so will cause the `bypassDnd` argument to be treated as `false`.
 
 For notifications to then actually ignore the DnD-status of a device, the channel must be created with `bypassDnd: true`, or the first notification on a channel that creates it must be sent with `channelBypassDnd: true`.
 
-**NOTE:** This does _not_ ignore the device's silent mode! Should you have a use case where you must notify your users (for instance in an emergency), you might want to use a package like [`sound_mode`](https://pub.dev/packages/sound_mode) or write your own platform specific code to set the `RingerMode` of the device as well as change the notification stream's volume before and after the notification.
+**NOTE:** This does _not_ ignore the device's silent mode! Should you have a use case where you must notify your users (for instance in an emergency), you might want to use a package like [`sound_mode`](https://pub.dev/packages/sound_mode) or write your own platform-specific code to set the `RingerMode` of the device as well as change the notification stream's volume before and after the notification.
 
 ### Release build configuration
 
@@ -602,7 +602,7 @@ Developers should also note that whilst accessing plugins will work, on Android 
 
 **Specifying actions on notifications**:
 
-The notification actions are platform specific and you have to specify them differently for each platform.
+The notification actions are platform-specific and you have to specify them differently for each platform.
 
 On iOS/macOS, the actions are defined on a category, please see the configuration section for details.
 
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -386,6 +386,7 @@ For apps that need the following functionality please complete the following in 
             android:stopWithTask="false"
             android:foregroundServiceType="<foreground service types>">
     ```
+* To be able to create channels that ignore the user's Do-Not-Disturb settings, specify the `<uses-permission android:name="android.permission.ACCESS_NOTIFICATION_POLICY" />` permission between the `<manifest>` tags. Developers will also need to follow the instructions documented [here](#bypassing-dnd)
 
 Developers can refer to the example app's `AndroidManifest.xml` to help see what the end result may look like. Do note that the example app covers all the plugin's supported functionality so will request more permissions than your own app may need
 
@@ -428,13 +429,21 @@ Note that when a full-screen intent notification actually occurs (as opposed to 
 
 Developers should also be across Google's requirements on using full-screen intents. Please refer to their documentation [here](https://source.android.com/docs/core/permissions/fsi-limits) for more information. Should you app need request permissions, the `AndroidFlutterNotificationsPlugin` class exposes the `requestFullScreenIntentPermission()` method that can be used to do so.
 
+### Bypassing dnd
+
+If your application needs the ability to send notifications that ignore the device's current Do-Not-Disturb settings, you must request these permissions from Android 6.0 onwards. You can do this using `AndroidFlutterNotificationsPlugin` by calling the `requestNotificationPolicyAccess()` method which will redirect the user to a settings page where your application may be explicitly whitelisted to bypass DnD.
+
+For notifications to then actually ignore the DnD-status of a device, the channel must be created with `bypassDnd: true`, or the first notification on a channel that creates it must be sent with `channelBypassDnd: true`.
+
+**NOTE:** This does _not_ ignore the device's silent mode! Should you have a use-case where you must notify your users (for instance in an emergency), you might want to use a package like [`sound_mode`](https://pub.dev/packages/sound_mode) or write your own platform specific code to set the `RingerMode` of the device as well as change the notification stream's volume before and after the notification.
+
 ### Release build configuration
 
 ⚠️ Ensure that you have configured the resources that should be kept so that resources like your notification icons aren't discarded by the R8 compiler by following the instructions [here](https://developer.android.com/studio/build/shrink-code#keep-resources). If you have chosen to use `@mipmap/ic_launcher` as the notification icon (against the official Android guidance), be sure to include this in the `keep.xml` file. If you fail to do this, notifications might be broken. In the worst case they will never show, instead silently failing when the system looks for a resource that has been removed. If they do still show, you might not see the icon you specified. The configuration used by the example app can be found [here](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/example/android/app/src/main/res/raw/keep.xml) where it is specifying that all drawable resources should be kept, as well as the file used to play a custom notification sound (sound file is located [here](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/example/android/app/src/main/res/raw/slow_spring_board.mp3)).
 
 #### ProGuard rules
 
-For flutter_local_notificaiton v19 and higher, the ProGuard rules are automatically provided by the GSON. The following documentation is for v18 and lower versions.
+For flutter_local_notification v19 and higher, the ProGuard rules are automatically provided by the GSON. The following documentation is for v18 and lower versions.
 
 Before creating the release build of your app (which is the default setting when building an APK or app bundle) you will need to customise your ProGuard configuration file as per this [link](https://developer.android.com/studio/build/shrink-code#keep-code). Rules specific to the GSON dependency being used by the plugin will need to be added. These rules can be found [here](https://github.com/google/gson/blob/main/examples/android-proguard-example/proguard.cfg). Whilst the example app has a Proguard rules (`proguard-rules.pro`) [here](https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/example/android/app/proguard-rules.pro), it is recommended that developers refer to the rules on the GSON repository in case they get updated over time.
 
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -435,7 +435,7 @@ If your application needs the ability to send notifications that ignore the devi
 
 For notifications to then actually ignore the DnD-status of a device, the channel must be created with `bypassDnd: true`, or the first notification on a channel that creates it must be sent with `channelBypassDnd: true`.
 
-**NOTE:** This does _not_ ignore the device's silent mode! Should you have a use-case where you must notify your users (for instance in an emergency), you might want to use a package like [`sound_mode`](https://pub.dev/packages/sound_mode) or write your own platform specific code to set the `RingerMode` of the device as well as change the notification stream's volume before and after the notification.
+**NOTE:** This does _not_ ignore the device's silent mode! Should you have a use case where you must notify your users (for instance in an emergency), you might want to use a package like [`sound_mode`](https://pub.dev/packages/sound_mode) or write your own platform specific code to set the `RingerMode` of the device as well as change the notification stream's volume before and after the notification.
 
 ### Release build configuration
 
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -386,7 +386,7 @@ For apps that need the following functionality please complete the following in 
             android:stopWithTask="false"
             android:foregroundServiceType="<foreground service types>">
     ```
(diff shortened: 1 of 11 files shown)


Levi-Lesches:
@mvarendorff2 Have you tested this and does it work? I noticed you added the parameters to Java and Dart, but I didn't see where you pass it between the two in method channels. Specifically, 

https://github.com/MaikuB/flutter_local_notifications/blob/8839a00cdfcdecc5090da5ccb766da42cdbe2a88/flutter_local_notifications/lib/src/platform_specifics/android/method_channel_mappers.dart#L40-L50

and 

https://github.com/MaikuB/flutter_local_notifications/blob/8839a00cdfcdecc5090da5ccb766da42cdbe2a88/flutter_local_notifications/lib/src/platform_specifics/android/method_channel_mappers.dart#L174-L181

Levi-Lesches:
Also I realize this is just a draft but I'm not sure how exactly this could work. You could set the importance of the channel to [IMPORTANCE_MAX](https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#IMPORTANCE_MAX()), which are

> ...allowed to makes noise, peek, and use full screen intents.

but it doesn't explicitly specify bypassing do not disturb. Moreover, this is already configurable in the plugin today by setting: 

```dart
final androidPlugin = plugin.resolvePlatformSpecificImplementation<AndroidFlutterLocalNotificationsPlugin>();
const channel = AndroidNotificationChannel(id: "myId", name: "myName", importance: Importance.max);
await androidPlugin?.createNotificationChannel(channel);
```

----

You might be able to use [`setBypassDnd()`](https://developer.android.com/reference/android/app/NotificationChannel#setBypassDnd(boolean)), which seems like it's the route you're trying to take, but note that

> this channel can interrupt the user in [NotificationManager.INTERRUPTION_FILTER_PRIORITY](https://developer.android.com/reference/android/app/NotificationManager#INTERRUPTION_FILTER_PRIORITY) mode.
> 
> Apps with Do Not Disturb policy access (see [NotificationManager.isNotificationPolicyAccessGranted()](https://developer.android.com/reference/android/app/NotificationManager#isNotificationPolicyAccessGranted())) can set up their own channels this way, but only if the channel hasn't been updated by the user since its creation
> 
> Otherwise, this value is only modifiable by the system and the notification ranker.

So I'm not sure apps would even be able to call this function, and even then it wouldn't help against [`INTERRUPTION_FILTER_NONE`](https://developer.android.com/reference/android/app/NotificationManager#INTERRUPTION_FILTER_NONE)

mvarendorff2:
@Levi-Lesches thanks a bunch for these insights! This is indeed just a draft - I am still fiddling around with some things trying to get it to work locally, so this is just not complete yet (including the missing mappings you mentioned which I only found a bit after creating this PR).

 `setBypassDnd` is exactly what I intend to use - my (possibly limited) understanding is, that a user may exempt apps from DnD restrictions in their system settings (as you've said) which I have done with the example app and that's generally fine with me.

Maybe some context as to what I am trying to achieve / support here: I am working on a project that should feature critical (push-) notifications for emergency situations. Getting this to work with Apple was shockingly straightforward (applying to the managed Entitlement for critical notifications and then just sending `{sound: {critical: true}}` in the APNS notification payload which completely ignores any DnD / silence settings on the iPhone) however on Android it appears we'd have to take the additional roundtrip through a custom notification channel with max importance and whatnot. So far my understanding has been that to achieve similar functionality, I'd want to have a notification channel with `importance: Importance.max, playSound: true` which has also been `setBypassDnd(true)`.

If I am on the completely wrong track with this approach, please let me know - I am not super deep into the whole notification thing yet but this is what was suggested to me as the closest Android equivalent to Apple's critical notifications.

Levi-Lesches:
> setBypassDnd is exactly what I intend to use

Make sure to read the docs in my previous comment -- not just any app can actually call this function. You'll need to provide methods to prompt the user for this special permission, then another method to check for this permission, and only THEN can a call to this function actually succeed

mvarendorff2:
@Levi-Lesches just in case you're curious - this is now tested and functional! We are using a slightly modified fork of this in one of our apps (modified to force notification sound at full volume because of our emergency alarm use-case) great success :)

mvarendorff2:
@JigishDalal it has quite evidently not been merged yet - you can use the fork by using a git reference in your pubspec.yaml:

```yaml
dependencies:
  flutter_local_notifications:
    git:
      url: https://github.com/mvarendorff2/flutter_local_notifications
      ref: feat/bypass-dnd-android
      path: flutter_local_notifications
```

I cannot guarantee to keep the fork up to date permanently however.

mvarendorff2:
@MaikuB The unit tests are definitely on my list to fix now; one question regarding the two Windows Build checks: Can you confirm that they are not related to the changes I made?

MaikuB:
@mvarendorff2 will need to find out the cause but based on what I've seen in the PR, it doesn't look it'll be related to the changes done

flutter_local_notifications/README.md
```diff
@@ -428,13 +429,21 @@ Note that when a full-screen intent notification actually occurs (as opposed to
 
 Developers should also be across Google's requirements on using full-screen intents. Please refer to their documentation [here](https://source.android.com/docs/core/permissions/fsi-limits) for more information. Should you app need request permissions, the `AndroidFlutterNotificationsPlugin` class exposes the `requestFullScreenIntentPermission()` method that can be used to do so.
 
+### Bypassing dnd
```
MaikuB, line 432:
```suggestion
### Bypassing Do Not Disturb (DnD)
```

Suggested to make it easier to find as this will show up in the table of contents and explain the acronym. Android docs also refer to it without hyphens e.g. https://developer.android.com/develop/ui/views/notifications

flutter_local_notifications/README.md
```diff
@@ -428,13 +429,21 @@ Note that when a full-screen intent notification actually occurs (as opposed to
...
+
+If your application needs the ability to send notifications that ignore the device's current Do-Not-Disturb settings, you must request these permissions from Android 6.0 onwards. You can do this using `AndroidFlutterNotificationsPlugin` by calling the `requestNotificationPolicyAccess()` method which will redirect the user to a settings page where your application may be explicitly whitelisted to bypass DnD.
```
MaikuB, line 434:
```suggestion
If your application needs the ability to send notifications that ignore the device's DnD mode, you must request these permissions from Android 6.0 onwards. You can do this using `AndroidFlutterNotificationsPlugin` by calling the `requestNotificationPolicyAccess()` method which will redirect the user to a settings page where your application may be explicitly whitelisted to bypass DnD.
```

- Uses the acronym
- settings -> mode . Using "mode" is more appropriate to describe that it's how the device is currently behaviour as its enter a "mode". It's also aligned with how Android describes it. Likely because settings is less clear as settings for DnD can be something set in advance and thereby makes using "current" redundant. An example of this is on setting up the bedtime schedule

flutter_local_notifications/README.md
```diff
@@ -428,13 +429,21 @@ Note that when a full-screen intent notification actually occurs (as opposed to
...
 
 #### ProGuard rules
 
-For flutter_local_notificaiton v19 and higher, the ProGuard rules are automatically provided by the GSON. The following documentation is for v18 and lower versions.
+For flutter_local_notification v19 and higher, the ProGuard rules are automatically provided by the GSON. The following documentation is for v18 and lower versions.
```
MaikuB, line 446:
```suggestion
For flutter_local_notifications v19 and higher, the ProGuard rules are automatically provided by the GSON. The following documentation is for v18 and lower versions.
```

Thanks for spotting the typo. Added a missing `s`

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -1888,6 +1919,37 @@ public void requestFullScreenIntentPermission(@NonNull PermissionRequestListener
     }
   }
 
+  @RequiresApi(api = VERSION_CODES.M)
```
MaikuB, line 1922: The `@RequiresApi` only functions as an annotation to help warn those calling the method, which would be the plugin itself. What the method would need is a version check. Currently when either `requestNotificationPolicyAccess()` or `hasNotificationPolicyAccess()` is called on a lower Android version, a runtime exception would occur. To get around this I would suggest to

- update the code within both methods to check the Android version so that if they're on a lower Android version that a `false` boolean result will make it over to the Dart/Flutter side through either  `this.callback.complete(false)` or `result.success(false)`
- remove the `@RequiresApi` annotation on both methods as they're effectively redundant

flutter_local_notifications/lib/src/platform_flutter_local_notifications.dart
```diff
@@ -188,6 +188,26 @@ class AndroidFlutterLocalNotificationsPlugin
   Future<bool?> requestNotificationsPermission() async =>
       _channel.invokeMethod<bool>('requestNotificationsPermission');
 
+  /// Requests access to notification policy. Returns whether the
```
MaikuB, line 191: Can you separate the sentences? See https://dart.dev/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -176,6 +177,10 @@ class AndroidNotificationDetails {
   /// The importance of the notification.
   final Importance importance;
 
+  /// Whether the notification channel should attempt to bypass Do Not Disturb
+  /// settings
```
MaikuB, line 181:
```suggestion
  /// settings.
```

MaikuB (changes requested):
Left some comments on PR. Some are on documentation but main thing to call out is that the PR could do adding logic to check the Android version of the device and act appropriately

mvarendorff2:
@MaikuB Thanks for the review! Everything should be cleaned up now 🤞 And good thing the tests were there, they caught a missing mapping that I somehow overlooked :)

mvarendorff2:
Running this in release mode in one of our projects indicates some issues - the settings don't open at all when calling `requestNotificationPolicyAccess()`. I'm looking into it

flutter_local_notifications/README.md
```diff
@@ -428,13 +429,21 @@ Note that when a full-screen intent notification actually occurs (as opposed to
...
 
+### Bypassing Do Not Disturb (DnD)
+
+If your application needs the ability to send notifications that ignore the device's DnD mode settings, you must request these permissions from Android 6.0 onwards. You can do this using `AndroidFlutterNotificationsPlugin` by calling the `requestNotificationPolicyAccess()` method which will redirect the user to a settings page where your application may be explicitly whitelisted to bypass DnD.
+
+For notifications to then actually ignore the DnD-status of a device, the channel must be created with `bypassDnd: true`, or the first notification on a channel that creates it must be sent with `channelBypassDnd: true`.
+
+**NOTE:** This does _not_ ignore the device's silent mode! Should you have a use-case where you must notify your users (for instance in an emergency), you might want to use a package like [`sound_mode`](https://pub.dev/packages/sound_mode) or write your own platform specific code to set the `RingerMode` of the device as well as change the notification stream's volume before and after the notification.
```
MaikuB, line 438:
```suggestion
**NOTE:** This does _not_ ignore the device's silent mode! Should you have a use case where you must notify your users (for instance in an emergency), you might want to use a package like [`sound_mode`](https://pub.dev/packages/sound_mode) or write your own platform specific code to set the `RingerMode` of the device as well as change the notification stream's volume before and after the notification.
```

use-case -> use case (hyphen isn't needed)

flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
MaikuB: Thanks for adding the missing `.` for other API docs as well :)

MaikuB:
Thanks for making the changes. I left one minor comment that's to do with the readme and also merged in https://github.com/MaikuB/flutter_local_notifications/pull/2608. The latter will fix the issues to do with the Windows side of the plugin. Would you be able to update your branch and fork so that it includes the fix? Whilst your PR doesn't affect the Windows side, would help ensure the pipeline is run with those changes included

mvarendorff2:
@MaikuB yes, definitely! I rebased my branch and fixed the typo :)

flutter_local_notifications/example/lib/main.dart
```diff
@@ -2347,6 +2371,57 @@ class _HomePageState extends State<HomePage> {
...
+            bypassDnd: true,
+            importance: Importance.max);
+
+    final AndroidFlutterLocalNotificationsPlugin? androidPlugin =
+        flutterLocalNotificationsPlugin.resolvePlatformSpecificImplementation<
+            AndroidFlutterLocalNotificationsPlugin>();
+
+    final bool? hasPolicyAccess =
+        await androidPlugin?.hasNotificationPolicyAccess();
+    if (hasPolicyAccess == true) {
```
MaikuB, line 2387:
```suggestion
    if (hasPolicyAccess ?? false) {
```

Spotted a logic issue as I was testing more thoroughly. This is meant to direct users to settings if they hadn't granted access, which means `hasNotificationPolicyAccess()` would actually return `false`. Unless this is fixed, those running the example for the first time will actually never be able to create a channel that bypass DnD

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -1215,6 +1219,16 @@ private static void setupNotificationChannel(
       } else {
         notificationChannel.setSound(null, null);
       }
+
+      if (BooleanUtils.getValue(notificationChannelDetails.bypassDnd)) {
+        notificationChannel.setBypassDnd(true);
```
MaikuB, line 1224: Whilst there's currently no harm with how it's working now, upon reading more of the [docs](https://developer.android.com/reference/android/app/NotificationChannel#setBypassDnd(boolean)) and more thoroughly testing the changes that then led to [this comment](https://github.com/MaikuB/flutter_local_notifications/pull/2562/files#r2077382116), I believe it'll more correct to guard calling `notificationChannel.setBypassDnd(true)` after checking the results of `notificationManager.isNotificationPolicyAccessGranted()`.  Logging the warning could still be done if access hasn't been granted. Let me know what you think. The intent here is be more defensive given that there's a variety of Android devices that could behave differently. Android could potentially also introduce a breaking change that calling `notificationChannel.setBypassDnd(true)` could throw an exception unless appropriate access is given. With that in mind, adding the defensive code now could add peace of mind by making these changes upfront

Hackmodford:
In case it wasn't entirely obvious, I just realized that you have to have the permission to override dnd before you can create the channel that bypasses dnd.

Hackmodford:
> @Hackmodford Yes, that is mentioned both in the docs and a warning is printed to console if you try to do it without the permissions. In case the docs added in this PR are not clear enough, I am definitely open to suggestions for improvement!

I'd consider adding a note that if the channel is created before the app has permission it will not override the dnd setting. This bit me because my existing code was creating the channels too early. To address the issue I added a function to remove the existing channel and replace it if the user wanted this functionality.

mvarendorff2:
@Hackmodford good call, thanks! I added mention of this requirement to the Readme as well as the Dart Doc for `requestNotificationPolicyAccess`, `bypassDnd` on `AndroidNotificationChannel` and `channelBypassDnd` on `AndroidNotificationDetails`.

MaikuB (approved):

Issue #528, opened 2020-03-14
Question: Volume settings and iOS critical alerts


Finn0811:
Hi!

First of all, thank you for your work on this addon, it really helps me a lot!

I'd like to ask two questions:
1. Is it possible to modify the volume used to play the notification sound? I'd like the user to be able to define a volume level, which will overwrite the current level.

2. Is it possible to send [critical alert notifications](https://developer.apple.com/documentation/usernotifications/unnotificationsettings/2963116-criticalalertsetting) on iOS? I currently don't see an option for that so I'd really like to see it in a futre version if possible.

MaikuB:
1. As far as I know, this isn't possible
2. Nothing has been implemented to allow for this. The code is open-source so you can contribute in adding features that you need

Finn0811:
Unfortunately I have no experience with native code, yet.
Maybe someone else can do it, otherwise I might try to look into it at some time.

MaikuB:
For what it's worth, I didn't either before this project and I only work on this in my spare time

Finn0811:
The main purpose of my app is to alert people using push notifications. Critical alerts on iOS are very handy because it overwrites the silent mode etc. :)

MaikuB:
You mention push notifications and that's the scenario critical alerts are used for but this plugin is for local notifications. It's more appropriate to be looking at the plugin you're using for push notifications (assuming you're using one) to request those permissions

Finn0811:
Currently I only transfer data messages and build notifications using this plugin because it's the only way to get background notifications working. 

My idea was to send an argument like "critical: true" and build a critical notification using this plugin if the user enabled that feature in their settings. If its disabled it'll just build a normal notification (already working).

MaikuB:
I assume you read that critical alerts require a special entitlement from Apple

Finn0811:
Yes, that should be no problem

MaikuB:
Closing due to lack of interest and updates. Please submit a PR if you need this functionality

Finn0811:
Still no macOS environment to work on this but for sure in the future ;)

systemtechnikfey:
> Still no macOS environment to work on this but for sure in the future ;)

Hi @Finn0811 ,
were you successful in implementing critical alerts?

Ted-Zhang:
> > Still no macOS environment to work on this but for sure in the future ;)
> 
> Hi @Finn0811 ,
> were you successful in implementing critical alerts?

[...]

PR #2589, opened 2025-03-31
[flutter_local_notifications] Add setting for ios critical alert volume

bannzai:
## Description

This PR adds support for Critical Alert notifications on iOS/macOS platforms. Critical Alert notifications are a special type of notification that can break through Do Not Disturb and silent mode settings, requiring specific entitlements from Apple.

Reference: https://developer.apple.com/documentation/usernotifications/unnotificationsound/criticalsoundnamed(_:withaudiovolume:)
Related issue: https://github.com/MaikuB/flutter_local_notifications/issues/528

## Diff
- Add settings about Critical Alert volume. Ability to set critical sound volume (between 0.0 and 1.0)

### Example
- New example demonstrating Critical Alert functionality
- Added necessary entitlements file for debug builds
  * I tried to add it to the release build too, but the Critical Alert entitlement cannot actually be added without applying to Apple. For verification purposes, I’ve added a DEBUG-specific .entitlements file.

## Checklist

- [x] I have read the [CONTRIBUTING](../CONTRIBUTING.md) document
- [x] My code follows the code style of this project
- [x] I have added tests to cover my changes
- [x] Tests passed
- [x] I have updated the example.

## Breaking Changes
None

## Platform Specific Notes

This feature is only available on:
- iOS 12.0 or newer
- macOS 10.14 or newer

Note: Critical Alert capability requires special entitlement approval from Apple through their developer portal.

### DIFF
--- a/flutter_local_notifications/example/lib/main.dart
+++ b/flutter_local_notifications/example/lib/main.dart
@@ -2733,8 +2733,10 @@ class _HomePageState extends State<HomePage> {
   Future<void> _showNotificationWithCriticalSound() async {
     const DarwinNotificationDetails darwinNotificationDetails =
         DarwinNotificationDetails(
-      criticalSoundVolume: 0.2,
-      interruptionLevel: InterruptionLevel.critical,
+      // Between 0.0 and 1.0
+      criticalSoundVolume: 0.5,
+      // If sound is not specified, the default sound will be used
+      sound: 'slow_spring_board.aiff',
     );
     const NotificationDetails notificationDetails = NotificationDetails(
       iOS: darwinNotificationDetails,
--- a/flutter_local_notifications/example/lib/main.dart
+++ b/flutter_local_notifications/example/lib/main.dart
@@ -745,7 +745,7 @@ class _HomePageState extends State<HomePage> {
                     ),
                     PaddedElevatedButton(
                       buttonText:
-                          'Request permission with Critical Alert Permission',
+                          'Request permission with critical alert permission',
                       onPressed: _requestPermissionsWithCriticalAlert,
                     ),
                     PaddedElevatedButton(
--- a/flutter_local_notifications/example/lib/main.dart
+++ b/flutter_local_notifications/example/lib/main.dart
@@ -273,6 +273,29 @@ class _HomePageState extends State<HomePage> {
     }
   }
 
+  Future<void> _requestPermissionsWithCriticalAlert() async {
+    if (Platform.isIOS || Platform.isMacOS) {
+      await flutterLocalNotificationsPlugin
+          .resolvePlatformSpecificImplementation<
+              IOSFlutterLocalNotificationsPlugin>()
+          ?.requestPermissions(
+            alert: true,
+            badge: true,
+            sound: true,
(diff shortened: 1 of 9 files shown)


flutter_local_notifications/lib/src/platform_specifics/darwin/notification_details.dart
```diff
@@ -139,4 +140,15 @@ class DarwinNotificationDetails {
   /// This property is only applicable to iOS 15.0 and macOS 12.0 or newer.
   /// https://developer.apple.com/documentation/usernotifications/unnotificationcontent/3747256-interruptionlevel
   final InterruptionLevel? interruptionLevel;
+
+  /// The volume for playing the critical alert sound.
```
MaikuB, line 144: Can you separate the first sentence so it has its own paragraph? This is so it can follow [this rule](https://dart.dev/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph)

MaikuB:
Thanks for the PR. I'll need to give this a test but based on what I've seen, looks straightforward and expect it to work. Props for going through the details in the PR description too. May end up applying something like this for PR templates :)

bannzai:
@MaikuB Thank you for your reviews! I'm fixed!

flutter_local_notifications/example/lib/main.dart
```diff
@@ -720,12 +743,23 @@ class _HomePageState extends State<HomePage> {
                       buttonText: 'Request permission',
                       onPressed: _requestPermissions,
                     ),
+                    PaddedElevatedButton(
+                      buttonText:
+                          'Request permission with Critical Alert Permission',
```
MaikuB, line 748: minor thing but can you change the text to use sentence casing?

flutter_local_notifications/lib/src/platform_specifics/darwin/notification_details.dart
```diff
@@ -139,4 +140,16 @@ class DarwinNotificationDetails {
...
+  ///
+  /// The value is between 0.0 and 1.0.
+  /// This is only used when the notification is configured as a critical alert.
+  /// Document: https://developer.apple.com/documentation/usernotifications/unnotificationsound/criticalsoundnamed(_:withaudiovolume:)
+  ///
+  /// Subject to specific approval from Apple:
+  /// https://developer.apple.com/contact/request/notifications-critical-alerts-entitlement/
+  ///
+  /// This property is only applicable to iOS 12.0 or newer.
```
MaikuB, line 153: I believe this is missing mention of how it's also only applicable macOS 10.14 or newer too? If so, see some of the existing API docs in the code on the convention being used when a property is specific to certain versions of iOS and macOS. For convenience it's in the format of

```
On iOS, this property is only applicable to iOS <...> or newer.
On macOS, this property is only applicable to macOS <...> or newer.
```

MaikuB (changes requested):
Got to test and looks good. Noticed some other minor things that I commented on

bannzai:
Thank you for your review! Done!

MaikuB (approved):

PR #2606, opened 2025-04-26
[flutter_local_notifications] Remove redundant macOS version check

bqubique:
Remove unnecessary macOS 10.14 check since [Flutter's lowest supported macOS version is 10.14](https://docs.flutter.dev/deployment/macos).
> The minimum macOS version that your app supports. Flutter supports deploying apps to macOS 10.14 (Mojave) and later.

Bump dependency versions of the example app.

### DIFF
--- a/flutter_local_notifications/example/macos/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
+++ b/flutter_local_notifications/example/macos/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
@@ -59,6 +59,7 @@
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
+      enableGPUValidationMode = "1"
       allowLocationSimulation = "YES">
       <BuildableProductRunnable
          runnableDebuggingMode = "0">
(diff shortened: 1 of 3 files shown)


MaikuB (changes requested):
Thanks for the PR. The dependency updates aren't compatible though and is why the various checks are failing. Could you rollback the changes to the example app?

bqubique:
@MaikuB No worries, sorry couldn't check this in time due to other things!
I reverted example's dependency changes.
Thank you!

bqubique:
@MaikuB any updates, could you check the PR?

bqubique:
@MaikuB ready 👍.

MaikuB:
Thanks for making the change. Upon closer inspection though, I realised there are actually multiple places now that could do with similar changes. Would this be something you can also include in the PR? Without doing so, it will only partially address the redundancy

bqubique:
@MaikuB sure, will try to target where I see the redundant checks and update the PR 👍.

bqubique:
@MaikuB could you please check the latest commit? Thanks.

MaikuB:
Thanks took a look. There are an additional things that could be removed as a result of the changes

- `ErrorMessages` struct
- `ErrorCodes` struct
- `buildNSUserNotification` function

Would you be able to remove these as well?

bqubique:
@MaikuB should be ready!

MaikuB:
Spotted another bit that could've been removed so in the interest of time, pushed the change myself. Once all checks are green then I'll merge this in :)

PR #2607, opened 2025-04-30
[flutter_local_notifications] Add dataMimeType and dataUri for Android

mattbajorek:


MaikuB (changes requested):
Thanks for the PR. Would you be able to update the example app so it can fetch and show the value of these properties as well

MaikuB:
I'm going to close this one due to lack of response. I tried to push an update to the example app but wasn't able to either. I've created https://github.com/MaikuB/flutter_local_notifications/pull/2661 to be able to do this and will credit you once merged in. Thanks for your PR

Issue #2609, opened 2025-05-07
[flutter_local_notifications] How about using Named Parameters for some functions?


TaeBbong:
Hi @MaikuB ! I've used this package for my projects, and I'm really appreciate about this nice package.

I have some enhancement points;

* **Using Named Parameters for functions**
Functions we use such as zonedSchedule() or periodicallyShow() provides positional parameters like:
```dart
Future<void> zonedSchedule(
    int id,
    String? title,
    String? body,
    TZDateTime scheduledDate,
    NotificationDetails notificationDetails, {
    required AndroidScheduleMode androidScheduleMode,
    String? payload,
    DateTimeComponents? matchDateTimeComponents,
  })
```

When we use this, we sometimes get confused about which order to put parameters in it.
So I suggest named parameters like:
```dart
Future<void> periodicallyShow({
  required int id,
  String? title,
  String? body,
  required RepeatInterval repeatInterval,
  required NotificationDetails notificationDetails,
  required AndroidScheduleMode androidScheduleMode,
  String? payload,
});
```

If you think this would work well, I would like to get on this work then make pull request.

PR #2610, opened 2025-05-08
[flutter_local_notifications] Apply named parameters for functions in lib/src/flutter_local_notifications_plugin.dart

TaeBbong:
As mentioned in issue #2609 , 
I suggest using named parameters instead of positional so that we can use functions much more easily.

* No changes on functionalities, features.
* Only interface changed.
* All test codes and example files using plugin has been updated.

### DIFF
--- a/flutter_local_notifications/CHANGELOG.md
+++ b/flutter_local_notifications/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [vNext]
+
+* Bumped `msix` dev dependency in example app. This to fix the [issue](https://github.com/YehudaKremer/msix/issues/303) where the `msix` package stopped being able to created MSIX installers
+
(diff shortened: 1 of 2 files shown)


TaeBbong (commented):
Checked all changes

Levi-Lesches:
> Only interface changed.

Note that this is a _massive_ breaking change which breaks literally every use of this plugin out there... I very much agree that it would have been nicer to have this from the start, and I'm not saying it should never be done, but this isn't exactly a small change to do now.

TaeBbong:
@Levi-Lesches Yes, I totally agree with you. It will breaks every usecase of this plugin.
How about apply this changes into next big release?
Maybe other maintainers' decision would be needed:)

Issue #2633, opened 2025-06-09
Problem with custom sound when app is closed


wantroba:
**Describe the bug**
I'm having trouble showing notifications with custom sounds when I trigger notifications via WorkManager (https://pub.dev/packages/workmanager)

**To Reproduce**

1. Import the package workmanager: https://pub.dev/packages/workmanager
2. Set up a simple task that triggers a notification with a custom sound

```
Workmanager().initialize(callbackDispatcher, isInDebugMode: false);
      Workmanager().registerPeriodicTask(
          checkStatusPeriodicTaskName, checkStatusPeriodicTaskName,
          frequency: Duration(minutes: 15));
```

3. Here is the notification channel config:
```
AndroidNotificationDetails androidPlatformChannelSpecifics =
        AndroidNotificationDetails(
      "Title",
      "Message",
      importance: Importance.high,
      priority: Priority.high,
      color: ColorUtils.primaryColorLight,
      playSound: true,
      groupKey: groupKey,
      sound: RawResourceAndroidNotificationSound(
        'beep',
      ),
    );
```
4. Wait for the task
5. See error:
```
2025-06-09 18:02:44.456  8545-8545  MethodChan...ifications com.app.sample           E  Failed to handle method call (Ask Gemini)
                                                                                                    java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.res.Resources android.content.Context.getResources()' on a null object reference
                                                                                                    	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.hasInvalidRawSoundResource(FlutterLocalNotificationsPlugin.java:1746)
                                                                                                    	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.extractNotificationDetails(FlutterLocalNotificationsPlugin.java:1722)
                                                                                                    	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.show(FlutterLocalNotificationsPlugin.java:1660)
                                                                                                    	at com.dexterous.flutterlocalnotifications.FlutterLocalNotificationsPlugin.onMethodCall(FlutterLocalNotificationsPlugin.java:1451)
                                                                                                    	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:267)
                                                                                                    	at io.flutter.embedding.engine.dart.DartMessenger.invokeHandler(DartMessenger.java:292)
[...]

PR #2634, opened 2025-06-10
Don't set applicationContext to null onDetachedFromEngine

wantroba:
This error (#2633)  was occurring because applicationContext was null. So I made this change and the problem was solved. I also noticed that the awesome_notifications package does not set applicationContext to null at any time: https://github.com/dokumanx/awesome_notifications/blob/5e0959c57c612ba525a8118dbbfa416f32426d06/android/src/main/java/me/carda/awesome_notifications/AwesomeNotificationsPlugin.java

Please let me know if there is any reason for applicationContext to be set to null. Otherwise, just approve this PR. Thanks

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -1397,7 +1397,6 @@ public class FlutterLocalNotificationsPlugin
   public void onDetachedFromEngine(@NonNull FlutterPluginBinding binding) {
     this.channel.setMethodCallHandler(null);
     this.channel = null;
-    this.applicationContext = null;
   }
 
   @Override


MaikuB:
Thanks for the PR. The reason this was originally done was based what I had seen done by multiple major plugins back done to release resources. Even now I still see some plugins do so but some don't either so perhaps this isn't "official" advice

wantroba:
> Thanks for the PR. The reason this was originally done was based what I had seen done by multiple major plugins back done to release resources. Even now I still see some plugins do so but some don't either so perhaps this isn't "official" advice

It makes sense and the lib works perfectly for all cases except when I need to trigger a notification with the app in the background using the WorkManager lib (error #2633). I believe the resource is being released before the notification is triggered. As the awesome_notifications lib also did not set it to null, I made an adjustment and it solved my problem.

MaikuB:
Sounds like you're consistently running into the issue but sounds odd. The reason it is released is when no engine exists and would've expected more members of the community to report this too. I vaguely remember trying both plugins together and not running into this issue either. To confirm, when you use both plugins, do you call the following methods in the top-level/static function that you pass to the workmanager plugin and do so before calling the API to show a notification?

- https://api.flutter.dev/flutter/dart-ui/DartPluginRegistrant/ensureInitialized.html
- https://api.flutter.dev/flutter/widgets/WidgetsFlutterBinding/ensureInitialized.html

MaikuB:
FYI I tried reproducing this myself but periodic tasks don't seem to be working for me in general. I've seen others mention this on the workmanager repo as well. Do you by chance have a link to a repo with a minimal app that can reproduce this? Even when I clone its repo, its example doesn't work with a periodic task. A one-off works though I only did on debug mode. Does one-off with a delay work for you or do you run into the same issue?

iampato:
Add this :
```dart
 @pragma('vm:entry-point')
  static Future<void> messagingBackgroundHandler(RemoteMessage message) async {
```
seems to have solved it

MaikuB:
Thanks @iampato. I was suspecting this was the cause but was hoping @wantroba would respond with a minimal app. I know the work manager plugin essentially has a helper method that calls `WidgetsFlutterBinding.ensureInitialized()`. Either way I can't reproduce this and @wantroba hasn't responded for a long time so will be closing this

PR #2649, opened 2025-07-10
fix(windows): non-ascii characters in payload

### LINKED ISSUE #2648: [Windows] When the payload contains non-ASCII characters, onDidReceiveNotificationResponse does not work.
**Describe the bug**

When the payload contains non-ASCII characters, onDidReceiveNotificationResponse does not work in windows.

**To Reproduce**

1. Show a local notification with non-ASCII characters in payload.
2. Click local notification
3. Get this exception
```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: FormatException: Unexpected extension byte (at offset 10)
```
4. `onDidReceiveNotificationResponse` not trigger

**Expected behavior**

No error, and trigger `onDidReceiveNotificationResponse`

**Sample code to reproduce the problem**

yoyo930021:

Fix #2648

### DIFF
--- a/flutter_local_notifications_windows/src/plugin.cpp
+++ b/flutter_local_notifications_windows/src/plugin.cpp
@@ -32,14 +32,14 @@ struct NotificationActivationCallback :
       vector<StringMapEntry> entries;
       for (ULONG i = 0; i < count; i++) {
         auto item = data[i];
-        const std::string key = CW2A(item.Key);
-        const std::string value = CW2A(item.Value);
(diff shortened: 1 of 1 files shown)


Levi-Lesches (approved):
Code looks good to me. I'm not really that knowledgeable on character encodings, so I would think if this has been tested against both ascii and non-ascii payloads then it should be good.

MaikuB:
Thanks for the PR!

PR #2660, opened 2025-07-21
[flutter_local_notifications] Android Semantic actions

### LINKED ISSUE #2656: Support Android SemanticActions
When building `AndroidNotificationAction` allow setting `SemanticAction`, this is done through the `setSemanticAction` method on the Builder.

https://developer.android.com/reference/android/app/Notification.Action.Builder#setSemanticAction(int)

Adding support for this allows us to make use of Android Auto messaging notifications. Required values `SEMANTIC_ACTION_REPLY` & `SEMANTIC_ACTION_MARK_AS_READ`

https://developer.android.com/training/cars/communication/notification-messaging#create_actions

szechyjs:
Resolves #2656

Adds a `semanticAction` attribute to `AndroidNotificationAction` which gets passed through to the builder.

### DIFF
--- a/flutter_local_notifications/lib/src/platform_specifics/android/enums.dart
+++ b/flutter_local_notifications/lib/src/platform_specifics/android/enums.dart
@@ -269,6 +269,48 @@ enum AudioAttributesUsage {
   final int value;
 }
 
+/// The available semantic actions for Android notifications.
+enum SemanticAction {
+  /// Corresponds to [`Notification.Action.SEMANTIC_ACTION_NONE`](https://developer.android.com/reference/android/app/Notification.Action#SEMANTIC_ACTION_NONE).
+  none(0),
+
+  /// Corresponds to [`Notification.Action.SEMANTIC_ACTION_REPLY`](https://developer.android.com/reference/android/app/Notification.Action#SEMANTIC_ACTION_REPLY).
+  reply(1),
+
+  /// Corresponds to [`Notification.Action.SEMANTIC_ACTION_MARK_AS_READ`](https://developer.android.com/reference/android/app/Notification.Action#SEMANTIC_ACTION_MARK_AS_READ).
+  markAsRead(2),
+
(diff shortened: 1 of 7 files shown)


flutter_local_notifications/lib/src/platform_specifics/android/notification_details.dart
```diff
@@ -93,6 +95,13 @@ class AndroidNotificationAction {
   /// Set whether the notification should be canceled when this action is
   /// selected.
   final bool cancelNotification;
+
+  /// Sets the SemanticAction for this Action. A SemanticAction denotes what an
```
MaikuB: A few issues on the API docs

- Can you follow [this guide](https://dart.dev/effective-dart/documentation)? Stand out are making sure first sentence is its own paragraph based on [this rule](https://github.com/MaikuB/flutter_local_notifications/blob/master/CONTRIBUTING.md)
- Can you look at making this a better/more explicit explanation on what this would do? On its own the first sentence `Sets the SemanticAction for this Action` doesn't make sense and reads awkward with how it mentions action twice. I'd suggest looking at the official [Android docs](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Action.SemanticAction) for inspiration. As an example, "Provides meaning for this action." is a better explanation in plain English.
- Use "action" over "Action"

flutter_local_notifications/lib/src/platform_specifics/android/enums.dart
```diff
@@ -269,6 +269,48 @@ enum AudioAttributesUsage {
   final int value;
 }
 
+/// The available semantic actions for Android notifications.
```
MaikuB: The API doc here doesn't explain what a semantic action is. Based on [this](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Action.SemanticAction), a better explanation would along the lines of "The available meanings that can be associated with a notification action."

MaikuB (changes requested):
Thanks for the PR. It largely looks fine to me and left a couple of comments relating to API docs. Besides that, would you be able to do the following

- update the example app to demonstrate how the changes can be used
- add more tests as opposed to just updating existing ones. Stand out for me is how there isn't a test to check that setting `invisible` to `true` ensures it is also `true` when passed through the method channel

PR #2674, opened 2025-08-21
[flutter_local_notifications] Add support for specifying custom Activity to open on notification tap (Android)

uncledoc:
## Changes
- Added an optional initialization parameter for specifying the target Activity name.
- If the parameter is not provided, the default behavior remains unchanged.

## Why
In some applications, the default launcher Activity is not suitable as a notification entry point. 
This enhancement gives developers more flexibility and control over app navigation from notifications.

## Notes
- Backward compatible: existing code without the new parameter continues to work as before.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -132,6 +132,7 @@ public class FlutterLocalNotificationsPlugin
   private static final String CALLBACK_HANDLE = "callback_handle";
   private static final String DRAWABLE = "drawable";
   private static final String DEFAULT_ICON = "defaultIcon";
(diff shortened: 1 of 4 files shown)


MaikuB:
Thanks for the PR and sorry for the delay in getting back. I'll need to test this out but do you have some examples where an app shouldn't use the default activity? I have some further thoughts on this but wanted to know before I share

uncledoc:
I need this functionality primarily to help migrate a large existing application to Flutter.

However, the improvement isn't limited to migration. It also allows for greater isolation of specific flows. For instance, launching certain activities with FLAG_ACTIVITY_NEW_TASK (e.g., for in-app purchases or specific login flows) or catering to specific security use cases where task separation is required.

MaikuB:
Thanks for more details on this. Based on what you said, as initialisation is supposed to be a one-time thing, perhaps it would be more beneficial if the activity could be set per individual notification?

uncledoc:
That's a great suggestion and certainly an improvement over my current approach. I'll need to think about the best way to implement that change.

PR #2677, opened 2025-08-26
[flutter_local_notifications] Added ability to bind to ForegroundService

nathanielheitsch:
This PR adds the necessary components to the ForegroundService components to allow developers to listen for a broadcast and bind to the ForegroundService when it is created. This will promote the activity and reduce the likelihood of it being frozen or killed due to resource constraints by the OS. Updated the README and example app to show how it can be integrated if chosen.

Tested with the example app on Android 9 & 15

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -413,6 +413,75 @@ When specifying the large icon bitmap or big picture bitmap (associated with the
 
 ⚠️ For Android 8.0+, sounds and vibrations are associated with notification channels and can only be configured when they are first created. Showing/scheduling a notification will create a channel with the specified id if it doesn't exist already. If another notification specifies the same channel id but tries to specify another sound or vibration pattern then nothing occurs.
 
+
+### Bind ForegroundService to your FlutterActivity
+
+In your activity (e.g., `MainActivity.kt`), set up a broadcast receiver and a `ServiceConnection` to manage binding and unbinding. This is not required to use a `ForegroundService` but it will decrease the likelyhood of your activity being [killed or frozen by the OS while the activity is in the background](https://source.android.com/docs/core/perf/cached-apps-freezer#handling-custom-features):
+
+```kotlin
+class MainActivity: FlutterActivity() {
+    private var isServiceBound = false
+    private val serviceStartedAction = "com.dexterous.flutterlocalnotifications.FOREGROUND_SERVICE_STARTED"
+    private val serviceStoppedAction = "com.dexterous.flutterlocalnotifications.FOREGROUND_SERVICE_STOPPED"
+
+    private val serviceConnection = object : ServiceConnection {
+        override fun onServiceConnected(name: ComponentName?, binder: IBinder?) {
+            isServiceBound = true
+            Log.d("MainActivity", "Service bound")
+        }
+        override fun onServiceDisconnected(name: ComponentName?) {
+            isServiceBound = false
+            Log.d("MainActivity", "Service disconnected")
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


flutter_local_notifications/README.md
```diff
@@ -413,6 +413,75 @@ When specifying the large icon bitmap or big picture bitmap (associated with the
 
 ⚠️ For Android 8.0+, sounds and vibrations are associated with notification channels and can only be configured when they are first created. Showing/scheduling a notification will create a channel with the specified id if it doesn't exist already. If another notification specifies the same channel id but tries to specify another sound or vibration pattern then nothing occurs.
 
+
+### Bind ForegroundService to your FlutterActivity
```
MaikuB, line 417: The table of contents is missing an update that allows for jumping to this section

flutter_local_notifications/README.md
```diff
@@ -413,6 +413,75 @@ When specifying the large icon bitmap or big picture bitmap (associated with the
...
+
+In your activity (e.g., `MainActivity.kt`), set up a broadcast receiver and a `ServiceConnection` to manage binding and unbinding. This is not required to use a `ForegroundService` but it will decrease the likelyhood of your activity being [killed or frozen by the OS while the activity is in the background](https://source.android.com/docs/core/perf/cached-apps-freezer#handling-custom-features):
```
MaikuB, line 419: Can this start off by saying it's optional and only applicable for apps using foreground services? Thinking behind this is some readers may not pick this up or know enough about Android development to realise

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -2306,6 +2306,8 @@ private void startForegroundService(MethodCall call, Result result) {
             Intent intent = new Intent(applicationContext, ForegroundService.class);
             intent.putExtra(ForegroundServiceStartParameter.EXTRA, parameter);
             ContextCompat.startForegroundService(applicationContext, intent);
+            applicationContext.sendBroadcast(new Intent("com.dexterous.flutterlocalnotifications.FOREGROUND_SERVICE_STARTED")
```
MaikuB, line 2309: Let me know what you think but I believe it's best to do the following

- add additional folders that so that `com.dexterous.flutterlocalnotifications.intent.action` namespace exists. This is to follow convention of having the strings behind a namespace where last segment has `action` like other actions e.g. `android.intent.action.<...>` and `htc.com.intent.action.<...>`
- create a class within the `action` (i.e.  `com.dexterous.flutterlocalnotifications.intent.action` namespace) that holds public static strings for `com.dexterous.flutterlocalnotifications.action.FOREGROUND_SERVICE_STARTED`  and `com.dexterous.flutterlocalnotifications.action.FOREGROUND_SERVICE_STOPPED`. This to allow developers an easier way to reference the strings through the plugin. The best name I can come up with for the class is `FlutterLocalNotificationsPluginIntent`. It'd be similar to how developers can reference [`Intent.ACTION_VIEW_LOCUS`](https://developer.android.com/reference/android/content/Intent#ACTION_VIEW) . In this case, they can reference `FlutterLocalNotificationsPluginIntent.FOREGROUND_SERVICE_STARTED` and `FlutterLocalNotificationsPluginIntent.FOREGROUND_SERVICE_STOPPED`

MaikuB (changes requested):
Thanks for the PR and apologies for delay on getting back on this. I left some comments on the PR with some requested/proposed changes. Let me know what you think and happy to discuss further

PR #2681, opened 2025-09-05
[flutter_local_notifications] Allow notification actions on ios to work with remote notifications

Jonny1987:
## Changes

* Allow didReceiveNotificationResponse to continue even if isAFlutterLocalNotification is false
* When isAFlutterLocalNotification is false, the notification's payload is the whole of response.notification.request.content.userInfo rather than just the PAYLOAD key, since it doesn't exist for non-local notifications

## Why

Currently, to show notifications you need to send a data only message so that flutter_local_notifications can show the message, which will allow the action button to work correctly. (including a notification in the notification message when sending, as well as a category, will cause flutter_firebase_messaging to show it automatically but actions don't work). However, data only messages on iOS are not completely reliable and are sometimes delayed or never get received.

So it makes sense to allow actions to function correctly when a notification message that contains a notification and a category is received and automatically shown by flutter_firebase_messaging.

## Notes

Since the payload is set to response.notification.request.content.userInfo rather than response.notification.request.content.userInfo['PAYLOAD'] (since the PAYLOAD key won't exist if it's not a local notification), the response object (`NotificationResponse`) returned by the method `getNotificationAppLaunchDetails` and used in the functions given for `onDidReceiveNotificationResponse` and `onDidReceiveBackgroundNotificationResponse` will have a payload that isn't just the data that was sent in the notification message but also the `aps` block of the notification message and other keys such as `gcm.message_id`,`google.c.a.e`, `google.c.fid`, `google.c.sender.id`. Have these extra keys shouldn't cause any issues although it's not ideal to have extra keys in the data for these functions.


Jonny1987:
I've now used method swizzling to block the onMessageOpenedApp channel message of firebase messaging so that the `FirebaseMessaging.onMessageOpenedApp.listen` will not get used for iOS since flutter_local_notification's onDidReceiveNotificationResponse is used for all notificaitons (local and non-local)

MaikuB:
After looking at what you've mentioned, this is what I've come to understand

- You noticed there are issues with data only messages sent via push notifications so looked for a more reliable solution that requires use of notification messages
- Your changes now impact how the FCM plugin itself behaves

Can you please confirm if the following is correct? If so, then this is a specific to your case. There may be others that could be benefit from it but is a scenario where forks or a custom implementation are more appropriate. Merging this will affect the FCM plugin's behavior specific to your case but will affect the rest of the community. It also creates an inconsistent mental model as this plugin, which is for local notifications, is now handling push/remote notifications. Additionally, there is inconsistent behaviour that you mentioned between Android and iOS

PR #2696, opened 2025-09-19
Windows unit tests no longer use the Windows SDK

### LINKED ISSUE #2695: Windows unit tests are too flaky
They keep randomly "failing" It'll show 20 success messages, then say "19 tests passed". There's just something different with the runtime environment, and I'm not sure showing notifications on a server is the best way to go about it. 

I'm thinking to reduce the validation down to: 
- build the DLL to ensure everything still compiles
- Convert all the details classes to XML and verify some basic structure + elements, using Dart
- some error handling before the C++ layer, like trying to initialize() twice, using bad GUIDs, etc

Fix is in #2696

Levi-Lesches:
Fixes flaky Windows unit tests on every PR, without losing too much. To actually test notifications, run the example app manually. 

Fixes #2695


MaikuB (changes requested):
Thanks for looking into this. Could the issue have been addressed without having to change the main plugin the way it's been done in this PR? The issue that sticks is how the plugin required changes and makes use of a global variable to determine if the plugin should use a real implementation or a mock. Two antipatterns/code smells are happening as a result

Levi-Lesches:
Added a new constructor marked `@visibleForTesting` that allows passing in the mocked bindings directly

flutter_local_notifications_windows/lib/src/plugin/ffi.dart
```diff
@@ -22,10 +24,28 @@ extension on String {
...
 class FlutterLocalNotificationsWindows extends WindowsNotificationsBase {
   /// Creates an instance of the native plugin.
-  FlutterLocalNotificationsWindows();
+  FlutterLocalNotificationsWindows()
+      : _bindings = NotificationsPluginBindings(
+            DynamicLibrary.open('flutter_local_notifications_windows.dll'));
+
+  /// Creates an instance of this plugin with the given (mocked) bindings.
+  @visibleForTesting
+  FlutterLocalNotificationsWindows.withBindings(
```
MaikuB, line 46: I was about to mention perhaps to rename it to `FlutterLocalNotificationsWindows.private` as it would be consistent with what you had on Linux. There was also an element of how the `with<...>` convention can lead to issues (repetition from readability perspective and could break if more arguments got introduced).

Rather than taking this approach, I noticed first-party plugins are taking an optional, nullable argument with the `@visibleForTesting` annotation applied to it. I think this is actually a good approach as implementation and test code would point to the same constructor. Would you be able to switch it so that the only constructor follows that pattern? An example of what I mean can be found [here](https://github.com/flutter/packages/blob/6640af6fafb019e9c879c52b4021f62f3e37cc19/packages/file_selector/file_selector_ios/lib/file_selector_ios.dart#L13)

flutter_local_notifications_windows/lib/src/plugin/ffi.dart
```diff
@@ -22,10 +24,28 @@ extension on String {
       this[23] == '-';
 }
 
+/// Does a basic syntax check on XML.
+bool _checkXml(String xml) {
```
MaikuB, line 28: It's file private but look this could have been private within the `FlutterLocalNotificationsWindows` class. Would you be able to adjust this?

Note: on a related note and not that it has to be done in this PR but I believe I missed picking this up when it comes to `_globalLaunchCallback` as well

flutter_local_notifications_windows/lib/src/ffi/mock.dart
```diff
@@ -0,0 +1,55 @@
...
+// ignore_for_file: type_annotate_public_apis, always_specify_types
+
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+
+import 'bindings.dart';
+
+/// Mocked FFI bindings.
+class MockBindings implements NotificationsPluginBindings {
```
MaikuB, line 11: Can you place this in the test folder? It's purely test related and is the approach taken when creating mocks  via mocktail or mockito or hand crafted ones. Note this is also currently a stub not a mock. A mock provides the ability to record the interactions that can then be used to verify what took place. Can you also rename the class and adjust the docs accordingly to reflect this?

flutter_local_notifications_windows/lib/src/plugin/stub.dart
```diff
@@ -1,8 +1,22 @@
+import 'package:meta/meta.dart';
+
 import '../details.dart';
+import '../ffi/bindings.dart';
 import 'base.dart';
 
 /// The Windows implementation of `package:flutter_local_notifications`.
 class FlutterLocalNotificationsWindows extends WindowsNotificationsBase {
```
MaikuB, line 8: With the comment I left that allows the FFI implementation to take an argument for the bindings used behind the scenes, does this perhaps render the stub redundant now? Based on my own knowledge and testing, it would be redundant but let me know if I've missed anything

MaikuB (changes requested):
Thanks for making the changes. I took a pass and believe there are some improvements that could be done

PR #2701, opened 2025-10-03
Update minSdkVersion from 21 to 24

SunnatilloShavkatov:


alexmercerind:
Why? Its good to have support for older API levels, if we already have compatible code in this plugin.

MaikuB:
Agreed. I know there's a smaller percentage of devices so the minimum will change. There's no reason given in the PR itself either. Will close this PR but in future, if you plan to raise a PR to any repository, please put in a description with more information

PR #2702, opened 2025-10-11
[flutter_local_notifications] Fix Android notification callback not firing after app auto-start

tikoua:
Fixed an issue on the Android platform where clicking a notification after the app auto-started would not trigger the onDidReceiveNotificationResponse callback.
The original implementation only handled the onNewIntent scenario. However, when the app was auto-restarted in the background (e.g., by JobScheduler or other mechanisms), clicking a notification did not invoke the expected callback. To address this, additional logic was added in the onAttachedToActivity callback to parse and handle the specific notification action.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -1427,7 +1427,7 @@ public class FlutterLocalNotificationsPlugin
       }else if(SELECT_NOTIFICATION.equals(mainActivityIntent.getAction())){
         Map<String, Object> notificationResponse = extractNotificationResponseMap(mainActivityIntent);
         if (SELECT_FOREGROUND_NOTIFICATION_ACTION.equals(mainActivityIntent.getAction())) {
-          processForegroundNotificationAction(intent, notificationResponse);
+          processForegroundNotificationAction(mainActivityIntent, notificationResponse);
         }
         channel.invokeMethod("didReceiveNotificationResponse", notificationResponse);
       }


Adelmostafa31:
the intent in line 1430 not exist.

tikoua:
> the intent in line 1430 not exist.

You’re right — the intent at line 1430 doesn’t exist.
I’ve fixed it in a follow-up commit 83b71ebf0f08c005dfdebcca0610f92eb88cc348.

MaikuB:
Can you provide instructions on how to reproduce the issue that is supposed to be fix by this PR? Would this require certain phones to also reproduce the issue?

tikoua:
> Can you provide instructions on how to reproduce the issue that is supposed to be fix by this PR? Would this require certain phones to also reproduce the issue?

This issue is not device-specific, and most Flutter apps will never encounter it. However, in my case, I’m developing an instant-messaging app, and on Android we use a daemon process to revive the main process. This means the Dart runtime and Flutter engine may be started in the background. If a notification arrives during that time and the user taps it, the issue is triggered.
In short: after the app auto-starts in the background, but before the user manually opens the app, tapping a notification will reproduce the problem.

MaikuB:
Thanks for the explanation do you have more detailed guidance on how I could reproduce this? If possible an minimal app that can reproduce this would be great. Asking not because I don't believe you but so I can test and see if there's other ramifications of this change. From what I could see, it looks like there could also be a timing issue that this PR may in theory cause the callback to be fired on app launch from a notification from a cold start. To prevent this from happening, I believe [this PR](https://github.com/MaikuB/flutter_local_notifications/pull/2212) would need to be revisited. In my mind, it now sounds like best approach is to avoid a method to add the extra info into the intent and just do it whenever an intent is processed that a callback is fired. This presumably would help track if a notification has already been processed that it would avoid the timing issue

Issue #2338, opened 2024-06-09
Can't access attached activity lifecycle in Kotlin Native through MethodChannel call inside onDidReceiveBackgroundNotificationResponse


mtiendat:
**Describe the bug**
Global variable activityPluginBinding.activity is null when it call in setMethodHandler and app is running in background 

**To Reproduce**

1. Set Up flutter notification
2. Set action for this notification
3. Set trigger action notification in background, call method channel inside this trigger in onDidReceiveBackgroundNotificationResponse.
4. Create a class FlutterFlugin and create method(close_app) and set channel.setMethodHandlder to work in background. Implement ActivityAware to listener lifecycle, save activityPluginBinding.activity to global variable activity
5. Call activityPluginBinding.activity.finish() in close_app method
6. global variable activityPluginBinding.activity is always null in background, only work in foreground

**Expected behavior**
Can access global variable activityPluginBinding.activity inside channel.setMethodHandlder, can finish app

**Sample code to reproduce the problem**
 ```
@pragma('vm:entry-point')
  static Future<void> notificationTapBackground(
    NotificationResponse notificationResponse,
  ) async {
    if (actionClose.id == notificationResponse.actionId) {
    
      const method = MethodChannel('close_app_plugin');
      method.invokeMethod('close_app');
    }
  }

  static Future<void> initializeService() async {
    if (Platform.isIOS || Platform.isAndroid) {
      await _flutterLocalNotificationsPlugin.initialize(
        const InitializationSettings(
          iOS: DarwinInitializationSettings(),
          android: AndroidInitializationSettings('mipmap/ic_launcher'),
        ),
        onDidReceiveBackgroundNotificationResponse: notificationTapBackground,
      );
    }
    Platform.isAndroid
        ? await _flutterLocalNotificationsPlugin
            .resolvePlatformSpecificImplementation<
                AndroidFlutterLocalNotificationsPlugin>()
            ?.requestNotificationsPermission()
        : await _flutterLocalNotificationsPlugin
            .resolvePlatformSpecificImplementation<
                IOSFlutterLocalNotificationsPlugin>()
            ?.requestPermissions(
              alert: true,
              badge: true,
              sound: true,
            );

    await _flutterLocalNotificationsPlugin
        .resolvePlatformSpecificImplementation<
            AndroidFlutterLocalNotificationsPlugin>()
        ?.createNotificationChannel(channel);
    await _flutterLocalNotificationsPlugin
[...]

PR #2736, opened 2025-12-19
[flutter_local_notifications] Add `FlutterLocalNotificationsPlugin.setConfigureFlutterEngineCallback` for Android

alexmercerind:
### Summary

If a `AndroidNotificationAction` with `showsUserInterface: false` is clicked, a new Flutter engine is created by the plugin internally. I have some platform channels inside my `MainActivity`, which I want to be accessible inside this newly created engine as well. https://github.com/MaikuB/flutter_local_notifications/issues/2338 describes a similar situation.

Thus, something similar to following can be used to configure the newly created Flutter engine e.g. registering a new platform channel:

```kt
override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)
    FlutterLocalNotificationsPlugin.setConfigureFlutterEngineCallback {
        configureFlutterEngine(it)
    }
}
```

```kt
override fun onCreate(savedInstanceState: Bundle?) {
    super.onCreate(savedInstanceState)
    FlutterLocalNotificationsPlugin.setConfigureFlutterEngineCallback(::configureFlutterEngine)
}
```


MaikuB:
Thanks for the PR. Will break up what some things I wanted to raise

> If a AndroidNotificationAction with showsUserInterface: false is clicked, a new Flutter engine is created by the plugin internally.

You may know this already but just in case you didn't, note this is very much intentional as it allows logic to run without the app running

>  I have some platform channels inside my MainActivity, which I want to be accessible inside this newly created engine as well.

Can you describe your use case here from the end user's point of view?  What happens when the app with the main isolate isn't running? Presumably no logic is run so is one of the reasons why I wanted to ask as it sounds rather odd and specific scenario to cater for. Further to that there's a few things that seem off to me so can you confirm if my understanding of the following is correct:

- This PR is only Android-specific
- This PR allows accessing platform channels (and plugins) created in the main isolate via the `onDidReceiveBackgroundNotificationResponse`. The logic only works on Android (note: if so, what happens on other platforms?)

I'm asking these questions as if these are both true, it would leading an inconsistent mental model. Additionally, when I took a look at other (notable) plugins supporting background isolates (e.g. the FCM plugin),  I hadn't spotted similar changes either. Other thing is presumably the platform channels are on the main isolate. If so, my understanding was [messaging passing](https://docs.flutter.dev/perf/isolates#message-passing-between-isolates) should be used. Have you looked into this?

All up what I'm getting to here is that provided I understand the problem correctly, I believe there is an out of the box solution for your scenario without relying on something that I've not seen done elsewhere. This could in turn lead to incorrect expectations of what the "standard" expectation as developers switch between plugins (i.e. work with the broader plugin ecosystem) that also make use of background isolates

Edit: please let me know I'm missing something including if I've misunderstood too. If you have references to other plugins supporting a similar thing to this PR then this would help a great deal too

MaikuB:
Closing this since I've not back on this one for a while

PR #2759, opened 2026-02-19
[flutter_local_notifications] Add openAppNotificationSettings() for Android and iOS

hamadcc:
Closes #2752

## Summary
- Add `openAppNotificationSettings()` to Android and iOS platform implementations.
- Android opens the app’s notification settings screen (falls back to app settings if needed).
- iOS attempts to open the app’s notification settings (iOS 15.4+), otherwise it falls back to app settings (iOS may still redirect depending on OS/device).

## Test plan
- `cd flutter_local_notifications && flutter test`
- Manual: verified from example app on Android emulator and iOS simulator/device

### DIFF
--- a/flutter_local_notifications/README.md
+++ b/flutter_local_notifications/README.md
@@ -100,11 +100,13 @@ Note: the plugin requires Flutter SDK 3.22 at a minimum. The list of support pla
 * [Android] Full-screen intent notifications
 * [Android] Start a foreground service
 * [Android] Ability to check if notifications are enabled
+* [Android] Open app notification settings
 * [iOS (all supported versions) & macOS 10.14+] Request notification permissions and customise the permissions being requested around displaying notifications
 * [iOS 10+] Request CarPlay notification permissions for notifications to appear in CarPlay interface
 * [iOS 10 or newer and macOS 10.14 or newer] Display notifications with attachments
 * [iOS 12.0+] Support for custom notification settings UI via "Configure Notifications in <application name>" button in notification context menu (API available on macOS 10.14+ but UI button does not appear in practice)
 * [iOS and macOS 10.14 or newer] Ability to check if notifications are enabled with specific type check
+* [iOS 15.4+] Attempt to open the app's notification settings page (falls back to app settings on older iOS versions)
 * [Linux] Ability to to use themed/Flutter Assets icons and sound
 * [Linux] Ability to to set the category
 * [Linux] Configuring the urgency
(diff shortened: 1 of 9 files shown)


MaikuB (changes requested):
Thanks for the PR. I have a few questions/requests

- Could you add the method to the platform interface and then make the API available through the cross-platform facing API as well? This may end up allowing the example app changes to be refactored as well
- Could you add macOS support?
- The documentation you added says the APIs on iOS will try to open the app's notification settings. This is normally done through the `UIApplicationOpenNotificationSettingsURLString` that you also mention in the code comments. I see other plugins/libraries use this too (e.g. app_settings plugin) but did not see any code for it. Did you by chance forget to try to add in logic to do this? Currently it always open the general settings

hamadcc:
Hi @MaikuB, thanks for the review.

You're right about the platform interface / cross-platform API point. I only added the method to the platform-specific implementations, but it makes sense to expose it through the shared API as well, and I'll update that.

On iOS, I had a best-effort flow using `UIApplicationOpenNotificationSettingsURLString` with fallback to `UIApplicationOpenSettingsURLString`. Manual testing showed navigation can vary by OS/device. A later simplification that only opened general app settings ended up in the PR by mistake.

I'll update it to try `UIApplicationOpenNotificationSettingsURLString` first on supported iOS versions, then fall back to the app's general settings page.

For macOS, I may not be able to take that on right away as I'm a bit occupied at the moment, but I can make the other two changes first.

hamadcc:
Hi @MaikuB, I’ve pushed an update that addresses the first and third points:

- Exposed `openAppNotificationSettings()` via the platform interface + `FlutterLocalNotificationsPlugin` (and refactored the example/tests accordingly).
- Updated iOS to try `UIApplicationOpenNotificationSettingsURLString` first and fall back to `UIApplicationOpenSettingsURLString`.

Could you please take another look when you have a moment? Thanks!

MaikuB (changes requested):
Thanks for making the changes. Looks good to me though the pipeline is failing as there are some formatting changes to address. Would you be able to make the changes to fix this?

Another thing is let me if you intend to tackle macOS as part of this PR or if it'll be done in a separately. Either is fine with me and was more for my own awareness as there may be an opportunity where I can implement it

hamadcc:
@MaikuB Thanks. I've pushed a formatting fix, so the pipeline failure should be resolved now.

For macOS, I won't be able to take that on, so it won't be part of this PR. Please feel free to handle that separately if you'd liike.

MaikuB (approved):

PR #2760, opened 2026-02-19
[flutter_local_notifications] Add custom notification view support with text replacement and actions

Rishyash:
- Added CustomViewMapping class to map view IDs to text/actions
- Added CustomNotificationView class for custom layouts
- Support for customContentView, customBigContentView, and customHeadsUpContentView
- Dynamic text replacement in XML layouts from Flutter
- Button/view click action handling
- Comprehensive unit and integration tests
- Backward compatible with existing API

### DIFF
--- a/flutter_local_notifications/android/gradle/wrapper/gradle-wrapper.properties
+++ b/flutter_local_notifications/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,0 +1,7 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.0-milestone-1-bin.zip
+networkTimeout=10000
+validateDistributionUrl=true
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -966,6 +966,7 @@ public class FlutterLocalNotificationsPlugin
               intent.putExtra(NOTIFICATION_ID, notificationDetails.id);
               intent.putExtra(ACTION_ID, mapping.actionId);
               intent.putExtra(PAYLOAD, notificationDetails.payload);
+              intent.putExtra(CANCEL_NOTIFICATION, true);
               
               if (notificationDetails.tag != null) {
                 intent.putExtra(NOTIFICATION_TAG, notificationDetails.tag);
@@ -1576,7 +1577,8 @@ public class FlutterLocalNotificationsPlugin
     mainActivity = binding.getActivity();
     Intent mainActivityIntent = mainActivity.getIntent();
     if (!launchedActivityFromHistory(mainActivityIntent)) {
-      if (SELECT_FOREGROUND_NOTIFICATION_ACTION.equals(mainActivityIntent.getAction())) {
+      if (SELECT_FOREGROUND_NOTIFICATION_ACTION.equals(mainActivityIntent.getAction())
+          || SELECT_NOTIFICATION.equals(mainActivityIntent.getAction())) {
         Map<String, Object> notificationResponse =
             extractNotificationResponseMap(mainActivityIntent);
         processForegroundNotificationAction(mainActivityIntent, notificationResponse);
@@ -2186,7 +2188,8 @@ public class FlutterLocalNotificationsPlugin
     if (SELECT_NOTIFICATION.equals(intent.getAction())
         || SELECT_FOREGROUND_NOTIFICATION_ACTION.equals(intent.getAction())) {
       Map<String, Object> notificationResponse = extractNotificationResponseMap(intent);
-      if (SELECT_FOREGROUND_NOTIFICATION_ACTION.equals(intent.getAction())) {
+      if (SELECT_FOREGROUND_NOTIFICATION_ACTION.equals(intent.getAction())
+          || SELECT_NOTIFICATION.equals(intent.getAction())) {
         processForegroundNotificationAction(intent, notificationResponse);
       }
       channel.invokeMethod("didReceiveNotificationResponse", notificationResponse);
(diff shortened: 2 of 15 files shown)


Mohit8G:
Any update for review?

flutter_local_notifications/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,7 @@
+distributionBase=GRADLE_USER_HOME
```
MaikuB, line 1: Can you remove the gradle/wrapper folder that includes this file? Plugins don't need this checked in eg. check the repo for first party plugins as a reference

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -848,6 +876,127 @@ private static byte[] castObjectToByteArray(Object data) {
...
+  /**
+   * Retrieves the layout resource ID by name.
+   * 
+   * @param context The application context
+   * @param layoutName The name of the layout resource (without .xml extension)
+   * @return The resource ID, or 0 if not found
+   */
+  private static int getLayoutResourceId(Context context, String layoutName) {
+    if (layoutName == null || layoutName.isEmpty()) {
+      Log.e(TAG, "Custom layout name is null or empty");
```
MaikuB, line 888: This is a general comment and question for similar scenarios where an error is logged or an exception has been swallowed: what's the reason to only log this? These look like erroneous scenarios and if so, it would be better to throw an exception (e.g. make use of the `PluginException` class) so that plugin consumers can be notified to handle scenarios like this

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -848,6 +876,127 @@ private static byte[] castObjectToByteArray(Object data) {
...
+            } catch (Exception e) {
+              // Silently continue if text setting fails
+            }
+          }
+          
+          // Set click action if provided
+          if (mapping.actionId != null) {
+            try {
+              Intent intent = getLaunchIntent(context);
+              intent.setAction(SELECT_NOTIFICATION);
```
MaikuB, line 965: `SELECT_NOTIFICATION` is for when a user taps on a notification. However, this portion of the code is for actions so this code and other additions you have should be adjusted to use `SELECT_FOREGROUND_NOTIFICATION_ACTION` instead.

On a related note, have you checked to see if having actions triggered callbacks via a background isolate is possible? Conscious on if adding support for it later will result in breaking changes or backwards compatibility issues that it may be better to add support for them from the get go

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -848,6 +876,127 @@ private static byte[] castObjectToByteArray(Object data) {
...
+   * @param customView The custom view configuration
+   * @return A configured RemoteViews object or null if layout not found
+   */
+  @Nullable
+  private static RemoteViews createCustomRemoteViews(
+      Context context,
+      NotificationDetails notificationDetails,
+      CustomNotificationView customView) {
+    
+    if (customView == null || customView.layoutName == null) {
```
MaikuB, line 926: This method looks quite long overall. Are you able to refactor it in a logical matter so that it calls other methods?

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -1010,6 +1159,14 @@ private static void setStyle(
       Context context,
       NotificationDetails notificationDetails,
       NotificationCompat.Builder builder) {
+    // Skip setting style if custom views are present
```
MaikuB, line 1162: I would be inclined let it clash if it comes to it if it comes to it. This way it aligns with how the native APIs work and aligns with how the rest of plugin is designed to do so as well. In other words, this section could be removed.

More importantly, there is also a key part missing in this PR where the Flutter side should allow the consumer of this plugin to specify a  subclass of `StyleInformation` that when processed on the Android, maps to [NotificationCompat.DecoratedCustomViewStyle](https://developer.android.com/reference/androidx/core/app/NotificationCompat.DecoratedCustomViewStyle) and will call `builder.setStyle()` accordingly. Without this, the notification content can appear cropped . It already appears cropped with the additions made in the example app so those should also be updated to provide a more complete working example once support for  [NotificationCompat.DecoratedCustomViewStyle](https://developer.android.com/reference/androidx/core/app/NotificationCompat.DecoratedCustomViewStyle) is added

flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -848,6 +876,127 @@ private static byte[] castObjectToByteArray(Object data) {
...
+          
+          // Set click action if provided
+          if (mapping.actionId != null) {
+            try {
+              Intent intent = getLaunchIntent(context);
+              intent.setAction(SELECT_NOTIFICATION);
+              intent.putExtra(NOTIFICATION_ID, notificationDetails.id);
+              intent.putExtra(ACTION_ID, mapping.actionId);
+              intent.putExtra(PAYLOAD, notificationDetails.payload);
+              intent.putExtra(CANCEL_NOTIFICATION, true);
```
MaikuB, line 969: Rather than assuming the action should cancel the notification, this could be an option left to developers to decide on. As reference, you can see how Android notification actions in the plugin provide a boolean property that allows developers decide if they want the action to cancel the notification. Once this is added in, it would be better to encapsulate the action id and if the action cancels the notification through another class e.g. `CustomViewActionMapping`

MaikuB (changes requested):
Apologies for the delay and thanks for the PR. I was conscious of this affecting the release I wanted to get out with web support and have had to deal with some personal matters.

I have reviewed this and left some feedback on this. Let me know if you have questions on the feedback I left.  A few of the key callouts

- PR is missing a key addition that provides a mapping to [NotificationCompat.DecoratedCustomViewStyle](https://developer.android.com/reference/androidx/core/app/NotificationCompat.DecoratedCustomViewStyle) and calling [setStyle()](https://developer.android.com/reference/androidx/core/app/NotificationCompat.Builder#setStyle(androidx.core.app.NotificationCompat.Style)) accordingly. This is why the additions in the example app doesn't fully show the notification content. Making use of the style is mentioned in [official Android docs](https://developer.android.com/develop/ui/views/notifications/custom-notification)
- Seeing if support for actions through background isolate is possible and can be added in same PR
- Looking at how the errors/exceptions have been handled

PR #2775, opened 2026-04-27
[flutter_local_notifications] highlight app row when opening DND access settings

claudius-kienle:
When `requestNotificationPolicyAccess()` is called, the plugin opens `Settings.ACTION_NOTIFICATION_POLICY_ACCESS_SETTINGS` - a system page listing every installed app. The user then has to scroll the list and locate the calling app themselves, which is a poor UX for a permission the app just prompted for.
                                                                                                                                                                                                                                  
This PR attaches the `:settings:fragment_args_key` and `:settings:show_fragment_args` extras (with the package name as the key) to the launch intent. Android's Settings app uses these to scroll the matching row into view and briefly highlight it.

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -1984,9 +1984,15 @@ public class FlutterLocalNotificationsPlugin
       permissionRequestProgress = PermissionRequestProgress.None;
     } else {
       permissionRequestProgress = PermissionRequestProgress.RequestingNotificationPolicyAccess;
-      mainActivity.startActivityForResult(
-          new Intent(Settings.ACTION_NOTIFICATION_POLICY_ACCESS_SETTINGS),
-          NOTIFICATION_POLICY_ACCESS_REQUEST_CODE);
+      // Highlights the app's row on the settings list.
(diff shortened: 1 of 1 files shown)


flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -1984,9 +1984,15 @@ public void requestNotificationPolicyAccess(@NonNull PermissionRequestListener c
       permissionRequestProgress = PermissionRequestProgress.None;
     } else {
       permissionRequestProgress = PermissionRequestProgress.RequestingNotificationPolicyAccess;
-      mainActivity.startActivityForResult(
-          new Intent(Settings.ACTION_NOTIFICATION_POLICY_ACCESS_SETTINGS),
-          NOTIFICATION_POLICY_ACCESS_REQUEST_CODE);
+      // Highlights the app's row on the settings list.
+      String pkg = applicationContext.getPackageName();
```
MaikuB: Please rename `pkg` to `packageName`. Looking to avoid shorthand and abbreviations for readability and aligns with rest of codebase

MaikuB (changes requested):
Thanks for the PR. Have one comment that's an easy one to address around variable naming

claudius-kienle:
Thank you for responding. I renamed the variable as suggested.

MaikuB (approved):

Issue #2615, opened 2025-05-12
AOT Compilation Error with flutter_local_notifications


AndriiReverchuk:
**Describe the bug**
When building a Flutter application for Windows, the AOT compilation fails with an error related to the `flutter_local_notifications_windows` plugin. Specifically, the snapshot generator crashes with an "Unexpected object" error referring to the `NativeLaunchDetails` class in the plugin's bindings.

**To Reproduce**
1. Add flutter_local_notifications ^19.2.0 to pubspec.yaml
2. Configure the plugin according to the README
3. Run `flutter build windows --release `
4. Build fails during AOT compilation with error code -1073740791

**Expected behavior**
The Windows application should build successfully without any errors related to the AOT compilation.

**Sample code to reproduce the problem**
I am using the standard Flutter application with local notifications added. The error occurs during the build process, specifically in the AOT compilation step.

**Error details**
Unexpected object (Class with illegal cid, full-aot): 0x1e3e8fa1ba1 Library:'package/src/ffi/bindings.dart' Class: NativeLaunchDetails
Dart snapshot generator failed with exit code -1073740791
Target aot_elf_release failed: Exception: AOT snapshotter exited with code -1073740791
**Environment info**
flutter_local_notifications: ^19.2.0
[√] Flutter (Channel stable, 3.29.3, on Microsoft Windows [Version 10.0.26100.3915], locale uk-UA)
[√] Windows Version (Њ ©Єа®б®дв Windows 11 Pro 64-bit, 24H2, 2009)
[√] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[√] Chrome - develop for the web
[√] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.13.6)
[√] Android Studio (version 2022.2)
[√] VS Code (version 1.95.3)
[√] Connected device (3 available)
[√] Network resources
**Additional context**
The issue appears to be specific to the Windows build and related to FFI bindings in the plugin. The error only occurs during the AOT compilation for release mode.

bitscoper:
I'm also facing the same error `MSB8066` with `exit code -1073740791`.

@MaikuB

MaikuB:
I can't reproduce this with the example app. Can you provide a link to repository hosting a minimal app that can reproduce the issue?

bitscoper:
> I can't reproduce this with the example app. Can you provide a link to repository hosting a minimal app that can reproduce the issue?

[...]

PR #2788, opened 2026-06-20
[flutter_local_notifications_windows] Fix Issue #2615 for windows build failure

SurenaSKQ:
Read the following for the rationale behind discarding the struct altogether:

https://github.com/MaikuB/flutter_local_notifications/issues/2615#issuecomment-4745825231

This MAY be a Dart-side bug actually, so this fix is more of a hack to get apps compiling; I'm not sure what exactly is fucking the AOT snapshotter and honestly I'm not going to investigate it. Thus, it may be a good idea to open up a channel with flutter devs to take a look at the issue too.


SurenaSKQ:
Welp, it fixed itself. Probably a flutter bug that got resolved.

PR #2791, opened 2026-06-28
[Various]Expose notification delivery and response timestamps to close issue #2780

### LINKED ISSUE #2780: Expose the notification delivery timestamp in NotificationResponse
## Problem

For repeating local notifications, the Dart `NotificationResponse` tells the app which notification was acted upon, **but it does not expose the timestamp of the delivered notification occurrence**.

On iOS, this information appears to be available from:

```objc
response.notification.date
```

However, it is not currently forwarded to Dart.

This makes it difficult to safely process responses to repeating reminders and know which reminder the response is referring too:

- a repeating notification fires more than once;
- the user responds late;
- the app was terminated when the notification was delivered;
- the same callback is replayed or delivered more than once.
- Using the callback processing time is not sufficient because it may differ from the scheduled occurrence.

## Proposed change
Add an optional delivery timestamp to NotificationResponse, for example:

`final DateTime? deliveredAt;`
On iOS and macOS, populate it from:

`response.notification.date`
The value should be forwarded consistently for:

- foreground notification responses;
- background notification actions;
- notification launch details.

Platforms that cannot provide the timestamp could return null.

## Backwards compatibility
This should be a non-breaking API addition if:

- the new field is nullable;
- the constructor argument is optional;
- existing callback signatures remain unchanged;
- platforms that do not support the value return null.

## Use case
A medication reminder app uses native repeating notifications to stay within the iOS pending-notification limit. When the user taps "Taken" on the reminder option, the app needs to associate the response with the correct scheduled dose occurrence.

For example, an 08:00 repeating reminder may be recorded as taken at 10:30. The app should record:

scheduled occurrence: 08:00
action time: 10:30
Exposing the native delivery timestamp allows this to be handled accurately and idempotently.

## Possible implementation outline
Extend the platform-interface NotificationResponse model with an optional delivery timestamp.
Include the timestamp in the Apple-platform response dictionary.
Parse and forward it through the foreground, background-dispatcher, and launch-details paths.
Add tests covering supported platforms and the nullable fallback.
I would be happy to prepare a PR if this approach fits the package API.

ventisep:
Fixes #2780

## Summary

Adds optional notification timing fields to `NotificationResponse`:

- `notificationDeliveredAt`: the time the OS delivered the notification, where available.
- `responseReceivedAt`: the time the native plugin received the notification response callback.

These fields are currently populated on iOS and macOS using Apple’s UserNotifications APIs.

## Motivation

Apps can receive notification responses later than the user interaction, especially when handling notification actions after the app/background isolate has been started.

Previously, apps that needed to record when a notification was delivered or when the native response callback was received had to fall back to the time Dart processed the callback. That can be inaccurate if Dart handling is delayed.

This change exposes the native timing information so apps can distinguish:

- when the notification was delivered
- when the native response callback was received
- when Dart eventually processed the response

The new fields are nullable, so existing platforms and existing callers remain backwards compatible.

## Implementation notes

- Adds nullable `notificationDeliveredAt` and `responseReceivedAt` fields to the platform interface `NotificationResponse`.
- iOS uses `UNNotification.date` for `notificationDeliveredAt`.
- iOS captures `responseReceivedAt` once when `userNotificationCenter:didReceiveNotificationResponse:withCompletionHandler:` is invoked.
- macOS mirrors the same behavior using `UNNotification.date` and the native notification response delegate callback time.
- Dart method-channel parsing converts epoch millisecond values into nullable `DateTime` fields.
- Background callback dispatching also maps the new fields into `NotificationResponse`.

## Testing

Tested using:

- `melos bootstrap`
- `melos run analyze`
- `melos run test:unit <default>`

Also validated manually on a physical iPhone with a repeating local notification action. The response included:

- `notificationDeliveredAt=2026-06-28T13:01:00Z`
- `responseReceivedAt=2026-06-28T13:02:12.593Z` (1 minute after notification delivered when user selected an action)

Dart processing occurred later at `2026-06-28T13:03:12Z`, (1 minute later when the user launched the dart app)  confirming that the new fields preserve OS/native timing independently of delayed Dart processing.


flutter_local_notifications/lib/src/callback_dispatcher.dart
```diff
@@ -4,6 +4,28 @@ import 'package:flutter/services.dart';
...
+  if (value == null) {
+    return null;
+  }
+  if (value is int) {
+    return DateTime.fromMillisecondsSinceEpoch(value, isUtc: true);
+  }
+  if (value is num) {
+    return DateTime.fromMillisecondsSinceEpoch(value.toInt(), isUtc: true);
+  }
+  if (value is String) {
```
MaikuB: Can you highlight where in the PR that this code is needed to handle parsing a string?

ventisep: You're right. The string handling was defensive rather than required by the current native implementations. I have reviewed (with the help of codex) and iOS sends NSNumber from numberWithLongLong and macOS sends Int64, so Dart should receive an int via the standard Flutter codec. I’ll remove the String and num branches and keep the helper aligned with the actual platform channel contract.

ventisep:
That changes to the dismissal flow clashed with some of the changes for introducing the timestamps for exposing notification delivery and response timestamps.  I have now merged the latest master and retested using the unit and integration tests which all pass.

ventisep (commented):
I believe these changes are ready to merge now.  Can you review?  Or if you would like me to redo the changes on top of the new master I can do that instead.  I would like to remerge these with the master though as I am currently using my own forked version in my code and I don't want to be out of synch with the other maintenance changes others are making for too long.  Best Regards.  Paul Ventisei

MaikuB:
I'll get to this next when I can. I'm currently busy as I've been dealing with personal life matters where I'm moving. In the mean time I'll kick off the workflow but if there were conflicts, GitHub would have advised this already. I know this is a separate matter from you keeping your own fork up to date though

ventisep:
> I'll get to this next when I can. I'm currently busy as I've been dealing with personal life matters where I'm moving. In the mean time I'll kick off the workflow but if there were conflicts, GitHub would have advised this already. I know this is a separate matter from you keeping your own fork up to date though

No problem.  completely understand this is really not that urgent, I can keep my fork for a good while yet and I can always redo the changes on later versions if I really need to.  Hope the move goes well.

flutter_local_notifications/example/lib/main.dart
```diff
@@ -3668,6 +3680,19 @@ class SecondPageState extends State<SecondPage> {
   String? _payload;
   Map<String, dynamic>? _data;
 
+  String get _responseDelay {
```
MaikuB, line 3683: As this is computation and not directly to do with info returned by the native APIs via the plugin, could you remove this? Helps avoid confusion with what is available from the plugin

flutter_local_notifications/ios/flutter_local_notifications/Sources/flutter_local_notifications/FlutterLocalNotificationsPlugin.m
```diff
@@ -1033,6 +1045,14 @@ - (NSMutableDictionary *)extractNotificationResponseDict:
   NSNumber *notificationIdNumber = [NSNumber numberWithInteger:notificationId];
   notitificationResponseDict[@"notificationId"] = notificationIdNumber;
   notitificationResponseDict[PAYLOAD] = payload;
+  NSNumber *notificationDeliveredAt =
+      millisecondsSinceEpoch(response.notification.date);
```
MaikuB, line 1049: Whilst it realistically shouldn't happen, I believe `response.notification.date` could be null/nil. If so, can you add a defensive check around this?

flutter_local_notifications/test/macos_flutter_local_notifications_test.dart
```diff
@@ -765,6 +765,109 @@ void main() {
       ]);
     });
 
+    test('getNotificationAppLaunchDetails maps response timestamps', () async {
```
MaikuB, line 768: Is it possible to add tests on other platforms that don't support returning the timestamps that they assert that null is returned?

MaikuB (changes requested):
Apologies for the delay. Found some time whilst early days in my relocation. I tested the PR out and left some comments. Can you see if they make sense and can be tackled?

Issue #2795, opened 2026-07-07
Android 14+: PendingIntent.getActivity() should explicitly set background activity launch mode


Luca-CZ:
**Describe the bug**

On Android 14+ / API 34+, the plugin creates activity `PendingIntent`s without explicitly setting the background activity launch mode through `ActivityOptions.setPendingIntentCreatorBackgroundActivityStartMode(...)`.

According to Android's documentation on background activity launches, apps should explicitly opt in or opt out when creating activity `PendingIntent`s on newer Android versions instead of relying on implicit system behavior. In the current implementation, `FlutterLocalNotificationsPlugin` uses `PendingIntent.getActivity(...)` without passing an `ActivityOptions` bundle, which can cause security/compliance concerns and leaves the behavior undefined from the plugin's side.

This appears to affect the `PendingIntent.getActivity(...)` calls used for:
- notification taps
- notification actions that open the app UI

Android reference:
- https://developer.android.com/guide/components/intents-filters#using-a-pending-intent
- https://developer.android.com/reference/android/app/ActivityOptions#setPendingIntentCreatorBackgroundActivityStartMode(int)
- https://developer.android.com/reference/android/app/PendingIntent#getActivity(android.content.Context,int,android.content.Intent,int,android.os.Bundle)

**To Reproduce**

1. Fork this repository and use the example app, or integrate the plugin into an app targeting Android API 34+.
2. Build the Android app or inspect the plugin's Android native implementation.
3. Review `FlutterLocalNotificationsPlugin.java` and locate the `PendingIntent.getActivity(...)` calls used for notification selection and foreground notification actions.
4. Observe that the calls do not pass an `ActivityOptions` bundle with an explicit background activity start mode.

You can also reproduce this through static review/security validation:

1. Use a project that depends on this plugin version and targets Android 14+.
2. Run static analysis or security/compliance tooling against the Android build.
3. Inspect the flagged `PendingIntent.getActivity(...)` call sites.
4. Observe that the plugin does not explicitly opt in or opt out of background activity launch behavior.

**Expected behavior**

For Android 14+ / API 34+, the plugin should explicitly set the background activity start mode for every `PendingIntent.getActivity(...)` call by passing an `ActivityOptions` bundle.

A secure default would be to deny background activity starts unless explicitly required:

```java
if (VERSION.SDK_INT >= VERSION_CODES.UPSIDE_DOWN_CAKE) {
[...]

PR #2796, opened 2026-07-07
fix(android): set background activity launch mode for PendingIntent

Luca-CZ:
see https://github.com/MaikuB/flutter_local_notifications/issues/2795

## Summary
Adds explicit background activity launch handling for Android 14+ activity `PendingIntent`s created by the plugin.

This updates `PendingIntent.getActivity(...)` usage to pass an `ActivityOptions` bundle on API 34+ and explicitly set the pending intent creator background activity start mode.

## Motivation
Android 14 introduced stricter handling around background activity launches. When creating activity `PendingIntent`s, apps should explicitly opt in or opt out of background activity start behavior instead of relying on implicit or system-defined defaults.

Previously, the plugin created activity `PendingIntent`s for notification taps and notification actions without passing `ActivityOptions.setPendingIntentCreatorBackgroundActivityStartMode(...)`.

This made the behavior implicit on API 34+ and could trigger Android security/compliance findings. Making the behavior explicit allows the plugin to:

- align with Android guidance for activity pending intents
- avoid relying on system-defined background launch behavior
- use a secure default for notification-related activity launches

The change is backward compatible because it only applies on Android 14+ and keeps the existing behavior for older Android versions.

## Implementation notes

- Adds `ActivityOptions` usage to Android activity `PendingIntent` creation.
- Applies to the main notification tap `PendingIntent.getActivity(...)` path.
- Applies to notification action `PendingIntent.getActivity(...)` when `showsUserInterface` is true.
- Uses `Build.VERSION.SDK_INT >= VERSION_CODES.UPSIDE_DOWN_CAKE` to gate the new behavior.
- Uses `ActivityOptions.MODE_BACKGROUND_ACTIVITY_START_DENIED` as the explicit default.
- Passes `activityOptions.toBundle()` to the 5-argument `PendingIntent.getActivity(...)` overload on API 34+.
- Preserves the existing 4-argument `PendingIntent.getActivity(...)` call on older Android versions.

## Testing
Tested using:

melos bootstrap
melos run analyze
melos run test:unit <default>

### DIFF
--- a/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
+++ b/flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
@@ -6,6 +6,7 @@ import static android.provider.Settings.ACTION_REQUEST_SCHEDULE_EXACT_ALARM;
 import android.Manifest;
 import android.annotation.SuppressLint;
 import android.app.Activity;
+import android.app.ActivityOptions;
 import android.app.AlarmManager;
 import android.app.Notification;
 import android.app.NotificationChannel;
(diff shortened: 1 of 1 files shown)


MaikuB:
Thanks for the PR.  I noticed  what you mentioned around your testing approach. What's notably missing is mention around functional testing. The example app provides a lot of coverage around scenarios. I will check these too but how much of functional testing have you done via the example app and your own apps? Additionally, which scenarios have you tested?

PR #2797, opened 2026-07-07
[flutter_local_notifications] add conversationId/parentChannelId to AndroidNotificationChannel

PolynomialDivision:


flutter_local_notifications/android/src/main/java/com/dexterous/flutterlocalnotifications/FlutterLocalNotificationsPlugin.java
```diff
@@ -1214,6 +1215,81 @@ private static void setupNotificationChannel(
...
       NotificationManager notificationManager =
           (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+
+      // Conversation channels require API 30+. Below that, getConversationId()
+      // always returns null (the compat lib skips the platform call), causing
+      // the delete-and-recreate guard to fire on every call and discard user
+      // customisations. Fall through to the regular path on older devices.
+      if (!StringUtils.isNullOrEmpty(notificationChannelDetails.conversationId)
+          && !StringUtils.isNullOrEmpty(notificationChannelDetails.parentChannelId)
+          && VERSION.SDK_INT >= VERSION_CODES.R) {
```
MaikuB, line 1225: - The version check can be done upfront. If the version check fails,  there's no need to check the other conditions sooner. This would also make the code more aligned with what's described in the comments
- Please remove the comments as they are redundant since the code already describes what happens

flutter_local_notifications/android/build.gradle
```diff
@@ -43,7 +43,7 @@ android {
 
 dependencies {
     coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.1.4'
-    implementation "androidx.core:core:1.3.0"
+    implementation "androidx.core:core:1.13.0"
```
MaikuB, line 46: What was the reason behind picking this specific version?

MaikuB (changes requested):
Thanks for the PR. I left some feedback around some requested changes and a question to get more information

PR #2801, opened 2026-07-17
[flutter_local_notifications_windows] Feature parity for Windows apps without package identity

jjoelj:
For Windows apps NOT packaged as MSIX
- Adds ability to cancel individual notifications
- Adds ability to get all active notifications

Removes text in README stating this was not possible

Follows the same format as https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/c8f76d072df53d3622fb5440d63afb06cb9e7a10/Microsoft.Toolkit.Uwp.Notifications/Toasts/Compat/ToastNotifierCompat.cs#L115-L117


MaikuB (changes requested):
Rather than only provide a hardcoded default group name, could you also provide a way for those using the plugin to specify what group name they want as well? Ideally, the current group name in your PR is the default value of Dart parameter as well that can then be overridden

PR #2810, opened 2026-07-25
[flutter_local_notifications] add Android 16 live update notifications

### LINKED ISSUE #2773: [Feature Request] Support Android 16 Promoted Ongoing (Live Update) notifications
## Description

Android 16 (API level 36) introduces **Promoted Ongoing notifications**, also referred to as "Live Update" notifications. These are ongoing notifications that receive elevated visual treatment in the status bar and notification shade, designed for real-time status updates such as navigation, timers, delivery tracking, and emergency alerts.

To opt a notification into this behaviour, the app must set a boolean extra on the notification's `Bundle`:

```kotlin
builder.extras.putBoolean("android.requestPromotedOngoing", true)
```

The constant is defined as `Notification.EXTRA_REQUEST_PROMOTED_ONGOING`.

**There is currently no way to set arbitrary notification extras through `flutter_local_notifications`.** The plugin constructs the `NotificationCompat.Builder` internally and does not expose the `extras` bundle, so developers cannot use this Android 16 feature without bypassing the plugin entirely.

## Use case

Google designed Promoted Ongoing notifications for apps that show time-sensitive, continuously updating information:

- **Emergency/safety alert apps** displaying active warnings with countdowns
- **Navigation apps** showing turn-by-turn directions
- **Ride-sharing and delivery tracking** with live ETAs
- **Sports scores and live events** with real-time updates
- **Timers and stopwatches**

These are precisely the kinds of notifications that benefit most from the elevated status bar and notification shade placement that Promoted Ongoing status provides.

## Current workaround

The only way to use Promoted Ongoing notifications today is to bypass `flutter_local_notifications` entirely and post notifications via native platform code (Kotlin/Java) through a method channel.

This is problematic because:

1. It creates **two separate notification posting paths** in the app, one managed by the plugin and one managed by custom native code.
2. It **does not work in background `FlutterEngine` contexts** (e.g. the FCM background message handler, WorkManager tasks) without additional setup to register the method channel in the background isolate.
3. It **defeats the purpose** of using a cross-platform notification plugin, forcing developers to maintain platform-specific notification logic anyway.

## Proposed API

Either of the following approaches would solve this:

### Option A: Dedicated field (narrower scope, simpler)

Add a `requestPromotedOngoing` boolean to `AndroidNotificationDetails`:

```dart
AndroidNotificationDetails(
  'channel_id',
  'Channel Name',
  ongoing: true,
  requestPromotedOngoing: true,
  // ...
)
```

On the native side this would call:

```kotlin
builder.extras.putBoolean("android.requestPromotedOngoing", true)
```

### Option B: General extras map (broader scope, more flexible)

Add an `extras` parameter to `AndroidNotificationDetails` that maps to `Notification.Builder.extras`:

```dart
AndroidNotificationDetails(
  'channel_id',
  'Channel Name',
  ongoing: true,
  extras: <String, Object>{
    'android.requestPromotedOngoing': true,
  },
  // ...
)
```

This would allow developers to set any notification extras without requiring a plugin update for each new Android feature. The supported value types could be limited to `bool`, `int`, `double`, and `String` to keep serialisation straightforward.

**Option A is probably the better choice** for this specific feature, as it is discoverable, type-safe, and can include validation (e.g. warning if `ongoing` is not also set to `true`). Option B could be considered as a complementary addition for future-proofing.

## Android documentation references

- [Promoted Ongoing notifications guide](https://developer.android.com/develop/ui/views/notifications/promoted-ongoing)
- [`Notification.EXTRA_REQUEST_PROMOTED_ONGOING`](https://developer.android.com/reference/android/app/Notification#EXTRA_REQUEST_PROMOTED_ONGOING) constant

## Related

Android 16 also introduces `Notification.Builder.setShortCriticalText(CharSequence)`, which sets the text displayed in the status bar chip for Live Update notifications (e.g. "3 min", "2.1 mi"). This would also be a valuable addition but is a separate feature request, as it involves a builder method rather than an extras bundle entry.

## Environment

- Android 16 (API level 36)
- `flutter_local_notifications` latest stable
- Flutter 3.x

vlaushkin:
Resolves #2773 and adds the Android "live update" side of #2755.

Adds support for Android 16 (API 36) promoted ongoing "live update"
notifications, together with the Android 17 semantic colouring of the progress
track. Everything is applied through `androidx.core`'s `NotificationCompat` /
`NotificationManagerCompat` wrappers, which gate on the runtime SDK internally.
The new APIs are no-ops (or return `false`) on older devices, so `minSdk` is
unchanged and none of them need an `SDK_INT` check.

## Summary
- `AndroidNotificationDetails.requestPromotedOngoing` (bool) → `NotificationCompat.Builder.setRequestPromotedOngoing(...)`.
- `AndroidNotificationDetails.shortCriticalText` (String) → `setShortCriticalText(...)`, the text shown in the status bar chip (e.g. `"3 min"`, `"2.1 mi"`).
- New `ProgressStyleInformation` style → `NotificationCompat.ProgressStyle`: `segments` and `points` (each with length/position, colour, `semanticStyle` and a stable `id`), `progress`, `progressIndeterminate`, `styledByProgress`, and `progressTrackerIcon` / `progressStartIcon` / `progressEndIcon`.
- `AndroidFlutterLocalNotificationsPlugin.canPostPromotedNotifications()` → `NotificationManagerCompat.canPostPromotedNotifications()`, so apps can check whether the user allows promotion. Mirrors the existing `areNotificationsEnabled()`.
- Bumps `androidx.core` from 1.3.0 to 1.19.0, which provides these APIs.

Android references: [promoted ongoing / live update](https://developer.android.com/develop/ui/views/notifications/live-update), [`NotificationCompat.ProgressStyle`](https://developer.android.com/reference/androidx/core/app/NotificationCompat.ProgressStyle), [Android 17 semantic coloring](https://developer.android.com/about/versions/17/features).

## Approach — Option A vs Option B
You asked in #2773 which approach I'd recommend. I went with **Option A (dedicated typed fields)** rather than **Option B (a generic `extras` map)**: a typed field is discoverable, documented and type-safe, and it keeps raw Android extra keys out of app code. A generic extras passthrough would also set a precedent for adding future Android features as raw keys. The same principle applies to the rest of the feature: `shortCriticalText` is a typed field, progress is a typed `ProgressStyleInformation` class, and the semantic styles are a `SemanticStyle` enum mirroring the existing `SemanticAction`.

## Why this is one PR, and why it doesn't close #2755
`requestPromotedOngoing` already has an effect on its own: given the app's `POST_PROMOTED_NOTIFICATIONS` permission, it promotes an ongoing notification (a status bar chip), with no special style required. The other pieces add the content. `shortCriticalText` sets the chip's text, and `ProgressStyle` (the Android part of #2755) provides the progress track behind the navigation and delivery cases. Bundling them keeps the example a real live update. Happy to split if you'd prefer.

#2755 also asks for the iOS equivalent ("live data"), which is deliberately left out. It is a genuinely different feature: iOS Live Activities are built with ActivityKit and a per-app SwiftUI widget extension, not the notification APIs this plugin wraps, and there is a separate `live_activities` package for it. This PR implements the Android side only, which is why it references rather than closes #2755.

## `semanticStyle` and the 1.19.0 bump
Segments and points carry a `semanticStyle` (`unspecified` / `info` / `safe` / `caution` / `danger`), so the system can colour parts of the track by meaning.

- The bump raises the bar for apps, not just for the plugin. The AAR metadata of `androidx.core` 1.19.0 requires consumers to build with `compileSdk 37` and AGP 9.1.0 or newer, where 1.18.0 asked for 36 and 8.9.1. That matches the compileSdk 37 / AGP 9.1.1 you just landed, but an app still on 36 will fail `checkDebugAarMetadata`.
- The styling takes effect on API 37. androidx passes the value to the platform only through its `Api37Impl`, so on Android 16 it is carried but dropped. Per the androidx docs, "if an app specifies both color and semantic style, the color overrides the style", so the navigation example sets no explicit colours.

On an Android 17 device, segments with colours set render in those colours; with the colours removed they render in the system's green/amber/red palette.

## Example app
Three buttons in the Android section, each of which first calls `canPostPromotedNotifications()` and shows a SnackBar if promotion is unavailable:
- **food delivery**, a staged `ProgressStyle` with coloured segments and points, a tracker icon that advances (bag → truck → check), store and home start/end icons, and a `shortCriticalText` chip;
- **navigation**, three segments carrying the `safe` / `caution` / `danger` semantic styles (traffic along the route), with the ETA in the chip;
- **timer**, the lightweight path: a promoted ongoing notification with a chronometer countdown and no style, so the chip content comes from `when`.

Adds `POST_PROMOTED_NOTIFICATIONS` to the example manifest and a few vector drawables.

## Tests
- Added `requestPromotedOngoing` / `shortCriticalText` to the existing channel-argument expectations.
- New `show with progress style settings` test asserting the full `ProgressStyleInformation` map sent over the channel: segments, points, ids, colours, semantic styles, and tracker/start/end icons.

## Test plan
- [x] `melos run analyze` (all packages, `--fatal-infos`)
- [x] `flutter test` in the main package (131 tests)
- [x] Android Java build + unit tests (`melos run test:unit:android`)
- [x] Manual on a physical Android 16 device: `dumpsys` shows the notification promoted (`PROMOTED_ONGOING` retained) with the `ProgressStyle` template, `shortCriticalText` and all icons; the status bar chip renders when the shade is collapsed
- [x] Manual on a physical Android 17 device: the semantic styles render on the track

Happy to adjust naming, scope or approach.


MorganMLGman:
Hi @MaikuB, is there anything blocking review of this PR? It would be great to have for a live progress notification from a foreground service. Thanks @vlaushkin for the work.

MaikuB:
Thanks for the PR @vlaushkin. I've just landed some changes to do bumping the compileSdk and AGP so happy for changes to `setSemanticStyle()` to be added in via this PR. I'll then give this PR a bash afterwards

vlaushkin:
Thanks! Rebased onto master and added `setSemanticStyle` for `ProgressStyle`
segments and points, exposed as a `SemanticStyle` enum mirroring the existing
`SemanticAction`.

Two things worth knowing before you look:

- It needs `androidx.core` 1.19.0, whose AAR metadata requires apps to build
  with `compileSdk 37` and AGP 9.1.0 or newer, where 1.18.0 asked for 36 and
  8.9.1. That is fine for the plugin after your bump, but it applies to apps
  using it too.
- The styling itself takes effect on API 37, since androidx forwards the value
  to the platform only through its `Api37Impl`. Per the androidx docs, "if an
  app specifies both color and semantic style, the color overrides the style",
  so the navigation example now sets semantic styles and no colours. Checked on
  an Android 17 device: with colours the track renders in those colours, without
  them it renders in the system's semantic palette.

I have also updated the description, and fixed the point from the Copilot
review: the example no longer reports promotion as disabled when the device is
simply older than Android 16.

vlaushkin:
Makes sense, done in 6af4871. The Dart docs now say Android 16 and Android 17 instead of API levels 36 and 37.

PR #2813, opened 2026-08-07
[flutter_local_notifications_windows] fix crash when Windows push notification service is not running

j-bibodi:
The crash originates in ffi_api.cpp where ToastNotificationManager::CreateToastNotifier() and ::History() are called without any try/catch around winrt::hresult_error. When the Windows Push Notification service (WpnService) is stopped/disabled, these WinRT calls throw, propagating an unhandled C++ exception across the FFI boundary.

To reproduce the issue:

1. Execute "Stop-Service WpnUser*" in PowerShell to stop the Windows push notification service.
2. Run the sample/example app -- https://github.com/MaikuB/flutter_local_notifications/tree/master/flutter_local_notifications/example on Windows device
3. Observe app crashes/closes automatically.


MaikuB:
Thanks for the PR. It makes sense with what the code change is trying to do but when following the steps you mentioned whilst debugging, I still see an error. Any idea on what is going on and if I am missing something? The following screenshot is what I saw

Screenshot 2026-09-13 at 12 17 41 pm

aw-cf:
@MaikuB did you make sure to use the local version of the windows platform package instead of the public one?

A dependency override for the windows - and all other local packages - can be achieved by adding this to the `pubspec.yaml`:

```
dependency_overrides:
  flutter_local_notifications_linux:
    path: ../../flutter_local_notifications_linux
  flutter_local_notifications_platform_interface:
    path: ../../flutter_local_notifications_platform_interface
  flutter_local_notifications_web:
    path: ../../flutter_local_notifications_web
  flutter_local_notifications_windows:
    path: ../../flutter_local_notifications_windows
```

`flutter pub get` will show if the override worked or not:

```
Resolving dependencies...
Downloading packages...
...
  ! flutter_local_notifications_linux 8.0.1 from path ../../flutter_local_notifications_linux (overridden)
  ! flutter_local_notifications_platform_interface 12.2.0 from path ../../flutter_local_notifications_platform_interface (overridden)
  ! flutter_local_notifications_web 1.0.0 from path ../../flutter_local_notifications_web (overridden)
  ! flutter_local_notifications_windows 3.1.1 from path ../../flutter_local_notifications_windows (overridden)
...
```

The fix should be picked up and work as expected once this override is in place. It may also be a nice improvement to add these `dependency_overrides` to the `pubspec.yaml` by default.

Dart docs on `dependency_overrides`: https://dart.dev/tools/pub/dependencies#dependency-overrides

MaikuB:
Yes I did as melos handles this

PR #2815, opened 2026-08-11
[flutter_local_notifications] fix macOS subtitle not displayed when set via DarwinNotificationDetails

JIWON1923:
### Problem
On macOS, setting `subtitle` via `DarwinNotificationDetails` had no effect, and the subtitle was not displayed in the notification.

### Root Cause
`DarwinNotificationDetails.subtitle` is serialized into `platformSpecifics` in the method channel call, but the macOS Swift implementation was reading `subtitle` from the top-level `arguments`, where it is never set.

```dart
// subtitle is sent inside platformSpecifics
_channel.invokeMethod('show', {
  'title': title,
  'body': body,
  'platformSpecifics': {
    'subtitle': subtitle,
    ...
  },
});
```

``` swift
// Previous macOS implementation: subtitle was incorrectly read from top-level arguments
if let subtitle = arguments["subtitle"] as? String {
    content.subtitle = subtitle
}
```

## Fix
- Remove the dead code that reads `subtitle` from top-level arguments
- Read `subtitle` from `platformSpecifics`, consistent with the iOS implementation


MaikuB:
Thanks for catching that

PR #2820, opened 2026-09-01
[flutter_local_notifications_web] focus the tab on notification click + wait for SW activation + cache burst for SW file changes

coder-with-a-bushido:
1. Right now when using the plugin, the notification click doesn't focus the tab on web when the tab is open. This PR fixes that since that's the naturally expected behaviour.
2. initialize registered the service worker and immediately subscribed to messages. If the worker was still installing, notification clicks could be missed. Now, initialize waits until the worker is activated (10s timeout) before listening for clicks.
3. Browsers could also keep serving an old notifications_service_worker.js, so SW updates did not always take effect. So, I fixed it by adding ?v=1.0.0 to SW URL so a version bump forces browsers to fetch the new file instead of the cached one.


coder-with-a-bushido:
Hi @MaikuB, it would be nice to get this PR reviewed. It has minimal changes but I think these high impact for users. I am ready to do the changes requested to get this merged.

Thank you in advance :)

flutter_local_notifications_web/lib/src/plugin.dart
```diff
@@ -12,6 +13,27 @@ import 'handler.dart';
...
+    if (worker.state == 'activated' && !completer.isCompleted) {
+      completer.complete();
+    }
+  }
+
+  worker.onstatechange = ((Event _) {
+    completeIfActivated();
+  }).toJS;
+  completeIfActivated();
+  return completer.future.timeout(const Duration(seconds: 10));
```
MaikuB, line 34: What's the reason for 10 seconds? Looks like a magic number

